PR #6, opened 2023-09-10
Using the mapping protocol for ConfigReader

Amaras:
The [documentation for ConfigParser](https://docs.python.org/3/library/configparser.html) recommends using the mapping protocol for all new projects instead of the legacy API.

This PR aims to change all cases of this pattern:
```py
section = 'XXX'
# here self is a `ConfigParser` subclass instance
if self.has_section(section):
  key = 'YYY' 
  if self.has_option(section, key):
    var = self.get(section, key)
    # do something with var
  else:
    # log
else:
  # log and return
```

The goal is to use the following pattern instead:
```py
section_key = 'XXX'
try:
  section = self[section_key]
except KeyError:
  # log and return

key = 'YYY'
try:
  var = section[key]
except TypeError:
  # log and return : not a section
except KeyError:
  # log: no option YYY
# use var
```

I have also added several comments on parts of the code that could (theoretically) lead to a crash or security vulnerability

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,9 @@
 /fees/
 /docs/images/src/
 papi-web.ini
-/tmp/
+tmp/
 __pycache__/
+
+# Vim scratch files
+*.swp
+*.swo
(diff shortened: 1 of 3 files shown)


pascalaubry (approved):

pascalaubry (approved):

Amaras:
There are some TODOs left, but the small-scale test proved it works

Issue #4, opened 2023-09-07
Should we use the `ConfigReader` mapping protocol?


Amaras:
In Python, the usual way to handle errors is "EAFP" (It's easier to ask for forgiveness than permission).
This is especially true for interfacing with the operating system, etc.

The point is that the current version of the code uses a lot of code like this:
```py
section = 'XXX'
if not self.has_section(section):
	# log no section error
else:
	key = 'YYY'
	if not self.has_option(section, key):
		# log no key error
	else:
		var = self.get(section, key)
		# do something with var
```

This could probably be changed to the following pattern:
```py
try:
	section = self['XXX'] # config_reader['XXX']
	try:
		var = section['YYY']
		# do something with var
	except (KeyError, TypeError): # TypeError can happen if 'XXX' is a key and we expected a section
		# log no key error
except KeyError:
	#log no section error
```

I don't feel comfortable trying to understand the legacy API that is currently used, so this might be worthwhile (we should not get this to the main branch before equivalence of behaviour is established first, of course)

pascalaubry:
Yes we should, waiting for your PR to complete ;-)

Amaras:
Mostly done now with #6

Amaras:
I can probably be closed if #7 is merged in.
I can separate out the mapping protocol changes from that PR if the rest is not wanted.

PR #7, opened 2023-09-21
Use composition over inheritance

Amaras:
This PR is a change in architecture, using composition over straight inheritance from `ConfigReader`.

The comparison between the two design choices has been discussed for decades already, and _Design Patterns: Elements of Reusable Object-Oriented Software_ (a book listing all design patters present at the time in OOP code) has recommended the use of composition over inheritance since 1994.

One secondary aim of this PR is to allow for the generelization of configuration formats (mainly TOML and YAML, maybe JSON, .properties, etc.)
This would allow us to have a standardized format available to expert users, with its own documentation (if documented at all).
This should be discussed in its own issue, though.

### DIFF
--- a/common/papi_web_config.py
+++ b/common/papi_web_config.py
@@ -28,9 +28,9 @@ DEFAULT_WEB_LAUNCH_BROWSER: bool = True
 
 
 @singleton
-class PapiWebConfig(ConfigReader):
+class PapiWebConfig:
     def __init__(self):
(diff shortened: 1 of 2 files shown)


Amaras:
This should not be expected to work this week in any case, as I can see bugs in what I wrote from looking at the diff here.

Amaras:
From then on, I think this PR closes #4

PR #9, opened 2023-09-24
Added missing requirement

Amaras:
Adding a missing requirement to requirements.txt.
The problem arised when trying to launch the server, at a seemingly random time.

While it could be a dependency of another package, I prefer having it explicit, just in case.

Does affect the code, should be safe to merge right now.


Amaras:
Solved in a main commit, closing this

PR #20, opened 2024-01-22
Add two-way support for TRF

Amaras:
I opened this PR to focus efforts on the TRF support.

This is meant to bypass the papi database, especially by converting back and forth between the papi-web Tournament and TRF Tournament objects.

Further commits will follow when I'm more available.


Amaras:
In order to generate a TRF, we need the birthdate of each of the player, as well as their FIDE ID.
This increases the complexity of the change, as I don't know how to interact with Access well (we need to read those two values).

PR #28, opened 2024-03-09
Replace Django with Litestar and add a few features

### LINKED ISSUE #24: Consider using a web framework other than Django for performance
With respect to reports by a user, it seems as though Django is too slow for a quick startup.
I recommend considering a micro-framework like [Flask](https://flask.palletsprojects.com/en/3.0.x/), as it should start quicker (although we need to measure before we get the conclusion)

Amaras:
This PR replaces Django with Litestar (closes #24), and adds the following changes:

1. Adds checkmarks in the playerlist (boards screen at the start of the tournament): players with filled checkmarks are checked-in (wonderful when coupled with ChessEvent).
2. [WIP] Add flags to track illegal moves (on boards screens with `update = on` and `record_illegal_moves = on`): click on empty flag to add an illegal move flag to a player, click on a filled flag to remove it (currently not done).

More features may arrive before merging.


chessevent/action_selector.py
```diff
@@ -34,54 +34,54 @@ def __get_chessevent_tournaments(cls, event: Event) -> list[Tournament]:
         tournaments: list[Tournament] = []
         for tournament in event.tournaments.values():
             if not tournament.chessevent_tournament_name:
-                logger.warning(f'Connexion à Chess Event non définie pour le tournoi [{tournament.id}]')
+                logger.warning('Connexion à Chess Event non définie pour le tournoi [%s]', tournament.id)
```
adrienave, line 37: I'm confused with this kind of changes. Why removing f-strings that are more readable than other formatting?

Commit states it's for fixing Pylint warnings, but Pylint explicitly says f-strings are better: https://pylint.readthedocs.io/en/stable/user_guide/messages/convention/consider-using-f-string.html
Maybe the warnings were in fact only about using f-strings without interpolation like a bit below for "La création des fichiers Papi n\'est possible pour aucun tournoi" (https://pylint.readthedocs.io/en/stable/user_guide/messages/warning/f-string-without-interpolation.html)

data/tournament.py
```diff
@@ -160,28 +166,28 @@ def read_database(self):
     def _calculate_current_round(self):
         round_infos: dict[int, dict[str, bool]] = {}
         paired_rounds: list[int] = []
-        for round in range(1, self._rounds + 1):
-            round_infos[round] = {
+        for round_ in range(1, self._rounds + 1):
```
adrienave, line 169: Alternatively, could be renamed `tournament_round`, `chess_round`, `match_round` etc. if we want to avoid `_` at the end.

data/tournament.py
```diff
@@ -563,7 +607,7 @@ def _build_tournament(self, tournament_id: str):
             'chessevent_connection_id',
             'chessevent_tournament_name',
         ]
-        for key, value in section.items():
+        for key, _ in section.items():
```
adrienave, line 610: Not sure about the typing of `section`, but let's assume it's a `dict`.
Then `section.keys` could be used instead of `items`, because value is not necessary here.

pascalaubry (approved):
Ready to merge changes to dev.

data/util.py
```diff
@@ -773,3 +751,18 @@ def from_str(cls, value) -> Self:
...
+    NO_CHANGE = 1
+    RECENT_CHANGE = 2
+
+    def __bool__(self):
+        match self:
+            case NeedsUpload.YES:
+                return True
+            case NeedsUpload.NO_CHANGE | NeedsUpload.RECENT_CHANGE:
+                return False
+            case _:
```
adrienave, line 767: I'm wondering if this default case is even possible... What could be of type `NeedsUpload` but not part of the acceptable values for it? Unless the class is instantiated manually but that's not what you are supposed to do with `Enum` I think.
Just saying, because if we get rid of the default case, the method could be simplified to `return NeedsUpload.YES == self`.

database/papi_template.py
```diff
@@ -7,384 +7,384 @@
...
-			bz2.decompress(
-				base64.decodebytes(
-					b64
-				)
-			)
-		)
+    match papi_version:
+        case '3.3.6':
+            b64 = (
+                b'QlpoOTFBWSZTWZodmZEAE8l/////////////////////////////////////////////4DCfNe706t2+'
```
adrienave, line 19: Consider extracting this bytes content to a dedicated file at some point, it may be easier for the formatting.

web/messages.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@dataclass
+class Message:
+    text: str
+    level: int
+    tag: str | None = field(default=None, init=False)
+    html_class: str | None = field(default=None, init=False)
+
+    DEBUG = 10
```
adrienave, line 13: Maybe these values could be moved to an `Enum` in the same file.

adrienave (changes requested):

adrienave:
Just noticed the PR has been merge before my review, bad timing from myself.
It's okay, everything was about code smells / refactoring, nothing wrong in the logic 👍

chessevent/action_selector.py
Amaras: I totally agree with you, except logging is an exception to this rule: see the logging cookbook [formatting styles section](https://docs.python.org/3/howto/logging-cookbook.html#formatting-styles), and [optimization section](https://docs.python.org/3/howto/logging.html#optimization).

data/util.py
Amaras: A future enum member?

It's supposed to be deleted anyway once we get rid of papi as a base, but I agree with you on that.

I'm working in the continuity of this branch, so don't hesitate to repeat your review on the future PR I'll open

chessevent/action_selector.py
adrienave: Ok makes sense, I didn't know f-strings were not formatting the input at the same instant than format method.
Thanks for the references.

PR #37, opened 2024-08-31
Use database storage and new web interfaces

pascalaubry:


common/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+def format_timestamp_date_time(ts: float = None) -> str:
+    return datetime.strftime(datetime.fromtimestamp(ts if ts is not None else time.time()), '%Y-%m-%d %H:%M')
+
+
+def format_timestamp_date(ts: float = None) -> str:
+    return datetime.strftime(datetime.fromtimestamp(ts if ts is not None else time.time()), '%Y-%m-%d')
+
+
+def format_timestamp_time(ts: float) -> str:
+    return datetime.strftime(datetime.fromtimestamp(ts if ts is not None else time.time()), '%H:%M')
```
Amaras: For each parameter: should be a `float | None` instead of a `float`

common/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+RGB = namedtuple('RGB', ['red', 'green', 'blue'])
+
+logger: Logger = get_logger()
+
+
+def check_rgb_str(color: str) -> str:
+    rgb: RGB = hexa_to_rgb(color)
+    if rgb:
+        return rgb_to_hexa(hexa_to_rgb(color))
```
Amaras: Why recompute?

chessevent/event_selector.py
```diff
@@ -1,28 +1,27 @@
...
+    def __init__(self):
         self.__silent: bool = False
-        self.__config: PapiWebConfig = config
 
-    def run(self) -> bool:
-        events: List[Event] = get_events_sorted_by_name(False, with_tournaments_only=True)
-        self.__silent = True  # verbose on the first call only
+    @staticmethod
+    def run() -> bool:
+        events: List[Event] = EventLoader.get(request=None, lazy_load=True).events_sorted_by_name
```
Amaras: We should get rid of the `typing.List` annotation, and use the builtin `list`

data/loader.py
```diff
@@ -0,0 +1,194 @@
...
+    def get(request: HTMXRequest | None, lazy_load: bool):
+        if not request:
+            return EventLoader(lazy_load=lazy_load)
+        event_loader: EventLoader = request.state.get('event_loader')
+        if event_loader and lazy_load:
+            return event_loader
+        event_loader = EventLoader(lazy_load=False)
+        request.state['event_loader'] = event_loader
+        return event_loader
+
```
Amaras: This feels more like a `classmethod`, rather than a `staticmethod`.

data/loader.py
```diff
@@ -0,0 +1,194 @@
...
+            self._loaded_stored_events_by_id[uniq_id] = self._load_stored_event(uniq_id)
+            return self._loaded_stored_events_by_id[uniq_id]
+
+    @property
+    def event_uniq_ids(self) -> list[str]:
+        if self._event_uniq_ids is None:
+            self._event_uniq_ids = [
+                file.stem for file in PapiWebConfig().event_path.glob(f'*.{PapiWebConfig().event_ext}')
+            ]
+        return self._event_uniq_ids
```
Amaras: Maybe look into `functools.cached_property`

export.py
```diff
@@ -65,29 +63,34 @@ def build_exe():
...
         file for file in htmx_dir.glob('**/*')
         if file.is_file()
     ]
-    files += [Path('.') / 'database' / 'sql' / 'create_event.sql', ]
+    sql_dir: Path = Path('.') / 'database' / 'sql'
+    files += [sql_dir / 'create_event.sql', ]
+    yml_dir: Path = Path('.') / 'database' / 'yml'
+    files += [
+        file for file in yml_dir.glob('*.yml')
+    ]
```
Amaras:
```suggestion
    files += list(yml_dir.glob('*.yml'))
```

web/server_engine.py
```diff
@@ -42,24 +43,35 @@ def __init__(self):
...
+        url_map: dict[str, list[str]] = {}
+        for route in app.routes:
+            for handler in route.route_handlers:
+                if handler.name:
+                    if handler.name not in url_map:
+                        url_map[handler.name] = []
+                    url_map[handler.name].append(route.path)
+        for name in sorted(url_map.keys()):
+            logger.debug(f'{name}: {url_map[name]}')
+        uvicorn.run(app, host=papi_web_config.web_host, port=papi_web_config.web_port, log_level='info', )
```
Amaras: A `defautdict` would remove the line 69 check.

web/settings.py
```diff
@@ -9,17 +9,29 @@
...
+from web.views_admin_screen import AdminScreenController
+from web.views_admin_timer import AdminTimerController
+from web.views_admin_tournament import AdminTournamentController
+from web.views_background import BackgroundController
+from web.views_user import UserIndexController
+from web.views_user_event import UserEventController
+from web.views_user_screen import UserScreenController
+from web.views_user_screen_set import UserScreenSetController
+from web.views_user_tournament import UserCheckInController, UserIllegalMoveController, UserResultController, \
+    UserDownloadController
```
Amaras: All those are `Controller`s.
Maybe you should put them in a `web.controllers` module?
Also, you should probably have `web.controllers.user` and `web.controllers.admin` submodules, because all those `views_admin_*` and `views_user_*` seem like a bad organization.

data/loader.py
```diff
@@ -0,0 +1,194 @@
...
+        if not request:
+            return EventLoader(lazy_load=lazy_load)
+        event_loader: EventLoader = request.state.get('event_loader')
+        if event_loader and lazy_load:
+            return event_loader
+        event_loader = EventLoader(lazy_load=False)
+        request.state['event_loader'] = event_loader
+        return event_loader
+
+    def clear_cache(self, event_uniq_id: str = None):
```
Amaras:
```suggestion
    def clear_cache(self, event_uniq_id: str | None = None):
```

common/papi_web_config.py
```diff
@@ -42,7 +31,7 @@
 @singleton
 class PapiWebConfig:
```
Amaras: Couldn't all the properties not using the reader be class attributes instead?
I don't feel like properties are the right way to express constants.
Properties are meant to be used to do some work in addition to just returning instance attributes.

data/loader.py
```diff
@@ -0,0 +1,194 @@
...
+            ]
+        return self._event_uniq_ids
+
+    @property
+    def stored_events_by_id(self) -> dict[str, StoredEvent]:
+        if self._stored_events_by_id is None:
+            self._stored_events_by_id: dict[str, StoredEvent] = {}
+            for uniq_id in self.event_uniq_ids:
+                self._stored_events_by_id[uniq_id] = self.load_stored_event(uniq_id)
+        return self._stored_events_by_id
```
Amaras: Feels like a `functools.cached_property`.
The pattern to recognised a cached property is the following:
```py
@property
def something(self) -> T: # T is a type
	if self._something is None:
		# initialize self._something
	return self._something
```

data/loader.py
```diff
@@ -0,0 +1,194 @@
...
+            ], key=lambda event: (-event.stop, -event.start, event.name))
+        return self._coming_events
+
+    @property
+    def public_events(self) -> list[Event]:
+        if self._public_events is None:
+            self._public_events = sorted([
+                event for event in self.events_by_id.values()
+                if event.public
+            ], key=lambda event: event.name)
```
Amaras: Should be a generator expression. I think `sorted()` accepts any iterable:
```suggestion
            self._public_events = sorted((
                event for event in self.events_by_id.values()
                if event.public
            ), key=lambda event: event.name)
```

Or even (formatting could be better):
```suggestion
            self._public_events = sorted(filter(lambda event: event.public, self.events_by_id.values()), key = lambda  event: event.name)
```

Or better yet:
```suggestion
	self._public_events = sorted(filter(attrgetter('public'), self.events_by_id.values(), key=attrgetter('name'))
```

database/store.py
```diff
@@ -1,31 +1,27 @@
...
-    date: str
-    round: int | None
-    event: str | None
-    text_before: str
-    text_after: str
+    order: int | None = field(default=None)
+    date_str: str | None = field(default=None)
+    time_str: str | None = field(default=None)
+    text_before: str | None = field(default=None)
+    text_after: str | None = field(default=None)
```
Amaras:
```suggestion
    order: int | None = None
    date_str: str | None = None
    time_str: str | None = None
    text_before: str | None = None
    text_after: str | None = None
```

database/store.py
```diff
@@ -35,114 +31,143 @@ class StoredChessEvent:
...
-    order: int
+    name: str | None
+    order: int | None
+    fixed_boards_str: str | None
     first: int | None
     last: int | None
-    part: int | None
-    parts: int | None
-    number: int | None
+    last_update: float = field(default=0.0)
```
Amaras:
```suggestion
    last_update: float = 0.0
```

database/store.py
```diff
@@ -35,114 +31,143 @@ class StoredChessEvent:
...
     timer_id: int | None
+    players_show_unpaired: bool | None
     results_limit: int | None
-    results_tournaments_str: str | None
+    background_image: str | None
+    background_color: str | None
+    results_tournament_ids: list[int] = field(default_factory=list[int])
     stored_screen_sets: list[StoredScreenSet] = field(default_factory=list[StoredScreenSet])
+    last_update: float = field(default=0.0)
+    public: bool = field(default=True)
```
Amaras:
```suggestion
    last_update: float = 0.0
    public: bool = True
```

database/store.py
```diff
@@ -35,114 +31,143 @@ class StoredChessEvent:
...
-    results_tournaments_str: str | None
-    range: str | None
+    players_show_unpaired: bool | None
     first: int | None
     last: int | None
-    part: int | None
     parts: int | None
     number: int | None
+    public: bool = field(default=True)
+    last_update: float = field(default=0.0)
```
Amaras:
```suggestion
    public: bool = True
    last_update: float = 0.0
```

database/store.py
```diff
@@ -35,114 +31,143 @@ class StoredChessEvent:
...
 class StoredRotator:
     id: int | None
     uniq_id: str
-    families_str: str | None
-    screens_str: str | None
+    family_ids: list[int] | None
+    screen_ids: list[int] | None
     delay: int | None
+    show_menus: bool | None
+    public: bool = field(default=True)
```
Amaras:
```suggestion
    public: bool = True
```

database/store.py
```diff
@@ -35,114 +31,143 @@ class StoredChessEvent:
...
+    background_image: str | None = field(default=None)
+    background_color: str | None = field(default=None)
     update_password: str | None = field(default=None)
     record_illegal_moves: int | None = field(default=None)
-    allow_deletion: bool | None = field(default=None)
+    allow_results_deletion_on_input_screens: bool | None = field(default=None)
+    version: str | None = field(default=None)
+    timer_colors: dict[int, str | None] = field(default=None)
+    timer_delays: dict[int, int | None] = field(default=None)
+    last_update: float = field(default=0.0)
```
Amaras: Why not simply use `attribute: T = default` instead of `attribute: T = field(default=default)`? (when applicable, of course)

export.py
```diff
@@ -4,17 +4,15 @@
...
 from common.logger import get_logger
 
 logger: Logger = get_logger()
 
 BUILD_DIR: Path = Path('build')
 DIST_DIR: Path = Path('dist')
 DATA_DIR: Path = Path('export-data')
-CUSTOM_DIR: Path = Path('custom')
-basename: str = f'papi-web-{PAPI_WEB_VERSION}'
+basename: str = f'papi-web-{PapiWebConfig().version}'
```
Amaras: If you use class attributes, you don't need to initialize the singleton until absolutely necessary

data/event.py
```diff
@@ -1,305 +1,535 @@
...
+from data.screen import Screen
+from data.timer import Timer
+from data.tournament import Tournament
+from data.util import ScreenType
+from database.store import StoredEvent
+from web.views_background import BackgroundWebContext
 
 logger: Logger = get_logger()
 
+event_last_load_date_by_uniq_id: dict[str, float] = {}
```
Amaras, line 27: Are you sure you want to use a global variable for this?

data/event.py
```diff
@@ -1,305 +1,535 @@
...
+    def formatted_stop_date(self) -> str:
+        return format_timestamp_date(self.stop)
+
+    @property
+    def formatted_stop_time(self) -> str:
+        return format_timestamp_time(self.stop)
+
+    @property
+    def players_number(self) -> int:
+        return sum([len(tournament.players_by_name_with_unpaired) for tournament in self.tournaments_by_id.values()])
```
Amaras: Use generator expressions
```suggestion
        return sum((len(tournament.players_by_name_with_unpaired) for tournament in self.tournaments_by_id.values()))
```

data/event.py
```diff
@@ -1,305 +1,535 @@
...
+    @property
+    def public(self) -> bool:
+        return self.stored_event.public
+
+    @property
+    def tournaments_sorted_by_uniq_id(self) -> list[Tournament]:
+        if self._tournaments_sorted_by_uniq_id is None:
+            self._tournaments_sorted_by_uniq_id = sorted(
+                self.tournaments_by_id.values(), key=lambda tournament: tournament.uniq_id)
+        return self._tournaments_sorted_by_uniq_id
```
Amaras: Feels like a `functools.cached_property`

Amaras (changes requested):
I'm missing at least 50 files, but here's a first review.

204 files is too much to review quickly, I recommend publishing a pre-release version 2.4-web for beta testing, and once we can review everything properly for the merge, we can probably release version 2.4.0.

docs/81-urls.md
```diff
@@ -2,6 +2,9 @@
 
 # Papi-web - Annexe technique : URLs utilisées par le serveur web
 
+> [!CAUTION]
+> Cette page décrivant les URLs utilisées par l'application est obsolète.
+
```
Amaras, line 7: I agree that the URLs will be obsolete anyway because of the reorganization, but I still think public URLs are better.
In particular, screen readers don't necessarily detect changes if the URL doesn't change.

docs/02-v2-news.md
```diff
@@ -13,35 +13,26 @@ Papi-web est désormais livré sous la forme d'un exécutable Windows autonome.
...
 
-```
-[event]
-name = Tournoi amical 17 juin 2023
+La configuration des évènements a été réalisée à l'aide de code PHP en version 1.x, puis à l'aide d'un fichier de configuration de la version 2.0 à la version 2.3.
 
-[tournament]
-filename = amical-20230617
-```
+- À partir de la version 2.4, la configuration des évènements est réalisée à l'aide d'une interface web qui ne nécessite aucune compétence informatique.
```
Amaras:
```suggestion
À partir de la version 2.4, la configuration des évènements est réalisée à l'aide d'une interface web qui ne nécessite aucune compétence informatique avancée.
```

Amaras (commented):
Thank you for the documentation update.

We definitely need to discuss the disappearance of the public URLs, which I definitely don't agree with.

common/__init__.py
pascalaubry: Fixed.

common/__init__.py
pascalaubry: Fixed.

chessevent/event_selector.py
pascalaubry: Fixed (only occurrence).

data/loader.py
pascalaubry: Fixed.

data/loader.py
pascalaubry: This obviously needs to be used in most of the new classes.
Keep this thread open.

web/server_engine.py
pascalaubry: Fixed.

web/settings.py
pascalaubry: Feel the same, keep this thread open.

common/papi_web_config.py
pascalaubry: Indeed, fixed.

data/loader.py
pascalaubry: Fixed (your last suggestion was missing a parenthesis).
Will be even nicer later with cached_property.

data/loader.py
pascalaubry: Yes, I need to make a complete review of the new classes for this (closing this conversation since one is still opened on the same subject).

database/store.py
pascalaubry: Yeah, why not? Fixed ;-)

export.py
pascalaubry: Fixed for all the usages of PapiWebConfig.

data/event.py
pascalaubry: Well, couldn't find a better solution to replace the tmp files used previously.
Any other idea?

data/event.py
pascalaubry: Will also fix it soon.

docs/81-urls.md
pascalaubry: Not sure to understand "screen readers don't necessarily detect changes if the URL doesn't change".

Amaras: Basically this comment on Reddit: don't reinvent the wheel: https://www.reddit.com/r/htmx/comments/ti8vnj/comment/i2ukfct.
While there is a possibility to notify a screen reader of a change, the easier method of notifying a screen reader user of a change is changing the page (and the URL)

common/__init__.py
```diff
@@ -0,0 +1,45 @@
...
+
+def format_timestamp_date_time(ts: float | None = None) -> str:
+    return datetime.strftime(datetime.fromtimestamp(ts if ts is not None else time.time()), '%Y-%m-%d %H:%M')
+
+
+def format_timestamp_date(ts: float | None = None) -> str:
+    return datetime.strftime(datetime.fromtimestamp(ts if ts is not None else time.time()), '%Y-%m-%d')
+
+
+def format_timestamp_time(ts: float) -> str:
```
Amaras: Still missing the `ts: float | None` annotation

data/loader.py
pascalaubry: Fixed (used wherever possible).

data/loader.py
pascalaubry: Fixed.

common/papi_web_config.py
pascalaubry: Mark unresolved until @singleton is removed on all classes.

data/event.py
pascalaubry: Fixed.

common/papi_web_config.py
pascalaubry: @singleton replaced by metaclass=Singleton everywhere.

common/__init__.py
pascalaubry: IMO None is never passed, closing.

pascalaubry: Fixed.

web/settings.py
pascalaubry: Added sub-modules web.controllers, web.controllers.admin and web.controllers.user and renamed the classes.

COPYRIGHT.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Limites
+
+Dans tous les cas, si vous redistribuez ce produit vous devez :
+
+- le distribuer sous licence AGLP V3 (ou supérieure) et donc mettre à disposition tous les sources ;
+- faire référence au produit d'origine (Papi-web) ;
+- ne pas laisser d’ambiguïté sur l'origine du produit, qui pourrait laisser penser que vous en êtes l'auteur·ice ;
+
+Le terme distribution inclus ici toute livraison, même à un client unique, ainsi que la fourniture de services (hébergement, mode SaaS).
```
Amaras:
```suggestion
Le terme distribution inclut ici toute livraison, même à un client unique, ainsi que la fourniture de services (hébergement, mode SaaS).
```

Amaras (dismissed):
I can't review everything in this PR, so I'm not gonna delay it more.
Please make sure the next PRs are way smaller (270 files changed is definitely too much when it's not a pure documentation PR).

Issue #30, opened 2024-05-05
Check-in process is hard to use in practice


timothyarmes:
I've just arbited a small tournament for under 18s and tried out the new check-in feature for the first time.  While it worked functionally, it's actually quite awkward to use in practice. 

For context, I'm using an iPad to allow players to check-in and enter results.  While entering results presents no issues for the player, checking in requires being **very** accurate when taping the checkboxes next to the names.  On many occasions the players accidentally checked in the player above or below them, reducing my confidence in the accuracy of the pointage.  Even when it worked, it ofter took several attempts to click the right place.

I'd like to see the interface improved to follow the pattern of the results input : allow players to click anywhere on their name row, then pop up a box, with their name, to confirm or cancel the pointage.

Amaras:
Thank you for opening this issue.
The check-in feature is definitely not in its final state, as we aim to provide a better interface for the check-in process in a future version.

In the meantime, I'd recommend either zooming in to improve the accuracy of the check-in, or using ChessEvent and checking-in on the platform.

I'll keep you posted and leave this issue open until we work on the new check-in interface.

PR #46, opened 2024-11-07
Display a confirmation modal for check-ins #30

timothyarmes:


pascalaubry (approved):
Great!
Note: in a near future, I do not imagine player check-in on user input screens but on admin screens, but what you did is a real enhancement until we have a more complex authentication to allow admin screens to be used not only on the server.

timothyarmes:
I'm not sure why you would want to limit the check-in to the admin screens.

For me allowing the players to check themselves in on a tablet has been a fanstastic experience, saving me a lot of time, and avoiding hundreds of "how do you spell that" moments.

pascalaubry:
I do agree (too much time wasted) but I would like the players to be able to check-in on user screens, but for themselves only (as it is done on ChessEvent), not anybody can check for anybody.
Could you close #30?

PR #50, opened 2024-11-11
Display the licence type from Papi on admin check-in screens (#45)

### LINKED ISSUE #45: Feature request: licence check
I may be mistaken, but I not found an easy way to check that all players have the appropriate FFE licence using Papi.

This would be a really useful feature to have in Papi-Web.

timothyarmes:
Closes #45


pascalaubry (approved):
Nice.

PR #55, opened 2024-11-27
Migrate to a `pyproject.toml` project definition

Amaras:
This PR aims to migrate away from requirements.txt and use pyproject.toml instead.

The problem with the quasi-obsolete requirements.txt is clearly visible in the way JetBrains obsoleted the Requirements plugin in PyCharm.

Using a pyproject.toml is a more modern way to develop python applications, and the building backend is configurable.

One downside is that I had to push all code to the src/ directory, which might be a problem, as I did not test it yet.

The dependencies come from a run of `pipdeptree` on the virtual environment.

### Dev instructions

To install the dependencies, simply do `pip install -e .`.

I also added some development dependencies: `pipdeptree` to explore the dependency tree of our packages, and `ruff`, a Python linter and formatter.
To install those dependencies, run `pip install -e .[dev]`


Amaras:
Note that the requirements might need to be updated in the `pyproject.toml`, I did not check them thoroughly

pascalaubry:
Still not working, work in progress.

pascalaubry (approved):
Thank you Sammy, everything still works fine on Windows with your changes.

PR #58, opened 2024-12-04
Add alert messages to public screens

pascalaubry:
Add alert messages to public screens to allow arbitrers to quickly show messages to players and spectators.


timothyarmes (approved):
All looks good to me, and it's working at my end

PR #61, opened 2025-01-16
Clean data.util.Result

Molrn:
> We definitely need to clean up data.util.Result  

Following on this, here is a clean up I thought of for the enum. 
- put the papi hard coded values in a separate enum `PapiResult`
- use separate values for byes
- initialize values using opponent ids in `Result.from_papi_value`

### DIFF
--- a/src/data/util.py
+++ b/src/data/util.py
@@ -25,30 +25,47 @@ except AttributeError:
             yield batch
 
 
-class Result(IntEnum):
-    """An enum representing the results in the database.
-    Should be subclassed if the point value is not the default"""
+class PapiResult(IntEnum):
+    """An enum representing the results in the Papi database"""
     NOT_PAIRED = 0
     LOSS = 1
-    DRAW_OR_HPB = 2  # HPB = Halp Point Bye
+    DRAW_OR_HPB = 2  # HPB = Half Point Bye
     GAIN = 3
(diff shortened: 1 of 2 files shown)


src/data/util.py
```diff
@@ -25,61 +25,106 @@ def batched(iterable, n):
...
+                return cls.UNRATED_GAIN if is_unrated else cls.GAIN
+            case PapiResult.PAB_OR_FORFEIT_GAIN_OR_FPB.value:
+                return (
+                    cls.FULL_POINT_BYE if is_point_bye
+                    else cls.PAIRING_ALLOCATED_BYE if is_pairing_bye
+                    else cls.FORFEIT_GAIN)
+            case PapiResult.FORFEIT_LOSS:
                 return cls.FORFEIT_LOSS
-            case 5:
+            case PapiResult.DOUBLE_FORFEIT:
```
Amaras: Shouldn't those be `PapiResult.FORFEIT_LOSS.value` and `PapiResult.DOUBLE_FORFEIT.value` instead?

Amaras (changes requested):
Thank you for the PR.
It doesn't look too bad.
I'm requesting changes for possible type weirdness.

Amaras (approved):
LGTM

Issue #18, opened 2024-01-21
Use bbpPairings as a pairing engine


Amaras:
This is a tracking issue for the use of bbpPairings as the core pairing engine, instead of using papi.

This issue requires the use of a user interface (see #16).

Once done, this allows one to directly retrieve the pairings from the output, rather than reconstruct the tables from the papi database.

- [x] Export tournaments to TRF format (#62)
- [x] #69
- [x] #70

Molrn:
I started looking into this issue. Here is the [ToMaChess](https://github.com/Moritz72/ToMaChess) project, a python open-source project providing a UI to bbpPairings. I would suggest a similar implementation for papi-web:
- add bbp as an executable to the project
- for pairing:
  - Convert a Tournament object into a TRF(bx) file (bbp input file format)
  - generate the pairings into a another file using the bbp executable

Output format:
```
[round number]
[table 1 white player id] [table 1 black player id]
...
[table n white player id] [table n black player id]
```
@Amaras I see that you have worked on TRF support in #20 which was aborted because of lack of FIDE ID and birthdate. For a an internal use of TRF for bbp pairings, it seems possible to ignore these by:
- using any uuid instead of the FIDE ID 
- using a default birth year

Would this be an ok workaround of TRF for pairing generation, or is this issue conditioned by the presence of FIDE ID and birthdate in papi-web? If not, has this evolved since #20 has been closed ?

Amaras:
Those two problems are totally solvable, since the information is available in the papi database, I just didn't know how to interact with the database at that time.

What is not so straightforward, is that some information on the kind of results are not represented directly.
Namely, there is absolutely no way to represent an unrated game (played games, but with less than 2 moves).
Plus, you can only know if a result is a PAB, forfeit win, or full-point bye depending on the scheduled opponent (ID 1 for PAB, opponent ID for FW, None for FPB).
This is annoying to deal with, and I did not have this information at that time (only found it recently when researching for tie-breaks).

I might start work on TRF export soon-ish, TRF16 for now, and TRF25 some day (after the spec is finalised by FIDE TEC)

timothyarmes:
It would be a really interesting evolution to offer pairings.

We could start with a solution that writes the pairing results back into the Papi database since, for the moment, we still need to send the results back to the FFE site in a Papi boule.  In the future this may well evolve - if the FFE site accepts TRF then we can eventually move to our own database for everything.

Molrn:
[...]

PR #60, opened 2025-01-03
I18n

pascalaubry:
Made all the application support EN/FR and added the necessary to add future languages (6 experimental locales translated by an AI).

PR #62, opened 2025-01-16
Tournament export to TRF(bx) format

Molrn:
Related to #18 
Export of a tournament in the TRF format ([TRF guidelines](https://www.fide.com/FIDE/handbook/C04Annex2_TRF16.pdf)).  
From the fields defined in the guidelines, I tried including all the fields that were mandatory or that triggered a warning for either pairing or rating purposes. All the pairing fields are included (which should then work for bbp), but a few mandatory fields for rating are still missing:
- tournament federation (not found in Papi, could easily be added to papi-web)
- player rank in the tournament (not present in papi, I haven't yet dug into how to generate it)

On the web side, I added a download button to the tournament card in the admin part:
image

Files are downloaded in the format `{tournament name}.trf`
example of export: [TRF_export_example.txt](https://github.com/user-attachments/files/18444188/TRF_export_example.txt)

### DIFF
--- a/src/data/player.py
+++ b/src/data/player.py
@@ -1,10 +1,11 @@
 import base64
 from contextlib import suppress
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import date
 from functools import total_ordering, cached_property
 from logging import Logger
 from typing import TYPE_CHECKING, Self
+from trf import Player as TrfPlayer
 
 if TYPE_CHECKING:
     from data.tournament import Tournament
@@ -257,6 +258,23 @@ class Player:
         Otherwise, leave `self.points` as None."""
         with suppress(TypeError):
             self.points += points
+
+    @property
+    def to_trf(self) -> TrfPlayer:
+        self.compute_points(len(self.pairings))
+        return TrfPlayer(
+            self.ref_id,
+            f'{self.last_name}, {self.first_name}',
+            self.gender.to_trf,
+            self.title.to_trf,
+            self.rating,
+            self.federation,
+            self.fide_id,
+            self.date_of_birth.strftime('%Y/%m/%d') if self.date_of_birth else '',
+            self.points,
+            None,
+            [result.to_trf(round_nb) for round_nb, result in self.pairings.items()]
+        )
 
     @property
     def points_str(self) -> str:
--- a/src/data/player.py
+++ b/src/data/player.py
@@ -231,13 +231,13 @@ class Player:
     def federation_tuple(self) -> FederationTuple:
         return FederationTuple(self.federation)
 
-    def compute_points(self, max_round: int):
-        """Computes and stores the points of the player,
-        from round 1 to round `max_round` (returns None)"""
+    def compute_points(self, max_round: int) -> float:
+        """Computes and returns the points of the player,
+        from round 1 to round `max_round`"""
         # NOTE(Amaras) this does not rely on the fact that insertion order
         # is preserved in 3.6+ dict, because I can't be sure insertion order
         # is the correct (increasing) round order
-        self.points = sum(
+        return sum(
                 pairing.result.point_value
                 for round_index, pairing in self.pairings.items()
                 # NOTE(Amaras) if you were to include the current round
@@ -260,7 +260,6 @@ class Player:
             self.points += points
 
     def to_trf(self, player_id_to_trf_id: Callable[[int], int]) -> TrfPlayer:
-        self.compute_points(len(self.pairings))
         return TrfPlayer(
             startrank=player_id_to_trf_id(self.id),
             name=f'{self.last_name}, {self.first_name}',
@@ -270,7 +269,7 @@ class Player:
             fed=self.federation,
             id=self.fide_id,
             birthdate=self.date_of_birth.strftime('%Y/%m/%d') if self.date_of_birth else '',
-            points=self.points,
+            points=self.compute_points(max(self.pairings.keys())+1),
             games=[result.to_trf(round_nb, player_id_to_trf_id) for round_nb, result in self.pairings.items()]
         )
 
--- a/src/data/player.py
+++ b/src/data/player.py
@@ -4,7 +4,7 @@ from dataclasses import dataclass, field
 from datetime import date
 from functools import total_ordering, cached_property
 from logging import Logger
-from typing import TYPE_CHECKING, Self
+from typing import TYPE_CHECKING, Self, Callable
 from trf import Player as TrfPlayer
 
 if TYPE_CHECKING:
@@ -259,11 +259,10 @@ class Player:
         with suppress(TypeError):
             self.points += points
 
-    @property
-    def to_trf(self) -> TrfPlayer:
+    def to_trf(self, player_id_to_trf_id: Callable[[int], int]) -> TrfPlayer:
         self.compute_points(len(self.pairings))
         return TrfPlayer(
-            startrank=self.ref_id,
+            startrank=player_id_to_trf_id(self.id),
             name=f'{self.last_name}, {self.first_name}',
             sex=self.gender.to_trf,
             title=self.title.to_trf,
@@ -272,7 +271,7 @@ class Player:
             id=self.fide_id,
             birthdate=self.date_of_birth.strftime('%Y/%m/%d') if self.date_of_birth else '',
             points=self.points,
-            games=[result.to_trf(round_nb) for round_nb, result in self.pairings.items()]
+            games=[result.to_trf(round_nb, player_id_to_trf_id) for round_nb, result in self.pairings.items()]
         )
 
     @property
@@ -354,6 +353,11 @@ class Player:
         self.time_control_initial_time = initial_time
         self.time_control_increment = increment
         self.time_control_modified = modified
+
+    def starting_rank_comparison(self, other: "Player") -> bool:
+        return (
+            (self.rating, self.title, other.last_name, other.first_name) <=
+            (other.rating, other.title, self.last_name, self.first_name))
 
     def __le__(self, other):
         # p1 <= p2 calls p1.__le__(p2)
--- a/src/data/player.py
+++ b/src/data/player.py
@@ -263,17 +263,16 @@ class Player:
     def to_trf(self) -> TrfPlayer:
         self.compute_points(len(self.pairings))
         return TrfPlayer(
-            self.ref_id,
-            f'{self.last_name}, {self.first_name}',
-            self.gender.to_trf,
-            self.title.to_trf,
-            self.rating,
-            self.federation,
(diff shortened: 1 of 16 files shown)


Amaras:
Thanks for the PR.

First off, please update to the latest dev, since I have just merged #60.

In this case, I cannot merge this PR yet, as we need to take acceleration into account (XXA fields in TRF(bx) format).
I think it will be a case of computing a history of virtual points if applicable, then simply exporting them to TRF

Molrn:
@Amaras acceleration has been added.  

I hadn't yet looked into XX and BB fields as these are part of TRF(x) and TRF(bx). Do you think it would be interesting to implement a distinction between TRF 16, TRF(x) and TRF(bx) in the web part (like on a select input)?   

Besides acceleration, are there other XX fields that would have to be implemented?  

On a side note, xx_fields are very loose in the TRF library, such that BB fields can be implemented without any modification in the library

Amaras:
Here is the full list of TRF(x) extensions: http://www.rrweb.org/javafo/aum/JaVaFo2_AUM.htm#_Extensions_and_other_
Acceleration for round `N` must be a full history of virtual points for rounds 1 to `N` (including `N` if it still has acceleration).
bbpPairings has their own documentation I have not read in detail yet.

Plus, for pairing, `XXR` field is necessary, and `XXC black` or `XXC white` is needed for the first round

Molrn:
The export is now compliant to what BBP is expecting. I have tested it locally as a bbp input file, it outputs the same results as Papi. The main option that is missing is XXP (Forbidden pairs), which I think Papi is unable to do.

Amaras:
Thanks a lot, I'll check it when I have access to my Windows computer.
Could you actually do two exports: a pairing export (which you already did), and one rating export?
We'll make it compliant with TRF16 when we have all the information we need (I think I'll need to start again from scratch on tie-breaks, there have been a lot of modifications to data layout).

Technically, Papi can do forbidden pairs using family and club protections, although I have not found the internal pairing engine code that is responsible for forbidden pairs.

Molrn:
The blocking point for rating is mostly the player rankings. They are supposed to be in the papi database, the columns `Dep1`, `Dep2` and `Dep3` representing the tie-breaks and `Place` the rank, but these always appear empty in all the tournaments I look into.  

Unfortunately, I think this part might require a full implementation of Tie-breaks into papi-web. I have looked into tools such as BBP or JaVaFo that handles ranking/tie-break, I haven't found an equivalent but I encountered the project [marceljuenemann/tiebreak](https://github.com/marceljuenemann/tiebreak) which might be interesting to look into.  

For now, I suggest to implement the distinction between rating and pairing like this:
- using the extension `.trfx` for pairing and `.trf` for rating
- removing the `XX` and `BB` fields from the rating file
- use the following dropdown button in the UI  

image

Amaras:
Perfect.

Yeah, Papi computes them each time it is asked to, i.e. when it is asked for rankings, cross tables, or prizes (and the FFE server does as well each time it receives a Papi file), so you can't rely on them.

I'll look into the link you sent quickly.

I think what I asked could become a tracking issue in itself, so that we keep track of it.
We don't need to have the rating export just yet, since we don't have the tie-breaks.

Amaras:
Alright, I'll check once you merge the latest changes to dev.
Sorry to push it back once again

Amaras:
I'm getting a parsing error on the TRF exported for Avoine 2024.
I think it's linked to the WFM title taking too much space and not aligning to the other titled lengths

```
./bbpPairings.exe --dutch Avoine\ 2024.trf -c
Error parsing file Avoine 2024.trf: Invalid line "001   72 f WIM ROUMEGOUS, Marina                 1993 FRA      620416 1964/08/28  4.5         12 b =    22 w -   112 w 1   158 b 0   168 w 1   114 b 0   272 w 1   190 b 1   142 w 0"
```
This is linked to the title field not being aligned.

When correcting this error, there is this new error:
```
./bbpPairings.exe --dutch Avoine\ 2024.trf -c
Error parsing file Avoine 2024.trf: Invalid line "XXA    2 0.5:>4 0.0:>4 0.0:>4 0.0:>4 0.0:>4 0.0:>4 0.5:>4 0.0:>4 0.0:>4"
```
I don't think the values are correct, additionally

Molrn:
> 0.5:>4  

Parsing error fixed in the latest version  

> I don't think the values are correct, additionally  

I'll have to check this, I'm not sure what the acceleration is supposed in this case. I mostly adopted the existing function for computing virtual points. If you have more details about what is wrong about it it would really help.

Amaras:
The player IDs are wrong in the TRF, because you are using different IDs for the export:
you are using the starting rank correctly in the player list, however you are using the Papi `ref_id` in the acceleration lists

Molrn:
fixed, thanks :)

Amaras:
I think you should use the latest version of our TRF fork (aka the `master` branch), I fixed the wrong title field length on our branch

Molrn:
Something did not add up on this, so I looked up into it and actually changed my mind:
image  
extract from the [TRF 16 handbook](https://www.fide.com/FIDE/handbook/C04Annex2_TRF16.pdf)  

Gender is supposed to be spelled on one char (char 10) and title on 3 char from char 11 to 13. But unlike every other field, there is no space in between (char 11 should be empty). The examples I found of TRF files ([example 1](http://www.rrweb.org/javafo/aum/AcceleratedTRFXSample2.txt), [example 2](https://ratings.fide.com/download/example1.txt)) are using the same 2 character title definition as in papi (g, m, f, c, gf, mf, ff, cf).  

I think it is more of an ambiguity in the FIDE handbook, so I fixed it by using papi titles. Does it feel ok to use these ?

Amaras:
Indeed, I did not check that thoroughly enough…
Would the sex+title look like `m GM`, `wWGM`, and so on?
Feels weird.

Looking up the TRF25 final draft, it looks like it's the same thing.
I think it's a specification bug, but I should email the TEC commission to make sure.

On a related note, I think we should introduce back the unused “(Woman) Candidate Master” titles, since they're somewhat useful for players' egos.

Amaras:
`XXA  124  2.0  1.0  1.0  1.5  1.5  1.5  1.5  0.0  0.0`
I am pretty sure it's not possible to have those acceleration values in this order.
Are you relying on rounds being in sorted order?

Molrn:
I am not, this is how the history is generated:  
```python
vpoints_history = [
    self._calculate_player_virtual_points(player, round_nb)
    for round_nb in range(1, self._current_round+1)
]
```
Tracked the error, points per round were not computed at the right time. Here is the new acceleration for this player: `XXA  124  1.0  1.0  1.5  1.5  1.5  1.5  2.0  0.0  0.0`  

Let's check that these are the accurate values:  
`001  124 m    VEKEMANS, Denis                   1897 FRA      615846 1968/07/29  4.0        192 b 1   216 w 1    14 b 0   191 w =    34 b 0   265 w =   187 b 0   255 w 1   225 b 1`
Player is in group B.  
After round 2, he has 2 points -> vpoints = 1+potential_vpoints = 1+0.5 * (**2** // 1.5)  = 1.5. 
After round 6, he has 3 points -> vpoints = 1+0.5 * (**3** // 1.5) = 2.0
This seems to be accurate

Amaras:
Makes much more sense.
I'll check it a bit later today

Amaras:
```
C:\Users\Sammy>.\bbpPairings.exe --dutch "Downloads\Avoine 2024 (1).trfx" -c
Error parsing file Downloads\Avoine 2024 (1).trfx: The score for player 1 does not match the game results.
```
TRF has player 1 with 7.0 points, but 7.5 from his game history

Note that `Player.compute_points` compute points *before* the given round number.
To have points after, you need to compute for `Tournament.rounds + 1`

Molrn:
The tournament is now correctly parsed by the bbp checker. Problem is it spots differences during pairings, which I have not been able to identify the source(s) of.  

I identified one factor though: the first round pairing is black, not white as defaulted in the export (`XXC white1` --> `XXC black1`). Sidenote: I think this value is not stored in the papi DB, i.e. the use of a default value

Amaras:
In the case of Avoine 2024, I don't remember exactly, but: there might have been a change in one score group given after the pairing of round 2 was made, but before the pairing of round 3, which changes the whole score group.
There might have been permutations (people playing with the correct opponent, but wrong colours).
Papi also doesn't use the exact same algorithm as bbpPairings, so I think all 4 explanations cover 95-99% of all cases

src/data/player.py
```diff
@@ -230,13 +231,13 @@ def league_tuple(self) -> LeagueTuple:
     def federation_tuple(self) -> FederationTuple:
         return FederationTuple(self.federation)
 
-    def compute_points(self, max_round: int):
-        """Computes and stores the points of the player,
-        from round 1 to round `max_round` (returns None)"""
+    def compute_points(self, max_round: int) -> float:
```
Amaras: Not sure it's correct to change the API like this.
I think you should add a helper function (named something like `points_before`) and use it in `compute_points`

src/data/tournament.py
```diff
@@ -384,6 +421,63 @@ def print_real_points(self) -> bool:
...
+        for result in [
+            result_class.GAIN,
+            result_class.DRAW,
+            result_class.LOSS,
+            result_class.FORFEIT_LOSS,
+            result_class.PAIRING_ALLOCATED_BYE,
+            result_class.ZERO_POINT_BYE,
+        ]:
+            xx_fields[result.bbp_field] = f'{result.point_value:>4}'
+        return xx_fields
```
Amaras: I added `bb_fields` to our TRF fork, so you should use those instead, since we'll need to parse them when importing (required by FIDE for approval)

src/data/util.py
```diff
@@ -524,6 +578,18 @@ def to_papi_value(self) -> str:
             case _:
                 raise ValueError(f'Unknown value: {self}')
 
+    @property
+    def to_trf(self) -> str:
+        match self:
+            case PlayerGender.NONE:
+                return ''
+            case PlayerGender.FEMALE:
+                return 'f'
```
Amaras: The TRF16 format is `m/w`, not `m/f`

Amaras (changes requested):
A few nitpicks, the rest is good to me

src/data/tournament.py
Molrn: bb_fields support is lacking one feature: dumping into the TRF file. Here's a PR to fix it: https://github.com/papi-web-org/TRF/pull/1. Another release of the TRF fork might be cleaner too

Amaras: TRF v1.1.3-unofficial now out https://github.com/papi-web-org/TRF/releases/tag/v1.1.3-unofficial

Amaras (approved):
LGTM

PR #64, opened 2025-01-17
fix: gettext module override

Molrn:
Another error I ran into after the update of #60: as there is a function with the same name as the module `gettext`, it caused ambiguity when the module was used. It only occured when there was no locale defined, which is probably why it was not causing issue on your end.  

Other than that, the dev branch runs fine on my end. One small issue was that I had to manually create a `Locale` repository in my virtualenv, but I haven't found a clean way to prevent that from happening. @pascalaubry maybe this would be worth looking into

### DIFF
--- a/src/common/i18n.py
+++ b/src/common/i18n.py
@@ -1,4 +1,4 @@
-import gettext
+import gettext as gettext_lib
 import sys
 import threading
 from gettext import GNUTranslations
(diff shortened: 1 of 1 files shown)


pascalaubry (approved):
Nice.
Just don't understand how your locale/ could be missing.

PR #65, opened 2025-01-17
Add the players' category to the Players tab

pascalaubry:
Added a new column to the list of players with the players' category.


src/database/sqlite.py
```diff
@@ -762,6 +762,12 @@ def _upgrade(self):
             self.set_version(target_version)
             self.commit()
             logger.debug(f'Database %s has been upgraded to version %s.', self.file.name, target_version)
+        target_version = Version('2.4.21')
+        if self.version.public in ['2.4.20', ]:
+            self._execute('DROP TABLE `skipped_round`')
+            self.set_version(target_version)
+            self.commit()
+            logger.debug(f'Database %s has been upgraded to version %s.', self.file.name, target_version)
```
Amaras: This should probably be an arm of a match statement, judging from the architecture of the targets...

src/database/sqlite.py
```diff
@@ -1220,10 +1182,7 @@ def load_stored_tournaments(self) -> list[StoredTournament]:
             'SELECT * FROM `tournament` ORDER BY `uniq_id`',
             (),
         )
-        stored_tournaments: list[StoredTournament] = [self._row_to_stored_tournament(row) for row in self._fetchall()]
-        for stored_tournament in stored_tournaments:
-            stored_tournament.stored_skipped_rounds = self.load_stored_skipped_rounds(stored_tournament.id)
-        return stored_tournaments
+        return [self._row_to_stored_tournament(row) for row in self._fetchall()]
```
Amaras: Does it need to be a list, or can the consumer transform it into a list?
Assuming appropriate type hints, it would look something like this:
```suggestion
        yield from map(self._row_to_stored_tournament, self._fetchall())
```

Amaras (commented):
Without running the code, it looks pretty good.
I'll try to run it soon-ish before approving it

src/database/sqlite.py
pascalaubry: Much nicer indeed, last commit of the branch (tested for 2.4.0 databases).

src/database/sqlite.py
pascalaubry: Methods load_stored_*() are all consumed by load_stored_event() and the results are stored as lists (the goal is to load data at once, store it for future use, and close the the event database as soon as the data is read).
Thus list conversion can be done when storing the data in the StoredEvent or in the load_stored_*(), but it must be done.
Used iterators wherever possible in the last commit.

PR #66, opened 2025-01-17
Update pyproject.toml when exporting the application

pascalaubry:
Write the version number to pyproject.toml when creating the EXE file


Amaras (approved):
I'd like to rely on setuptools' dynamic version capabilities, but this is okay in the meantime

PR #68, opened 2025-01-18
Improved message management

pascalaubry:
Removed the MO/POT files from the repository
Removed the line numbers in POT/PO files
Detect locale errors on startup and exit (except for developers)
For developers, check that MO files are up-to-date


src/common/i18n.py
```diff
@@ -18,21 +19,38 @@
...
+            except Exception as ex:
+                print_interactive_error(f'Could not load locale [{l}]: {ex}.')
+                locale_error = True
+                if not DEVEL_ENV:
+                    sys.exit()
+        else:
+            print_interactive_error(f'Invalid locale [{l_entry.name}], MO file [{mo_file}] not found.')
+            locale_error = True
+            if not DEVEL_ENV:
+                sys.exit()
```
Amaras, line 53: What happens if `l_entry` is not a directory?
This should be an error as well

Amaras (commented):
Looks good, I just need the clarification on my comment

src/common/i18n.py
pascalaubry: Then l_entry is probably a file, most probably messages.pot, so not a locale.
Were you thinking of symlinks?

pascalaubry: Not sure it is worth checking this, but I can if you wish.

Amaras: No, makes sense.
I'll approve and merge

Amaras: No, makes sense.
I'll approve and merge

Amaras (approved):
LGTM

PR #72, opened 2025-01-20
Manually add players to the tournaments

pascalaubry:
At this step, adding a player can be done manually only, by providing all the needed information.
This feature is not enough to fully manage the tournaments but it can be tested as-is and merged if correct.
Adding players by relying on a local FIDE database (next step 1), on a local FFE database (next step 2) or on the central FFE server (next step 3) will be done in other branches.


src/data/event.py
```diff
@@ -147,6 +147,37 @@ def start(self) -> float:
     def stop(self) -> float:
         return self.stored_event.stop
 
+    @property
+    def federation(self) -> str:
+        return self.stored_event.federation
+
+    def set_player_default_ratings(self, player: Player):
```
Amaras, line 154: We should think about a proper (i.e. not hardcoded) way to define default ratings.
For now, it's okay since we are focused on FFE use.

src/database/store.py
```diff
@@ -177,6 +177,12 @@ class StoredEvent:
     stored_rotators: list[StoredRotator] = field(default_factory=list[StoredRotator])
     errors: dict[str, str] = field(default_factory=dict[str, str])
 
+    @property
+    def federation(self) -> str:
+        """Returns the federation of the event (used by TRF exports)."""
+        # TODO Store the federation od the event in the database from version 2.4.21, for the moment assume that all the events are organized in FRA ;-)
```
Amaras:
```suggestion
        # TODO Store the federation of the event in the database from version 2.4.21, for the moment assume that all the events are organized in FRA ;-)
```

Amaras (dismissed):
Apart from the nitpick, looks good to me

Amaras (approved):
Of course it's dismissed when I commit something afterwards...
LGTM

src/data/event.py
pascalaubry: +1

PR #73, opened 2025-01-21
Added the local FIDE database

pascalaubry:
The local FIDE database (tmp/fide.db) will be used to retrieve information from the FIDE database when adding or updating players.
ServerEngine checks for the existence and the obsolescence of the FIDE database on startup and proposes to createupdate the database.
Script download_federation_flags now also relies on the database.


pascalaubry:
Overridden by #74

PR #74, opened 2025-01-21
Added the local FFE and FIDE databases 

pascalaubry:
The local FFE database (tmp/ffe.db) will be used to retrieve information from the FFE database when adding or updating players.
ServerEngine checks for the existence and the obsolescence of the FFE database on startup and proposes to create/update the database.
FideDatabase and FfeDatabase both have a search_player() method that will be used to add players to tournaments.


src/data/util.py
```diff
@@ -944,6 +956,30 @@ def to_papi_value(self) -> str:
...
+                return PlayerTitle.INTERNATIONAL_MASTER
+            case 'WGM':
+                return PlayerTitle.WOMAN_GRANDMASTER
+            case 'GM':
+                return PlayerTitle.GRANDMASTER
+            case _:
+                raise ValueError(f"Unknown title value: {value}")
+
     @property
     def to_trf(self) -> str:
```
Amaras, line 984: You're missing CM and WCM titles there

src/database/store.py
```diff
@@ -180,7 +180,7 @@ class StoredEvent:
     @property
     def federation(self) -> str:
         """Returns the federation of the event (used by TRF exports)."""
-        # TODO Store the federation of the event in the database from version 2.4.21, for the moment assume that all the events are organized in FRA ;-)
+        # TODO Store the federation od the event in the database from version 2.4.21, for the moment assume that all the events are organized in FRA ;-)
```
Amaras:
```suggestion
        # TODO Store the federation of the event in the database from version 2.4.21, for the moment assume that all the events are organized in FRA ;-)
```

Amaras (changes requested):
Overall, looks good to me, apart from the slight problems I commented on.
Just two errors while parsing the FIDE list:
Francisco J. Cruz Arce (310751) has LSI,FI,SI,SLI,NA titles, which means that it probably overflows into the rating columns (`invalid literal for int() with base 10: 'A1946'`)
Corry Vreeken (1495045) has Honorary Woman Grandmaster title (WH) since 1987, which is not known in our system, and thus she is ignored during parsing.
This is something to take into account.

src/database/store.py
pascalaubry: Fixed on the repository.

src/data/util.py
pascalaubry: Fixed on the repository.

Amaras (approved):
LGTM

PR #75, opened 2025-01-22
Simplify i18i by removing useless i18n-*.md files

pascalaubry:
Removed useless files docs/i18n-*.md (previously automatically generated by i18n_update.py), only 86-i18n.md is kept.


Amaras (approved):
LGTM

PR #77, opened 2025-01-22
feat: pairings generation with bbpPairings

Molrn:
related to #18  

An implementation of pairings generation with the most minimal UI. BbpPairings is downloaded from the release when executing `pip install .`. The generated pairings are written back into the papi database.  
Pairings can only be generated if the tournament is ongoing and if the last round is over. In term of UI, a button appears if the pairings can be generated on the tournament card:  
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ __pycache__/
 /test/
 /export-test/
 /utils/i18n/models
+/resources/
 
 *.egg-info/
 
--- a/locale/de/LC_MESSAGES/messages.po
+++ b/locale/de/LC_MESSAGES/messages.po
@@ -4903,6 +4903,15 @@ msgid "Edit the properties of the tournament."
 msgstr "Bearbeiten Sie die Eigenschaften des Turniers."
 
 #: src/web/templates/admin_tournament_card.html
+msgid "Generate pairings for the next round."
+msgstr ""
+
+#: src/web/templates/admin_tournament_card.html
+#, ai_translation, fuzzy
+msgid "Pair"
+msgstr "Paarungen"
+
+#: src/web/templates/admin_tournament_card.html
 #, ai_translation, fuzzy
 msgid "Export"
 msgstr "Hafen"
--- a/locale/el/LC_MESSAGES/messages.po
+++ b/locale/el/LC_MESSAGES/messages.po
@@ -4905,6 +4905,15 @@ msgid "Edit the properties of the tournament."
 msgstr "Επεξεργασία των ιδιοτήτων του τουρνουά."
 
 #: src/web/templates/admin_tournament_card.html
+msgid "Generate pairings for the next round."
+msgstr ""
+
+#: src/web/templates/admin_tournament_card.html
+#, ai_translation, fuzzy
+msgid "Pair"
+msgstr "Ζευγάρωμα"
+
+#: src/web/templates/admin_tournament_card.html
 #, ai_translation, fuzzy
 msgid "Export"
 msgstr "Λιμένας"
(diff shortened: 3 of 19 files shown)


Amaras:
> BbpPairings is downloaded from the release when executing `pip install .`

I don't see this behaviour when doing `pip install -e .`, or even `pip install .`
I needed to run `python setup.py install`.

I think there needs to be either more documentation on that, or a console script to install bbpPairings.
I'll also want input from @timothyarmes for the bbpPairings installation, as he's on Mac.

Other than that, there seems to be pairing quirks, I'll need to investigate them.
The rest looks good.

Amaras:
Found the quirk: players who are supposed to have a ZPB are paired nonetheless.

Before pairing: Pascal Aubry (among others) should have a ZPB
Before pairing: Pascal Aubry is supposed to have a ZPB (as are others)

After pairing: Pascal Aubry is paired
After pairing: Pascal Aubry is paired (as are others)

timothyarmes:
I can't offer any feedback on the Mac version since it's currently impossible to install PapiWeb on a Mac until the Access database requirement is removed. I'm forced to run PapiWeb on a Windows VM.

However, for the Access-free future we're clearly going to need a Mac build.

Fortunately BBPairings looks like a dead project (or one that's simply completely bug free and needs no updates), so we don't need to worry too much about regular updates. I'll be able to build a Mac version that we can host ourselves when the time comes, so this isn't an issue to be concerned about.

Molrn:
> players who are supposed to have a ZPB are paired nonetheless.  

fixed ;)   

I am not familiar with the macos environment, I though tried running bbp in an Ubuntu environment without success.   

> I don't see this behaviour when doing pip install -e ., or even pip install .  

I have no idea what could cause this, I have tried running `pip install .` in different environments, with several pip versions, even in ubuntu, the file is correctly downloaded to `resources/bbpPairings-v5.0.1` every time. It is supposed to happen at the end of the install. To investigate this, could you please provide me more information about your environment ?

Molrn:
> I though tried running bbp in an Ubuntu environment without success  

Update: does not work in ubuntu 22.04, works in 24.04

Amaras:
Not fixed :(

Before pairing: Pascal Aubry, Alain Goursaud, Patrick Normand and Matei Ruellou are supposed to have a ZPB.
2025-01-23 15 18 00 127 0 0 1 f1c12266da8e


After pairing: all the players cited above are paired.
2025-01-23 15 18 13 127 0 0 1 52b779420e16

Amaras:
The fix works.
However, PAB are not shown in the input interface when assigned by bbpPairings.

SCHKLARTSCHIK Liliane has a PAB (given an X in her history)
2025-01-23 20 39 56 127 0 0 1 d6ab2bdeae35

She's not shown as having a PAB
2025-01-23 20 40 02 127 0 0 1 af015e351d52

Amaras:
Regarding the PAB not being visible, this apparently comes from the i18n branch being merged in #60.
I'll have to investigate it further, but the branch was deleted (I'll try to see if it can be restored now).
If you can fix the bug in this PR, it would be wonderful, but it can be done in a separate PR.

One weird behaviour with respect to your PR is that the URL of the event's tournaments tab becomes visible to the end user, which is not supposed to happen.
I think your `ClientRedrect` is the cause of that.

Amaras (commented):
alright, everything in pairing looks good.
One thing we forgot to do is bundling the bbpPairings executable in the release process (see `utils/export/export.py`).
Do you wish to put it in this PR, or should we do another one?

Molrn:
@Amaras I started looking into it, so far the regular export is not running on my end. If that's ok with you I'd prefer to do this in another PR

Amaras (approved):
Since we're pushing the export back to another PR, LGTM

PR #78, opened 2025-01-24
Use the XML FIDE export to build the FIDE local database

pascalaubry:


Amaras (approved):
LGTM and it works.
It takes a long time to record players in the database, but I don't think we can speed up XML parsing without changing languages or using parallelism tricks

PR #80, opened 2025-01-26
Reorganize templates

pascalaubry:
Last thing missing in this reorganization IMO is to move (/src/web/templates, /src/web/static) to (/web/templates, /web/static) to leave only python files in /src.


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools >= 75.5.0", "requests ~= 2.32.3",]
+requires = ["setuptools >= 75.8.0", "requests ~= 2.32.3",]
```
Amaras: I'll allow this only if you can point me to a security fix that justifies using the absolute latest version for everyone

src/web/templates/admin/screens/screen_modal.html
```diff
@@ -312,27 +312,27 @@ <h3>
                         {% if screen_type == 'input' %}
                             <div class="row">
                                 <div class="col-6">
-                                    {% include 'admin_screen_modal_input_exit_button.html' %}
+                                    {% include '../common/screen_or_family_or_rotator_modal/input_exit_button.html' %}
```
Amaras: I don't like this deeply nested hierarchy, especially with really long names.

With `players_show_unpaired.html`, that's the 7th level of nesting, and a path 92 characters long.

Why couldn't you simply do `admin/common/input_exit_button.html`?
With relative paths where needed, of course.

Same thing for `message.html` (maybe change that name in this case), `players_show_unpaired.html`, `edit_timer_colors.html`, and `edit_timer_delays.html`.

If we really need more separation between templates in common, we do that then

Amaras (changes requested):
Thanks for the reorganization.
One minor gripe (setuptools), one major (the 92 characters long path, see comment).

pyproject.toml
pascalaubry: No real need for this, just try to stick to the latest stable version of all our libs.
Back to 75.5.0 if you prefer.

src/web/templates/admin/screens/screen_modal.html
pascalaubry: Shortened a bit some filenames but I insist on keeping readable filenames event though they may seem long.

pascalaubry: Really shortened the names on https://github.com/papi-web-org/papi-web/commit/74a98c8bd92d31adcd9a731317d8bee0a85ed105 on branch factorize_template_code (factorizing inputs code made obvious that the folder names were becoming loger than the file names).

pyproject.toml
Amaras: Since we have a `>=` in the version specification, pip should install the latest version anyway.
As long as we have a version that does what we need and that is secure, we don't need to have the latest version for the build system

Amaras (approved):
LGTM

PR #81, opened 2025-01-26
Replace `setup.py` usage for BBP Pairings installation

Molrn:
> I don't see this behaviour when doing `pip install -e .`, or even `pip install .`
I needed to run `python setup.py install`.  

@Amaras I overlooked this part of the review of #77, only focused on making `pip install .` work. I tried making it work in development mode (-e), and realized it was not supported ([source](https://setuptools.pypa.io/en/latest/userguide/development_mode.html#legacy-behavior)). As using setup.py as a command line tool is deprecated ([source](https://packaging.python.org/en/latest/discussions/setup-py-deprecated/)), I think replacing setup.py as a whole is necessary.  

I suggest the following change, similar to the local FFE / FIDE databases recently added:  
- store the bbp executable in `tmp`
- at server launch, check if it is already installed, if not install
- disable related features (pairings generation) if BBP is not installed  

This would also solve the need to include bbp into the export.  

If not satisfying, I'll move to a documented script-based install included in the export as previously suggested, I wanted to suggest this solution before


Amaras:
I'm not fond of downloading an executable at runtime, that could be a source of security issues.

Downloading before export is only needed when the executable is not present already.
Maybe this could be a console script, although I don't know how that works exactly.

timothyarmes:
I also doubt that we'll get away with that for the Mac distribution, where we'll have to sign everything with a developer's account.

Amaras:
Cryptographic signature is definitely something we should start to worry about, since many people see their antivirus reject our releases.
We probably need a CI pipeline for automated testing, bundling, signing, etc. but we can wait a few months before working on it

src/web/server_engine.py
```diff
@@ -50,6 +51,7 @@ def __init__(self):
             print_interactive_error(_('Error while updating the FIDE database.'))
         if not FfeDatabase().check():
             print_interactive_error(_('Error while updating the FFE database.'))
+        BbpPairings().check_installed()
```
Amaras: This should only blow out in production.

In a development environment, it should be a warning instead.

src/common/bbp_pairings.py
```diff
@@ -1,51 +1,76 @@
...
+    LINUX_BUILD = "x86_64-pc-linux.tar.gz"
+
+    @property
+    def executable_path(self) -> Path:
+        return BASE_DIR / 'resources' / f'bbpPairings-{self.VERSION}' / 'bbpPairings.exe'
+
+    def check_installed(self):
+        """Check if BBP Pairings is installed, raise if it is not."""
+        if not self.executable_path.exists():
+            raise FileNotFoundError(_(
```
Amaras: I don't think this should blow out.
You should rather use the logs (and not translate), or use one of the `print_interactive_*` functions (and translate).

I think it'd better only in English, and not translated

Amaras (changes requested):
Slight changes I'd like to see:
Exception messages should probably not be translated; not having bbpPairings is only critical in production, not in dev.

Amaras (approved):

PR #82, opened 2025-01-27
Factorize template code and use permanent sessions

pascalaubry:
Use *_input() macros to lighten the HTML modals
Change the session backend to get permanent states


src/web/templates/admin/common/macros.j2
```diff
@@ -0,0 +1,539 @@
...
+            {% set _class='text-secondary' %}
+        {% elif pairing.full_point_bye %}
+            {% set _text='+' %}
+            {% set _class='text-secondary' %}
+        {% elif pairing.forfeit_loss or pairing.double_forfeit %}
+            {% set _text='<' %}
+            {% set _class='text-secondary' %}
+        {% elif pairing.forfeit_gain %}
+            {% set _text='>' %}
+            {% set _class='text-secondary' %}
```
Amaras, line 95: Should we continue with those, or should we use the TRF results strings?
This could require some getting used to from the part of users.

Amaras (commented):
LGTM, I'll approve when I have an answer to my question

src/web/templates/admin/common/macros.j2
pascalaubry: There are several ways to display the history of the players, and all of then have their pros and cons:
- The actual Papi-like way, easy for Papi users
- TRF for non French-speaking users
- Any translation of the TRF solution
My opinion:
- Today, we should keep the Papi-like way, not to disturb Papi users.
- Tomorrow, multiple solution should be available to the users and the choice between the solution should be done via a setting at user, event, federation or application level (for the federation, use the federation declared at event-level).
Anyway, the last point should not be discussed here but rather on seperate (WA?) thread.
@Amaras I added an item to the TODO list, OK with this?

Amaras: Definitely okay with this, I wanted this discussed.

I might prefer `F` (upper case) rather than lower case, but it's good enough for now

Amaras (approved):
LGTM

PR #91, opened 2025-01-30
Hide pairings and TRF export features from production environment

Molrn:


Amaras (approved):
LGTM

pascalaubry:
Just wonder if it wouldn't be easier to pass a devel variable to Jinja2 not to pass other variables for all the coming features.

Amaras:
> Just wonder if it wouldn't be easier to pass a devel variable to Jinja2 not to pass other variables for all the coming features.

I thought about passing an environment variable like `PAPI_WEB_EXPERIMENTS`, which could bring be used to define: `use_experiments = DEVEL_ENV or os.env.get("PAPI_WEB_EXPERIMENTS", False)` (up to syntax correctness)

EDIT: or maybe it could be a config option

pascalaubry:
First: I think we should dissociate:
1. "being in a devel environment": already done with DEVEL_ENV with the program acting differently when DEVEL_ENV is True or False;
2. "using the experimental features": not done yet and I agree with an environement variable (my preference for the name is EXPERIMENTAL)
Last: for these two different things, we should use two different context params, named devel and experimental.
NB: not fan of a config option, and I think papi-web.ini.[i18n].experimental should be removed.
I open a new branch on this.

PR #92, opened 2025-01-30
Removed the --use-fuzzy for the untrusted locales

pascalaubry:
Everything in the title, to solve the generation of approximative translations by the AI for untrusted locales.


Amaras (approved):
Makes sense to me

PR #118, opened 2025-02-14
Add keyboard navigation

timothyarmes:


pascalaubry (approved):
Just saw that the filters of the header of the Players table should be reached by the TAB key, anyway it works fine for me.

PR #119, opened 2025-02-14
Fix Infinite recursion error

Amaras:
I fixed a nasty infinite recursion error that made the software unusable on events with players.

I also moved estimation to `data.tournament.Tournament` rather than `tie_breaks/individual.py`

I made sure that we can enter events, so this should be mergeable soon.


timothyarmes (dismissed):

pascalaubry (approved):
Infinite recursion disappeared.

PR #122, opened 2025-02-15
Use client-side sessions

pascalaubry:
I had problems with server-side sessions and suppressed them by using client-side (for few days without any problem).
This is a non minor change that you may want to discuss.


pascalaubry:
Still having these exceptions when editing players the first time the modal is loaded with server-side sessions, don't you?
image

Amaras:
I don't see that on my end, federation flags don't seem to be loaded in the player modal (on dev, I refreshed without cache).

pascalaubry:
Federation inputs were broken, repared by #150.
Can you re-test please?

Amaras:
I confirm I still have the problem in #150.
I think we can merge this

Amaras (approved):

PR #125, opened 2025-02-15
Added the reload button to the header of the Players table

pascalaubry:
The reload button disappears when scrolling on the Players table, adding a reload icon to the sticky header makes the reload action available even at the bottom of the page.


Amaras (approved):

timothyarmes:
There's no point having two reload buttons though...

pascalaubry:
I do agree, the big button could be deleted IMO.
@Amaras no problem with that?

PR #126, opened 2025-02-16
feat: Add player ranks to TRF

Molrn:


src/data/tournament.py
```diff
@@ -106,6 +106,9 @@ def __init__(
         self._arbiter: str = ''
         self._boards: list[Board] | None = None
         self._unpaired_players: list[Player] | None = None
+        self._tie_breaks: list[TournamentTieBreak] = [
+            TournamentTieBreak.NONE,
+        ] * 3
```
pascalaubry: 4 tie-breaks at least are needed, maybe we should have 5.

src/data/tournament.py
```diff
@@ -612,6 +627,7 @@ def read_papi(self, update: bool = False):
         self._papi_read = True
         self._calculate_current_round()
         self._set_players_illegal_moves()  # load illegal moves for the current round
+        self._set_player_ranks()
```
pascalaubry: Isn't this rank calculation too heavy to be done systematically?
As this value is not needed on all screens, shouldn't it be calculated on-demand (for all the players of a tournament when needed for one player of the tournament)?

src/data/tournament.py
```diff
@@ -376,14 +392,11 @@ def players_by_fide_id(self) -> dict[int, Player]:
...
-            Player.__le__ = lambda self_, other: self_.starting_rank_comparison(other)
-            ordered_players = sorted(players, reverse=True)
-            return {index + 1: player for index, player in enumerate(ordered_players)}
-        finally:
-            Player.__le__ = le_method
+        ordered_players = sorted(
+            self.players_by_id.values(),
+            key=lambda player: player.starting_rank_sort_key,
+        )
+        return {index + 1: player for index, player in enumerate(ordered_players)}
```
Amaras:
```suggestion
        return {index: player for index, player in enumerate(ordered_players, start=1)}
```

src/data/util.py
```diff
@@ -569,6 +572,64 @@ def to_papi_value(self) -> str:
...
+            case TournamentTieBreak.BUCHHOLZ:
+                return individual.buchholz(
+                    player, tournament, max_round=max_round, papi_legacy=True
+                )
+            case TournamentTieBreak.BUCHHOLZ_CUT_TOP:
+                return individual.buchholz(
+                    player,
+                    tournament,
+                    max_round=max_round,
+                    cut_top=1,
```
Amaras: There is no cut top, it's always a bottom or median cut.
Also, in papi, the number of cut contributions depends on the number of rounds

src/data/util.py
```diff
@@ -569,6 +572,64 @@ def to_papi_value(self) -> str:
...
+                return individual.wins(
+                    player, tournament, max_round=max_round,
+                )
+            case TournamentTieBreak.KOYA:
+                return individual.koya(player,tournament, max_round=max_round)
+            case TournamentTieBreak.SONNENBORN_BERGER:
+                return individual.sonneborn_berger(
+                    player, tournament, max_round=max_round,
+                )
+            case TournamentTieBreak.CUMULATIVE, TournamentTieBreak.KASHDAN:
```
Amaras: Cumulatif == Progressive Scores, it's implemented.

src/data/tournament.py
Amaras: It might be, I have not profiled it.
It should probably be used in a cached property (namely `players_by_rank`).

src/data/util.py
```diff
@@ -569,6 +572,64 @@ def to_papi_value(self) -> str:
...
+                    cut_top=1,
+                    papi_legacy=True,
+                )
+            case TournamentTieBreak.BUCHHOLZ_CUT_TOP_BOTTOM:
+                return individual.buchholz(
+                    player,
+                    tournament,
+                    max_round=max_round,
+                    cut_top=1,
+                    cut_btm=1,
```
Amaras: Those two depend on the number of rounds of the tournament.

src/tie_breaks/individual.py
```diff
@@ -382,8 +381,7 @@ def sum_of_buchholz(
     """Computes the sum of Buchholz scores of the opponents before *max_round*
     If *max_round* is not provided, it will be set to the maximum round index
     of the player.
-    If *fore_modifier* is True, will use Fore Bochholz instead of total Buchholz.
-    If *papi_legacy* is True, will use the backwards compatible computation."""
```
Amaras: Ah, I evidently forgot to implement it

Amaras (changes requested):
Looks pretty good, but there are some things that are totally wrong

src/data/tournament.py
Molrn: do they need to be limited ? If we do not stick to the 3 in Papi, and if there is no official limit, I think we could have as many as wanted by the user.

Amaras: At some point, it's diminishing returns.
6 is already too much in my opinion (what kind of tie-breaks are you going to use 6 levels deep if you already use a performance tie-break before?)

pascalaubry: @Amaras Checked on Tornelo: we had tournament with 4 tie-beaks (including DE at first).
I think we should store the tie-breaks as a JSON list in the database and not limit the number of tie-breaks.
If a limit should be set, I think 5 is a good choice.

Amaras: If we're doing a JSON list in the database, maybe we should store the keyword arguments (namely the modifiers) in that list as well.

Molrn: We defiinitely need more than Papi's one word description. In terms of UI, I think a sortable accordion list would be a correct way to represent this (header: tie break name, content: tie break options), with a global checkbox 'Papi legacy'.

src/data/util.py
Molrn: I'm a bit confused by what these enum values are supposed to be, is it:
- BUCHHOLZ_CUT_TOP --> cut_top, cut_btm
- BUCHHOLZ_CUT_TOP_BOTTOM --> cut_btm
if so, these should probably be renamed.  

would you mind explaining the relationship between the number of cuts and the number of rounds in Papi ? is it direct (ex: cut_top=tournament.rounds) ?  I'm not sure how to reverse engineer Papi on this ^^'

Amaras: The defined modifiers for Buchholz are: Cut-n (n lowest contributions), and Med-n (n lowest and n highest contributions).

As for the number of cut rounds:
Screenshot_20250217-101836~2

src/data/player.py
```diff
@@ -475,37 +476,34 @@ def set_time_control(self, initial_time: int, increment: int, modified: bool):
...
+    @property
+    def board_number_sort_key(self) -> tuple[float, int, int, str, str]:
+        return -self.vpoints, -self.rating, -self.title, self.last_name, self.first_name
+
+    def rank_sort_key(
+            self, tournament: 'Tournament', max_round: int | None = None
+    ) -> tuple:
+        rank = (-self.points_after(max_round),)
+        for tie_break in tournament.tie_breaks:
+            rank += (-tie_break.compute_papi_player_value(self, tournament, max_round),)
```
Amaras: Be careful with that, because the tie-break functions compute tie-breaks *before* `max_round`

src/data/tournament.py
```diff
@@ -531,12 +550,20 @@ def to_trf(
             bb_fields=(self._trf_bb_fields() if trf_type == TrfType.PAIRING else {}),
         )
 
-    def _player_id_to_trf_id(self, player_id: int) -> int:
-        for trf_id, player in self.players_by_trf_id.items():
+    def _find_player_value_by_id(
+        self, player_id: int, players_by_value: dict[any, Player]
```
Amaras:
```suggestion
        self, player_id: int, players_by_value: dict[Any, Player]
```

Amaras (commented):
Looks good, just a few points to be cautious of

Amaras (approved):

PR #127, opened 2025-02-16
Moved BBP Pairings to /bin

pascalaubry:
Deleted /resources, the BBP pairings executable is now in /bin (but not on the repository, /utils/install_bbp_pairings.py can be used to install a new release).
Automatically install BBP Pairings for developers with PAPI_WEB_EXPERIMENTAL=1 when starting and always when exporting.


bin/bbpPairings/README.md
```diff
@@ -0,0 +1,3 @@
+This file is intended for developers to store different releases of BBP Pairings.
+
+Use utils/install_bbp_pairings to install a new release.
```
Amaras: Wrong path
```suggestion
Use utils/bbp_pairings/install_bbp_pairings.py to install a new release.
```

Amaras (approved):

PR #129, opened 2025-02-18
Improve player creation

timothyarmes:
Don't display validation errors before form submission
Don't preselect tournament if there are more than one.

### DIFF
--- a/src/web/templates/admin/players/ffe_search_results.html
+++ b/src/web/templates/admin/players/ffe_search_results.html
@@ -14,7 +14,11 @@
                     class="player-federation-flag pe-1"
                     src="{{ url_for('static', file_path='/images/federations/' ~ player.federation ~ '.svg') }}"
                 />
-                {{ player.ffe_licence_number }}
+                {% if player.club %}
+                    {{ player.ffe_licence_number }}
+                {% else %}
+                    -
+                {% endif %}
                 <span class="fw-bold">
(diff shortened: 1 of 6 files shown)


pascalaubry (dismissed):
Approved with my last commit that fixes error handling (all the rest is OK for me and works well).

pascalaubry:
@timothyarmes Please merge if you are OK with the last commit.

src/web/templates/admin/players/ffe_search_results.html
```diff
@@ -14,7 +14,11 @@
                     class="player-federation-flag pe-1"
                     src="{{ url_for('static', file_path='/images/federations/' ~ player.federation ~ '.svg') }}"
                 />
-                {{ player.ffe_licence_number }}
+                {% if player.club %}
```
pascalaubry: I dont understand why you check for club and not simply ffe_licence_number.

pascalaubry (dismissed):
Everything is clear for me except the comment on ffe_search_results.html (it may be the same to check club of ffe_licence_number, just clearer to me to check ffe_licence_number).

src/web/templates/admin/players/ffe_search_results.html
timothyarmes: Me neither, fixed.

pascalaubry (approved):

PR #130, opened 2025-02-19
Fix a crash in the case of empty screens

Amaras:


pascalaubry (approved):
Nice, thanks to KB for the bug report.

PR #131, opened 2025-02-19
Add tournament bye info

pascalaubry:
This PR is intended to prepare the players' history modal, the following fields have been added to Tournament:
- first_board_number
- paired_bye_points
- max_byes
- last_rounds_no_byes


src/common/papi_web_config.py
```diff
@@ -571,8 +572,20 @@ def local_url(self) -> str:
     """ The ChessEvent download URL. """
     chessevent_download_url: str = 'https://chessevent.echecs-bretagne.fr/download'
 
+    """ The default first board number for tournaments. """
+    default_first_board_number: int = 1
+
+    """ The points scored by players paired bye. """
+    default_paired_bye_points: Result = Result.FULL_POINT_BYE
```
Amaras:
```suggestion
    default_paired_bye_points: Result = Result.GAIN
```

src/data/util.py
```diff
@@ -183,6 +183,8 @@ def point_value(self) -> float:
                 | Result.FULL_POINT_BYE
             ):
                 return 1.0
+            case _:
+                raise ValueError(f'self={self}')
```
Amaras:
```suggestion
                raise ValueError(f'{self=}')
```

src/web/controllers/admin/index_admin_controller.py
```diff
@@ -98,6 +99,17 @@ def _get_record_illegal_moves_options(
         options[''] = _('By default - {option}').format(option=options[str(default)])
         return options
 
+    @staticmethod
+    def _get_paired_bye_points_options() -> dict[str, str]:
+        options: dict[str, str] = {
+            '': '',
+            str(Result.FULL_POINT_BYE.point_value): _('Full point bye'),
+            str(Result.HALF_POINT_BYE.point_value): _('Half point bye'),
```
Amaras:
```suggestion
            str(Result.GAIN.point_value): _('Full point bye'),
            str(Result.DRAW.point_value): _('Half point bye'),
```

src/web/controllers/index_controller.py
```diff
@@ -279,7 +279,7 @@ def value_to_form_data(value: str | int | float | bool | Path | None) -> str | N
         if isinstance(value, int):
             return str(value)
         if isinstance(value, float):
-            return f'{value:.2f}'
+            return str(value)
```
Amaras: Not convinced this will work as expected.
We should have some kind of rounding (uniformly applied)

Amaras (changes requested):
Did not run the code yet

src/web/controllers/index_controller.py
pascalaubry: Normalized all the float to string conversions.

Amaras (approved):

PR #133, opened 2025-02-20
Allow use of rotators without a screen ID

Amaras:
- **Fixed error in I18nTranslator.translate_message()**
- **Fixed misc type errors**
- **Fixed misc variable names warnings**
- **Fixed unused variables**
- **Added missing default case to util.py**
- **Version 2.4.22 Fixed variable names in tournament.py Added fields first_board_number, paired_bye_points, max_byes and last_rounds_no_byes to Tournament Remove table chessevent from create_event.sql**
- **Added bye information to tournaments to prepare the players' history modal**
- **Updated translations**
- **Normalized float to string conversion**
- **Updated translations**
- **Apply suggestions from code review**
- **Changed Result values**
- **Added URL for rotators without screen index**


src/web/controllers/user/screen_user_controller.py
```diff
@@ -408,3 +423,20 @@ async def htmx_user_rotator(
         if web_context.error:
             return web_context.error
         return self._user_screen_render(web_context)
+
+    @head(
+        path=[
```
pascalaubry, line 428: The interest of implementing the HEAD method is to suppress exceptions fro the log but it does not solve the initial problem: rotators and screens are not correctly displayed on JS-less clients.

pascalaubry (commented):
A minor improvment of the HEAD method, can be merged IMO.

Amaras:
Superseded by #137

PR #136, opened 2025-02-20
Rework events database migrations

Molrn:


src/database/sqlite/event_migration.py
```diff
@@ -0,0 +1,90 @@
...
+            logger.error(
+                'Database %s impossible to upgrade: version %s '
+                'is prior to the first upgradable version %s',
+                event_database.file.name,
+                event_database.version,
+                self.FIRST_UPGRADABLE_VERSION
+            )
+            return
+        if event_database.version > target_version:
+            raise NotImplementedError('Backwards compatibility not implemented')
```
Amaras:
```suggestion
            raise NotImplementedError('Forwards compatibility not implemented')
```
Backwards compatibility is the ability to use files from past versions.
Forward compatibility is the ability to use files from future versions.

Molrn: my bad, I was using 'backwards-compatibility' as 'The ability to rollback / to apply migration backwards'. Using 'forwards-compatibility' is though not the correct concept here, as it does not refer to 'using a newer file in an older application' but more 'rolling back a file to a previous version'

Amaras: I got the term and definition from here: https://en.wikipedia.org/wiki/Forward_compatibility
I think the term applies here (at least somewhat)

Molrn: I agree that automatic migrations as we have is considered 'backwards compatibility' (which I had been using wrong), but as I understand it forward compatibility is much more specific than this error. This error specifies that it is not possible to rollback a database to a previous version, not that the files from a newer version will remain compatible with a previous version of the app, which would mean being able to use files from newer versions without rolling them back.  

If that's ok with you, I changed the message to `Migrations rollback not implemented`

Amaras: > If that's ok with you, I changed the message to `Migrations rollback not implemented`

Perfect.

I (or @pascalaubry) will try it when we have some time to think about complex features

src/database/sqlite/event_migration.py
```diff
@@ -0,0 +1,90 @@
...
+class AbstractEventMigration(EventDatabase):
+    @abstractmethod
+    def forward(self):
+        pass
+
+
+class EventMigrationManager:
+    FIRST_UPGRADABLE_VERSION: Version = Version('2.4.0')
+    MIGRATION_BASE_MODULE_PATH: list[str] = ['database', 'migrations', 'events']
+    MIGRATION_MODULES: dict[Version, str] = {
```
pascalaubry: Versions should be hold by the migration modules and the mapping between the versions and the modules is not needed anymore.

pascalaubry (changes requested):
OK with the renumbering of the migration modules.

Molrn:
I kept migrations rollback out of this PR, because even though it is a few lines of code away from the changes of this migration, there still are a few ways to implement it, and I think @Amaras would like to have a say on it

pascalaubry (changes requested):
It is a minor thing but... Don't you think that having two digits for the minor and release numbers of the version in the module names (v2_04_08.py) would order them naturally?
Psychorigid you said?

Molrn:
ahah understandable ^^ done and rebased ;)

pascalaubry (dismissed):

pascalaubry (approved):

PR #137, opened 2025-02-20
Fix a bug, add basic responses to HEAD + others

Amaras:
- **Added URL for rotators without screen index**
- **Fix bug and add HEAD method**


pascalaubry (approved):
This PR does not solve the issues with JS-less clients and rotator/screen refresh but it removes the ugly warnings claiming that the HEAD method is not implemented.

PR #139, opened 2025-02-20
Fix missing estimation

Amaras:
When estimations are not set, TRF export does not work (`TypeError`).
This happens because players' estimation might be `None`, and papi estimation tries to do `400 - player.estimation`.

Two fixes for this:
  1. If `player.estimation` is not set, return 0 as a last resort;
  2. Compute estimations before exporting to TRF

- **Fix problem when estimation is not set**
- **Ensure all players are estimated before exporting**


pascalaubry (approved):
Tested on-the-fly tonight ;-)

PR #114, opened 2025-02-09
Adding tie-breaks computations

Amaras:
This ~(draft)~ PR aims to add tie-breaks to papi-web.

Right now, ~WIN, WON, BPG, BWG, PS (+CUT) and REP/GE~ all individual tie-breaks are implemented and tested.
The tests come from the tie-break exercises from FIDE TEC.
All but PTP and APPO have fully correct tie-breaks calculations.

I had to break off TournamentPlayer from Player for testing purposes, since there is too much information in this class to have a viable way to test the implementations.

PR #140, opened 2025-02-21
Tournament tie break management

Molrn:
An improvement of the tie-breaks such that it allows the handling of all tie-break functions
- tie breaks are stored in the tournament table as a json list, with their options
- on migration, the tie breaks from the papi database are added to the event database

it lacks a UI (not my strong point gotta admit ^^') and translations / tie-break types probably gotta be tweaked a bit, but I think this is a solid basis to base the tie-break on


src/data/tie_break.py
```diff
@@ -0,0 +1,476 @@
...
+            if option not in self.type.options:
+                raise ValueError(
+                    f'Option "{option.value}" is not allowed '
+                    f'for tie break "{self.type.value}"'
+                )
+            if not isinstance(value, option.type):
+                raise ValueError(
+                    f'Unexpected value for option "{option.value}": '
+                    f'"{value}" (expected_type: {option.type})'
+                )
```
Amaras, line 500: Some options are incompatible, I'll try to find the time to tell you which ones are.
In general, `papi_legacy` is incompatible with modifiers other than cuts

Amaras (commented):
Not bad from what I have quickly read, I'll review more in depth later

src/data/tie_break.py
Molrn: some incompatibilities could be removed by splitting the initial methods and having 2 separate ones. Example with sum of buchholz:
- from: sum_of_buchholz(fore_modifier, papi_legacy)
- to: sum_of_buchholz(papi_legacy), sum_of_fore_buchholz()

Amaras: I don't think I quite like splitting the functions.
It would involve quite a lot of code repetition, which I am not overly fond of.
I'm open to the idea, but not in favour.

Molrn: Could also be possible to have multiple TieBreakType entries for the same function:
- SUM_OF_BUCHHOLZ / SUM_OF_FORE_BUCHHOLZ  

To do so, we could have a `mandatory_options` a `forbidden_options` properties (SOB: `fore_modifier` forbidden, SOFB: `fore_modifier`=True mandatory / `papi_legacy` forbidden)
That could also be a way to split Buchholz into `Buchholz`, `Buchholz tronqué` and `Buchholz médian`

Molrn: or even simpler: a `overriden_options` where options can't be selected:
- SOB: fore_modifier=False
- SOFB: fore_modifier=True, papi_legacy=False

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -804,6 +804,134 @@ msgstr "échiquiers {board_numbers}"
...
+msgid "Tournament performance rating"
+msgstr "Performance ELO"
+
+#: src/data/tie_break.py
+msgid "Average performance rating of opponents"
+msgstr "Moyenne des performances ELO des adversaires"
+
+#: src/data/tie_break.py
+msgid "Perfect tournament performance"
+msgstr "Performance ELO parfaite"
```
pascalaubry, line 925: I don't see what "**Perfect** tournament performance" stands for.
Shouldn't it be "Perfect tournament performance rating" (cf "Tournament performance rating")?

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -4645,6 +4773,15 @@ msgstr "Dernières rondes sans bye :"
 msgid "The number of last rounds for which a player can get no byes (if not set, defaults to %(default)d)."
 msgstr "Le nombre de rondes finales pour lesquelles les joueur·euses ne peuvent prendre de bye (si non défini, %(default)d par défaut)."
 
+#: src/web/templates/admin/tournaments/tournament_modal.html
+msgid "Tie-Breaks"
+msgstr "Départages"
+
+#: src/web/templates/admin/tournaments/tournament_modal.html
+#, python-format
+msgid "Tie-Break %(index)d:"
```
pascalaubry: Enumerations are written like this on the other pages:
msgid "Tie-Break **#**%(index)d:"
msgstr "Départage **n°**%(index)d :" (space before ':' in French)

src/data/tie_break.py
```diff
@@ -0,0 +1,498 @@
+from collections.abc import Callable, Iterable
```
pascalaubry, line 1: @Molrn It would probably be useful that you check the way Tornelo implemented the tie-breaks; for instance cut_top and cut_bottom are options of a single Buchholz tie-break.
image
And maybe check we did not forget anything (but you probably checked the FIDE requirements):
image

locale/fr/LC_MESSAGES/messages.po
Molrn: that's the name of the tie-break in the [FIDE handbook](https://handbook.fide.com/chapter/TieBreakRegulations082024#:~:text=10.3%C2%A0%C2%A0%20Perfect%20Tournament%20Performance%20(PTP)). As for the translations, I'm cross-referencing them right now with the 'Livre de l'Arbitre', which does not include all of them

src/database/sql/create_event.sql
```diff
@@ -85,6 +85,7 @@ CREATE TABLE `tournament` (
     `last_ffe_upload` FLOAT NOT NULL DEFAULT 0.0,
     `last_ffe_rules_upload` FLOAT NOT NULL DEFAULT 0.0,
     `last_chessevent_download_md5` TEXT,
+    `tie_breaks` TEXT NOT NULL DEFAULT '[]',
```
pascalaubry: Optional fields are generally left NULL when empty, this makes possible the evolution of the storage format with no change to the database structure (see info.timer_colors|timer_delays and timer.colors|delays).

src/data/tie_break.py
Molrn: in  #114 @Amaras implemented all the individual tie-breaks described in the [FIDE Handbook section C07](https://handbook.fide.com/chapter/TieBreakRegulations082024), which are also the ones used in this TEC tie-break exercises documents.  

I also added Kashdan in this PR in order to have all the Tie-breaks defined in Papi. I think with FIDE tie-breaks + Papi, we are good. After that point, I suggest to only add tie-breaks if they are requested. If there are other tie-breaks that are relevant to you don't hesitate (for example Arranz in Tornelo that I did not know about) ;)  

Besides Arranz, the other tie-break Tornelo has that we haven't explicitely defined is Start number. It is used in Papi-web and [Tornelo](https://tornelo.com/knowledge-base/how-to-setup-tie-breaks/#:~:text=This%20tie%2Dbreak%20is%20always%20used%20as%20a%20last%20resort%20even%20if%20not%20chosen%20in%20the%20settings.) as default tie-break, but not explicitely in Papi-web. I we don't explicitely include it like Tornelo, I think we would need to at some point mention it in either documentation or directly in the UI

Molrn: Also about default tie-break: I still have no idea what Papi uses as default tie-break, and I ran out of ideas of what it could be. We probably will need to figure this out at some point  

it is not:
- the performance (even if it is implied in the 'livre de l'arbitre' section 4.2.8)
- any other tie-break method
- the starting rank
- the pairing number (tied to the starting rank)
- the id (because why not)
- Alphabetical (might be getting desperate here)

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -196,6 +198,25 @@ def _admin_validate_tournament_update_data(
...
                 raise ValueError(f'action=[{action}]')
+
+        if action == 'update':
+            tournament = web_context.admin_tournament
+            if (
+                tournament.stored_tournament.path == path
+                and tournament.stored_tournament.filename == filename
+                and tournament.file_exists
+            ):
+                tournament.update_tie_breaks(
```
pascalaubry: tournament.update_tie_breaks() should be called in _admin_tournament_update(), not here (in this method we only validate the data without changing it).

src/web/templates/admin/tournaments/tournament_modal.html
```diff
@@ -285,6 +285,40 @@ <h4>
                             ) }}
                         </div>
                     </div>
+                    {% if edit_tie_breaks %}
+                        <h4>
+                            {{ _('Tie-Breaks') }}
```
pascalaubry: Tie-breaks (no upper-case B)

pascalaubry (changes requested):
Everything clear to me except the place of the call to tournament.update_tie_breaks().
Needs another review on the tie-break code itself anyway.

src/data/tie_break.py
pascalaubry: +1: at this time only the tie-breaks requested by the FIDE should be implemented.

src/data/tie_break.py
pascalaubry: @Molrn I let you mark the incompatible options.
Note: the options are OK for the Papi tie-breaks (tests OK).

pascalaubry:
Please fill 85-database.md for the new fields.

src/data/tie_break.py
Molrn: Incompatible options:
- Buchholz: played_modifier=True / papi_legacy=True
- Sum of Buchholz: fore_modifier=True / papi_legacy=True  
- in general: papi_legacy=True / cut options (need to be 1, 2 or 3 depending on the number of rounds)  

All the other options appear to be compatible.  

Considering all the interaction issues appear to be related to papi_legacy, I suggest to treat papi_legacy differently than the other options. The purpose of this option seems to be to generate exactly the same tie-breaks as Papi, so generating tie-breaks with similar mistakes but more options than papi seems not relevant. Therefore, having in the form of each tie-break a checkbox 'papi_legacy' is in my opinion not a good idea.  

Instead, I suggest the following appraoch: having a global tie-break option 'papi_legacy'. If this option is selected, limit the amount of tie breaks to 3, and select the tie-breaks amongst the values of the PapiTieBreak Enum (similar to the current interface).  
As the options incompatibilites are not breaking anything or even generating false tie-breaks, I think it is safe not to implement this incompatibility in the TieBreak class, and to just rely on the interface to match these incompatibilities. It also saves the trouble of explaining them to the users.

docs/85-database.md
```diff
@@ -83,7 +77,9 @@
...
-| `chessevent_id`                       | `INTEGER` | REFERENCES `chessevent`(`id`)              | L'identifiant de la connexion à ChessEvent.                                                                                                                                             |
+| `chessevent_user_id`                  | `TEXT`    |                                            | L'identifiant utilisé pour le connexion à la plateforme ChessEvent.                                                                                                                     |
+| `chessevent_password`                 | `TEXT`    |                                            | Le mot de passe utilisé pour le connexion à la plateforme ChessEvent.                                                                                                                   |
+| `chessevent_event_id`                 | `TEXT`    |                                            | L'identifiant de l'évènement ChessEvent sur la plateforme.                                                                                                                              |
```
pascalaubry, line 82: These changes had been forgotten, thank you.

docs/85-database.md
```diff
@@ -95,6 +91,10 @@
...
 | `last_chessevent_download_md5`        | `TEXT`    |                                            | L'empreinte du dernier téléchargement du tournoi depuis la plateforme ChessEvent.                                                                                                       |
+| `first_board_number`                  | `INTEGER` |                                            | Le numéro du premier échiquier.                                                                                                                                                         |
+| `paired_bye_points`                   | `FLOAT`   |                                            | Points attribués aux exempts.                                                                                                                                                           |
+| `max_byes`                            | `INTEGER` |                                            | Le nombre maximum de byes auquel un·e joueur·euse peut prétendre                                                                                                                        |
```
pascalaubry, line 96: These too, btw can you add last_rounds_no_byes?

pascalaubry (approved):

PR #141, opened 2025-02-23
feat: Event Databases back-ups

Molrn:


src/common/papi_web_config.py
```diff
@@ -396,6 +396,12 @@ def project(self) -> str:
     """ The extension of archives event databases. """
     event_archive_ext: str = 'arch'
 
+    """ The base path where event database back-ups are stored. """
```
pascalaubry: For me 'backup' stands for « sauvegarde », so the folder name should simply be 'backup'.

src/data/loader.py
```diff
@@ -262,3 +264,80 @@ def archives_sorted_by_date(self) -> list[Archive]:
             ],
             key=lambda archive: archive.date,
         )
+
+
+@dataclass
+class BackUp:
```
pascalaubry: Same as previous comment: this class should be renamed to Backup, or maybe EventBackup.

src/database/sqlite/event_database.py
```diff
@@ -1288,6 +1301,7 @@ def upgrade(self):
                 f'Your Papi-web version ({papi_web_version}) can not open database {self.file.name} (version '
                 f'{self.version}), please upgrade.'
             )
+        self.create_backup()
```
pascalaubry: I has understood that the backup feature was reserved to developers (since upgrading **must** work for any production database), so the backup creation should only done if DEVEL_ENV.

pascalaubry: And BTW the utilities of /utils are not available to non-developers.

pascalaubry (changes requested):
OK for me if backups are reserved to developers.

pascalaubry (approved):
Thanks for the changes.

PR #142, opened 2025-02-23
Improve database import experience

timothyarmes:
This improves the FFE/FIDE database import experience.

In both cases a progress indicator is now shown to the user.  I the case of the FFE database the import time is identical.

For the FIDE import, it is slightly slower, (58 -> 78s) but I think this is still superior to showing no feedback at all for an entire minute.

The PR also fixes a UI issue with the player creation modal


pascalaubry (commented):
The display of the number of players written only works in debug mode for me.
image
image
Any idea?

pascalaubry (approved):

PR #146, opened 2025-02-26
Print participants

timothyarmes:


src/data/util.py
```diff
@@ -1229,4 +1229,57 @@ def performance_bonus(
         bonus = performance_table[index]
     if fractional_score < 0.5:
         bonus *= -1
-    return bonus
\ No newline at end of file
+    return bonus
+
+class PrintSplit(StrEnum):
+    NoSplit = auto()
```
pascalaubry, line 1235: Shouldn't we uppercase like in the other Enum classes?

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -596,6 +604,27 @@ def _admin_event_tab_render(
...
+                    data = (
+                        {
+                            'tournament_id': WebContext.value_to_form_data(tournament_id),
+                            'split': WebContext.value_to_form_data(PrintSplit.NoSplit.to_str()),
+                        }
+                    )
+
+                template_context |= {
+                    'modal': 'print',
+                    'tournament_options': web_context.get_tournament_options(),
```
pascalaubry, line 622: Shouldn't we use here a new method get_tournament_with_file_options() based on Event.tournaments_with_file_sorted_by_uniq_id()?

src/web/templates/admin/event.html
```diff
@@ -48,6 +48,20 @@
     {% endfor %}
 {% endblock admin_header_nav %}
 
+{% block admin_header_action_buttons %}
```
pascalaubry, line 51: image
Well it took me some time to find the print button :-)
Why didn't you create another Print button like Edit and Clone?

src/web/templates/admin/players/print_view.html
```diff
@@ -0,0 +1,81 @@
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html;" charset="iso-8859-1">
+        <title>{{ tournament.name }}</title>
+        <style type="text/css">
+            .papi_liste_c {font-family: Arial, Helvetica, sans-serif; font-size: 12px; font-weight: normal; color: #000000; background-color: #FFFFFF;}
```
pascalaubry: Am I the only one not to be nostalgic of Papi CSS classes?
(could we use classes like player-odd|even instead?)
Note: I really like using Jinja for printable outputs.

src/web/templates/admin/players/print_view.html
```diff
@@ -0,0 +1,81 @@
+<html>
```
pascalaubry: I think we can anticipate right now that some more print views will come and rename this file src/web/templates/admin/print/players.html.

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1068,3 +1072,75 @@ async def htmx_admin_player_check_out(
             player_id=player_id,
             check_in=False,
         )
+
```
pascalaubry: Shouldn't this method be in tournament_admin_controller.py using TournamentAdminWebContext?

pascalaubry (changes requested):
Not sure of all my comments, probably easier to talk about them.

src/data/util.py
timothyarmes: I based it on :

```
class ScreenType(StrEnum):
    Boards = auto()
    Input = auto()
    Players = auto()
    Results = auto()
    Image = auto()
```

src/web/templates/admin/event.html
timothyarmes: I wanted it availble everywhere, not just the config tab, if that's what you mean?

src/web/templates/admin/players/print_view.html
timothyarmes: So I thought it best that I look exactly like the print outs that we're all use to.  I can rename the classes though :)

src/web/controllers/admin/player_admin_controller.py
timothyarmes: It's a list of players, so it seems best in the players controller, no?

src/web/templates/admin/players/print_view.html
timothyarmes: Good idea.

pascalaubry (approved):

PR #151, opened 2025-02-26
Fixed select_input() when an empty option is already provided.

pascalaubry:
This PR was created because of this:
image
I wonder if macros.j2:232-236 should not simply be removed.


timothyarmes:
That code is useful.

For example in the create player dialog we start  with the empty option selected by default for the tournament options to force the arbiter to select the tournament.  However, we editing the player we no longer offer the empty option.  This is the best UX, and it's handled automatically by the select without needing to hand code an empty option.

However, I think your drop-down itself is wrong.  By default we get the value from the event - that's what should be selected if the user hasn't overridden it - it's the default option !  The empty option shouldn't be there at all.

There are other examples of this UI anomaly too:

image

If there's a default fallback then the empty option should never be in the list.

pascalaubry:
I forgot the need of forcing the choice of the tournament on player creation, but it added a useless empty option to other select inputs.
The list of the options shown to the user should be initialized by the controller, I did it on my last commit.
_If there's a default fallback then the empty option should never be in the list_: yes, it was the goal of this PR.

timothyarmes:
OK I understand.

timothyarmes (approved):

PR #155, opened 2025-02-27
Add rankings and 'Grille américaine' as printable documents

Molrn:


Molrn:
@timothyarmes I hide the round input for players list, but did not put a select for the round. Indeed, as it is an event modal with a switch on the tournament, printable rounds would need to switch depending on the selected tournament, adding in my opinion an unnecessary level of complexity. If it really is important in your opinion to have a select, I can modify it

timothyarmes:
Understood, not that big a deal.

Amaras (approved):
Code looks pretty good, works as expected

PR #156, opened 2025-02-28
Fixing estimations computations

Amaras:
This PR aims to fix estimations to get closer to papi values.

I found this thanks to the work done in #155


timothyarmes (dismissed):

timothyarmes (dismissed):
The only thing I'd say is that this is a case where some comments are justified.

It's not clear that this is a Papi compatibility function or what problem it solves, and it's not clear what exactly it's doing and why it's working it's way back and forwards though the list.

timothyarmes (approved):

PR #157, opened 2025-02-28
Fix: Take missing opponent ids into account in tie-breaks

Molrn:


pascalaubry:
Well, not confident enough on this part to review. @Amaras ?

src/tie_breaks/individual.py
```diff
@@ -639,7 +648,7 @@ def average_rating_opponents(
     ]
     ratings = []
     for pairing in pairings:
-        if pairing.unplayed:
+        if pairing.opponent_id is None or pairing.unplayed:
```
Amaras: `pairing.opponent_id is None` implies `pairing.unplayed`, unless there is a bug

src/tie_breaks/individual.py
```diff
@@ -389,7 +397,8 @@ def sum_of_buchholz(
     opponents: list[Player] = [
         tournament.players_by_id.get(pairing.opponent_id)
         for round_index, pairing in player.pairings.items()
-        if round_index < max_round
+        if round_index <= max_round
+        and pairing.opponent_id is not None
```
Amaras: The filtering is done line 412 already, why add it here?
I specifically used `get` as well (maybe the typing should reflect that)

src/tie_breaks/individual.py
```diff
@@ -279,6 +283,8 @@ def buchholz(
             else:
                 scores.append(dummy_points)
             continue
+        if pairing.opponent_id is None:
+            continue
```
Amaras: If we get there, we should have an `opponent_id`, that's why I didn't check.
When does this break?

src/tie_breaks/individual.py
```diff
@@ -355,6 +361,8 @@ def fore_buchholz(
...
```
Amaras: All the cases where there is no `opponent_id` should be covered already

src/tie_breaks/individual.py
```diff
@@ -678,6 +687,8 @@ def tournament_performance_rating(
     ratings = []
     score = 0
     for pairing in pairings:
+        if pairing.opponent_id is None:
+            continue
```
Amaras: `pairing.played` implies `pairing.opponent_id is not None`

src/tie_breaks/individual.py
```diff
@@ -724,6 +735,8 @@ def average_performance_rating_opponents(
     ]
     performance_ratings = []
     for pairing in played_games:
+        if pairing.opponent_id is None:
+            continue
```
Amaras: `pairing.played` implies `pairing.opponent_id is not None`

src/tie_breaks/individual.py
```diff
@@ -791,17 +804,28 @@ def perfect_tournament_performance(
         pairing
         for round_index, pairing in player.pairings.items()
         if round_index <= max_round and pairing.played
+        and pairing.opponent_id is not None
```
Amaras: `pairing.played` implies `pairing.opponent_id is not None`

src/tie_breaks/individual.py
```diff
@@ -847,6 +871,7 @@ def average_perfect_performance(
...
```
Amaras: `pairing.played` implies `pairing.opponent_id is not None`

Amaras (commented):
Okay, I don't actually understand why most of this PR is needed.
Some additions are fair, but I don't understand why the others are needed (those with comments).
Please give me an example where computations go wrong without the commented additions, and I'll test it.

src/tie_breaks/individual.py
Molrn: I implemented these checks not to take into account the implied value of opponent_id. I get that it might not be that necessary and I don't have a strong opinion on this, so I don't mind removing those.

Amaras (approved):

PR #158, opened 2025-02-28
Uniformize StrEnum usage

Molrn:
use the following pattern for all StrEnum classes:
```python
class PrintSplit(StrEnum):
    NO_SPLIT = 'no-split'

    @property
    def name(self) -> str:
    # Translated name
        match self:
            case PrintSplit.NO_SPLIT:
                return _('No split')
    
        def __str__(self) -> str:
            # If strong usage of this method, keep it as is to prevent generating issues 
            # --> usage hard to track (BoardColor / ScreenType)
            return self.name
```


pascalaubry (approved):
Nice and works fine.

PR #162, opened 2025-02-28
Add a unicity check on tie-breaks

Molrn:
image  
Prevent from using twice the same tie-break


locale/en/LC_MESSAGES/messages.po
```diff
@@ -2162,6 +2162,10 @@ msgstr ""
 msgid "The password of the tournament on the FFE website is made of 10 uppercase letters."
 msgstr ""
 
+#: src/web/controllers/admin/tournament_admin_controller.py
+msgid "Tie-break already at use."
```
Amaras: I think it should be "in use", rather than "at use"

Amaras (approved):

PR #165, opened 2025-03-01
Migration rollbacks for event databases

Molrn:
add the command `python utils/event_databases/migrate.py`:
````sh
Command migrating one or more event databases to a specific version.

options:
  -h, --help            show this help message and exit
  -v VERSION, --version VERSION
                        Version of the database to migrate to. Defaults to the last version.
  -e EVENT, --event EVENT
                        ID of the event to migrate. Defaults to all the events.
````  

Migrating to a previous version uses the backward methods defined in the EventMigration classes. By default that method raises a `NotImplementedError`. This disables the rollback feature for migrations we don't want / can't implement the backward method of (ex: 2.4.21)

### DIFF
--- a/src/database/sqlite/event_migration.py
+++ b/src/database/sqlite/event_migration.py
@@ -4,6 +4,7 @@ from pkgutil import iter_modules
 import re
 from abc import abstractmethod
 from logging import Logger
+from sqlite3 import OperationalError
 
 from packaging.version import Version
 
@@ -20,6 +21,11 @@ class AbstractEventMigration(EventDatabase):
     def forward(self):
         pass
 
+    def backward(self):
+        raise NotImplementedError(
+            "Rollback not implemented for this migration"
+        )
+
 
 class EventMigrationManager:
     EMPTY_DATABASE_VERSION: Version = Version('0.0.0')
@@ -29,6 +35,13 @@ class EventMigrationManager:
     def migration_modules(self) -> list[str]:
         return [
             f'{events.__name__}.{module}'
+            for module in self._migration_module_names
+        ]
+
+    @cached_property
+    def migration_versions(self) -> list[Version]:
+        return [
+            self._module_name_to_version(module)
             for module in self._migration_module_names
         ]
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: utils/event_databases/restore.py [hunk 1]
```diff
@@ -42,38 +40,38 @@
     if args.version:
         version = Version(args.version)
         if version > PapiWebConfig.version:
-            logger.error(
+            sys.stderr.write(
```
pascalaubry: Can you use print_interactive_error() and print_interactive_info() instead?

utils/event_databases/migrate.py
```diff
@@ -0,0 +1,51 @@
...
+    migration_manager = EventMigrationManager()
+    version = (
+        Version(args.version) if args.version else
+        migration_manager.last_migration_version
+    )
+    event_ids = [args.event] if args.event else EventLoader().event_uniq_ids
+    for event_id in event_ids:
+        with EventDatabase(event_id, True, False) as event_database:
+            if event_database.version == version:
+                sys.stdout.write(
```
pascalaubry: Can you use print_interactive_info() instead?

utils/event_databases/migrate.py
```diff
@@ -0,0 +1,51 @@
...
+    parser.add_argument(
+        '-v',
+        '--version',
+        type=str,
+        help=(
+            'Version of the database to migrate to. '
+            'Defaults to the last version.'
+        ),
+    )
+    parser.add_argument(
```
pascalaubry, line 28: Can you add a --all-events options to apply to all of the databases and make one of --all-events or --event mandatory?
(I was really surprised to see the script applying to all my databases by default...)

src/database/sqlite/event_migration.py
```diff
@@ -46,49 +59,98 @@ def last_migration_version(self) -> Version:
...
                 )
-            else:
+            except OperationalError as e:
+                raise PapiWebException(
+                    f'Database {database.file.name} '
+                    f'({database.version}) could not be upgraded '
+                    f'to version {migration_version}: "{e}"'
+                )
+
+    def _rollback(
```
pascalaubry, line 124: image
Instead of throwing exceptions in this method, could you print a simple error message and return a boolean?

pascalaubry (changes requested):
Apart from a few comments, works.
Still thinking that it is only comfort for developers, but OK let's cherish ourselves :-)

Molrn:
for now it is comfort for devs, but it's the kind of stuff that might come in handy if one day the app is deployed and something needs to be fixed in production ;) and comfort for devs is rather important for OS projects :)

pascalaubry (approved):
Thanks.

PR #166, opened 2025-03-02
Fix hidden round input in print modal

Molrn:
Once the modal was opened and closed, the input for rounds stayed hidden


Amaras (approved):
Works on my machine

PR #168, opened 2025-03-02
Fix the submit button on the tournament modal

timothyarmes:


Amaras:
This doesn't fix my use case

pascalaubry (approved):

Issue #164, opened 2025-03-01
Missing FR translations


pascalaubry:
## Bug description
The following translations a re missing:
msgid "Played modifier"
msgid "Papi legacy"
msgid "Fore modifier"
msgid "Exclude players"

Molrn:
I suggest removing for now these translations, and the other ones for tie-break options. Indeed, these translations are not currently used, and will only be used when implementing a detailed UI for tie-breaks. at this point, translations are gonna be deeply related to their usage in the UI. Removed in #170 (don't hesitate to close it if disagree it should be removed)

Molrn:
removed in #285, will be re-added when the UI for tie-break options will be implemented

PR #170, opened 2025-03-02
Remove Tie-break options translations

Molrn:
related to #164.


pascalaubry:
Well I feel like we will need these translations one day, so the goal of the issue was not to remove the strings but to find FR translations, even approximative (just don't have time now to search for acceptable proposals now).

Molrn:
PR explained in the related issue https://github.com/papi-web-org/papi-web/issues/164#issuecomment-2692917631, but if you'd rather keep them no problem, we can close then

PR #172, opened 2025-03-04
Goto a sensible tabs when entering an event

timothyarmes:


src/web/controllers/admin/index_admin_controller.py
```diff
@@ -688,6 +688,17 @@ def _prepare_event_modal_data(
                 for i in range(1, 4)
             }
         )
+        
+    @staticmethod
```
pascalaubry: Why didn't you add property Event.default_nav_id?

timothyarmes: I'm not sure that would be appropriate either, why should the Event class know anything about the tab ids defined in the event_admin_controller?  It's a rendering specific issue.

The controllers seem to be the appropriate place to define controller specific logic, no?  However on the same basis, I'd agree that the index_controller shouldn't know anything about the tabs defined by the event_controller, so I've moved that function into there.

pascalaubry: I do agree.

pascalaubry (changes requested):
Works fine for me, please juste rename EventAdminController._get_default_nav_id() to EventAdminController.get_default_nav_id() as it is used as a public method.

timothyarmes:
Good point.

pascalaubry (approved):

PR #173, opened 2025-03-04
Refactor controller classes for clarity

timothyarmes:


pascalaubry:
Why didn't you rename AbstractController (BaseController) and AbstractInputUserController (BaseInputUserController)?

pascalaubry:
For my understanding: what is the difference between
**from web.controllers.admin.base_event_admin_controller import BaseEventAdminController
[...]
'get_default_nav_id': BEAC.BaseEventAdminController.get_default_nav_id,****
and
**import web.controllers.admin.base_event_admin_controller as BEAC
[...]
'get_default_nav_id': BEAC.BaseEventAdminController.get_default_nav_id,**
?
(no need for me to adopt this syntax here)

timothyarmes:
> Why didn't you rename AbstractController (BaseController) and AbstractInputUserController (BaseInputUserController)?

Missed those two, thanks.

timothyarmes:
> For my understanding: what is the difference between **from web.controllers.admin.base_event_admin_controller import BaseEventAdminController [...] 'get_default_nav_id': BEAC.BaseEventAdminController.get_default_nav_id,**** and **import web.controllers.admin.base_event_admin_controller as BEAC [...] 'get_default_nav_id': BEAC.BaseEventAdminController.get_default_nav_id,** ? (no need for me to adopt this syntax here)

I can't use the `from xxx import yyy` syntax here due to a circular reference.  But I can import the entire module using the `import xxx as yyy` since I'm not referring to a specific export at import time

pascalaubry (commented):
My feeling is that get_default_nav_id() should be moved to BaseAdminController next to _admin_render() (and maybe renamed back to _get_default_nav_id...), this will remove the circular import.

timothyarmes:
> My feeling is that get_default_nav_id() should be moved to BaseAdminController next to _admin_render() (and maybe renamed back to _get_default_nav_id...), this will remove the circular import.

BaseAdminController shouldn't know or assume anything about the nav ids that are defined in the EventAdminController.  Data encapsulation.

pascalaubry:
OK, back home to look at it with no meeting at the same time ;-)
You are completely right, moving get_default_nav_id() was a bad idea.
I think we can do something very simple with no circular import and no additional method passed to the template context, by adding these lines at base_event_admin_conrtoller.py:160:
        if not web_context.admin_event_tab:
            # web_context.admin_event_tab = list(nav_tabs.keys())[0]
            if web_context.admin_event.player_count:
                web_context.admin_event_tab = 'players'
            elif web_context.admin_event.tournaments_by_uniq_id:
                web_context.admin_event_tab = 'tournaments'
            else:
                web_context.admin_event_tab = 'config'
image

timothyarmes:
Agreed, that's nicer

pascalaubry (approved):
Thanks and sorry not to have seen this today, too busy...

PR #175, opened 2025-03-04
Added missing information on tournament cards

pascalaubry:
Fixed missing async on htmx_admin_default_timers_update()
Added missing information on tournament cards
Converted StoredTournament.paired_bye_points from float to int
**Fixed database migration**

~~for developers, do the following after updating the tournament-card branch:~~
~~1. downgrade to 2.4.21: migrate.py --all-events --version=2.4.21~~
~~2. upgrade again: migrate.py --all-events~~

PR #176, opened 2025-03-04
Add point_values to the database

Amaras:
- Added point_values as JSON in the tournament table
- Added load and dump (with default values)
- Pushed version to 2.4.24 to avoid collisions with #175
- Added Youri as author


src/data/tournament.py
```diff
@@ -111,6 +112,7 @@ def __init__(
         self._papi_tie_breaks: tuple[
             PapiTieBreak, PapiTieBreak, PapiTieBreak
         ] | None = None
+        self._point_values: dict[Result, float] = PointValueType.STANDARD.point_values
```
pascalaubry: I thought that you would have put a PointValueType field here, it shall be easier to move from Papi to SQLite (just storing an int).

src/data/util.py
```diff
@@ -1357,3 +1357,26 @@ def name(self) -> str:
...
+
+
+class PointValueType(Enum):
+    STANDARD = 1
+    PAPI_3_POINTS = 2
+
+    @property
+    def point_values(self) -> dict[Result, float]:
+        match self:
+            case PointValueType.STANDARD:
```
pascalaubry, line 1294: I thought that you would have return a dict with all the possible results as keys.

src/database/access/papi/papi_database.py
```diff
@@ -79,6 +81,10 @@ def read_info(self) -> TournamentInfo:
             PapiTieBreak.from_papi_value(self._read_var('Dep2')),
             PapiTieBreak.from_papi_value(self._read_var('Dep3')),
         )
```
pascalaubry, line 82: Same remark as before : I thought you would have returned here STANDARD or PAPI_3_POINTS.

src/database/migrations/events/v2_04_25.py
```diff
@@ -0,0 +1,9 @@
+from database.sqlite.event_migration import AbstractEventMigration
```
pascalaubry: No database change, so this file is useless.

src/database/store.py
```diff
@@ -2,6 +2,7 @@
 
 from common.papi_web_config import PapiWebConfig
 from data.tie_break import TieBreak
```
pascalaubry: Import is not used.

pascalaubry (changes requested):
Not sure of having dict[Result, float] or PointValueType.
Maybe rename PointValueType to ResultValues?

src/data/util.py
Amaras: Why?
The fallback is enough

src/data/tournament.py
```diff
@@ -1247,6 +1247,7 @@ def update_papi_database_from_stored_tournament(self):
         with (PapiDatabase(self.file, write=True) as papi_database):
             if tie_breaks := self._update_papi_tie_breaks():
                 papi_database.update_tie_breaks(tie_breaks)
```
pascalaubry, line 1246: papi_database.update_point_values(self._point_value_type)?

src/data/util.py
pascalaubry: Now that I read Result.points() I understand.
Note: I have some warnings in this method:
image
I have to do this to get rid of the warnings:
image

pascalaubry (changes requested):

src/data/tournament.py
pascalaubry: Committed the missing underscore.

pascalaubry (commented):
Can not really test before merging dev.
Tell me if you want me to do it.

pascalaubry (approved):

PR #177, opened 2025-03-05
Create empty Papi file if none selected

Molrn:


src/web/templates/admin/tournaments/create_papi_modal.html
```diff
@@ -0,0 +1,50 @@
+<div class="modal-dialog modal-dialog-centered admin-create-papi-modal d-block">
```
pascalaubry, line 1: I really don't feel confortable with this modal closing when clicking out of the dialog box (just wanted to give the focus to my browser, it closed the modal and I had to re-edit the tournament to see it again).

src/web/templates/admin/tournaments/create_papi_modal.html
```diff
@@ -0,0 +1,50 @@
...
+    >
+        <div class="modal-content">
+            <div class="modal-header">
+                <h2 class="modal-title">
+                    {{ _('Missing Papi file') }}
+                </h2>
+            </div>
+            <div class="modal-body">
+                <h4 class="text-danger">
+                    {{ _('Warning: Papi file [%(path)s] not found. Many features are disabled.', path=admin_tournament.file) }}
```
pascalaubry: Suggestion: found. Many > found, many

src/web/templates/admin/tournaments/create_papi_modal.html
```diff
@@ -0,0 +1,50 @@
...
+                </h4>
```
pascalaubry: Suggestion: 
add here a natural language sentence like: Do you want Papi-web to create the file?
and change the button texts:
- Yes, create the file
- No I will do it later

src/web/templates/admin/tournaments/create_papi_modal.html
```diff
@@ -0,0 +1,50 @@
...
+                {{ admin_macros.select_input(
+                    name='papi_version',
+                    label=_('Papi version'),
```
pascalaubry: Papi version > Papi version:

pascalaubry (changes requested):
Works well, the most important comment is the one concerning modal closing.

timothyarmes (changes requested):
If I create a new event, then a new tournament, all with default values, then I get an error:

 ```
{% include 'tournaments/create_papi_modal.html' %}
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "\\Mac\Home\Development\OpenSource\papi-web\src\web\templates\admin\tournaments\create_papi_modal.html", line 3, in top-level template code
    hx-post="{{ url_for('admin-tournament-papi-create', event_uniq_id=admin_event.uniq_id, tournament_id=admin_tournament.id) }}"
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\timothyarmes\AppData\Local\Programs\Python\Python311\Lib\site-packages\litestar\template\base.py", line 53, in url_for
    return _get_request_from_context(context).app.route_reverse(route_name, **path_parameters)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\timothyarmes\AppData\Local\Programs\Python\Python311\Lib\site-packages\litestar\app.py", line 798, in route_reverse
    raise NoRouteMatchFoundException(
litestar.exceptions.http_exceptions.NoRouteMatchFoundException: 500: Received type for path parameter tournament_id doesn't match declared type <class 'int'>
```

src/web/templates/admin/tournaments/create_papi_modal.html
timothyarmes: It's global behaviour, I don't thing we can turn that off for one modal

src/web/templates/admin/tournaments/create_papi_modal.html
Molrn: with such a modal with a natural language question, having options on how to create the file feels wrong. I suggest switching to 2 modals: 
- "Missing Papi file" with the warning, the question and YES / NO buttons
- "Create Papi file" with options and CREATE / CANCEL buttons

"Missing Papi file".YES opens second modal

src/web/templates/admin/tournaments/create_papi_modal.html
Molrn: Despite this modal, there might be others that we'd like to keep the focus on, I think it could be worth having a second modal template for those, that isn't closed when the focus is lost. But it's something I'm not really comfortable with, so leaving it to @timothyarmes 's judgement

src/web/templates/admin/tournaments/create_papi_modal.html
timothyarmes: Two modals sounds awful.

Why are we even asking to create old formats?

Molrn: Besides papi version, I also had in mind to add datasources here in a second time, so a place for options is not just for papi-version  

> Why are we even asking to create old formats?  

That's what is done for the creation of Papi files from chessevent Tournaments. I did not question it, I don't know exactly the differences between these versions and how much are '3.3.6' and '3.3.7' still used today, so I think it's not my place to say

Molrn: image
Fine like this then ?

pascalaubry: Well I don't see any other option to add, and I agree that asking the Papi version can be deleted (it should also be done in the ChessEvent script).
Let's simplify!
image

Molrn:
> If I create a new event, then a new tournament, all with default values, then I get an error:

fixed :)

timothyarmes (changes requested):
If I say no thank you so as not to create the Papi file, then the newly created tournament doesn't appear on the page

timothyarmes:
Also, if the directory that I chose doesn't exist then it isn't created, and the creation fails without an error.
I think it should try to create the directory.  I have a folder for each event , and I don't want to have to create the folder in advance :)

pascalaubry:
> If I say no thank you so as not to create the Papi file, then the newly created tournament doesn't appear on the page

@Molrn Clicking on No thank you should not only dismiss the modal but also refresh the body.

src/web/templates/admin/tournaments/create_papi_modal.html
pascalaubry: Fixed by your last commit, thanks.

Molrn:
> If I say no thank you so as not to create the Papi file, then the newly created tournament doesn't appear on the page  
> Also, if the directory that I chose doesn't exist then it isn't created, and the creation fails without an error.  

all fixed

src/web/templates/admin/tournaments/create_papi_modal.html
pascalaubry: OK for me if I am the only one not confortable with losing the focus of the modal too easily.

pascalaubry (approved):
@timothyarmes I let you merge if OK for you.

pascalaubry (approved):

timothyarmes:
It's now working, but should any information in the modal to be added to the Papi, such as the cadence and numéro d'homologation?

timothyarmes (approved):

pascalaubry:
Indeed, you are right... 
@Molrn Is it possible to use the information given on the modal to fill the Papi file?
Or we consider from now on that there is no link between the two databases and we forget everything of the Papi file except the players' information and the pairings?

pascalaubry:
My opinion: we don't care of the Papi file anymore since we have now correct rankings :-)

Molrn:
I think this is overachieving for papi, like the shared informations between DB and Papi are gonna fluctuate until everything shifts to the DB. I'd be more into leaving the papi-database empty at creation for now, and to fill it once  every value is in the SQLite DB and the feature shifts to a full on papi export

PR #178, opened 2025-03-05
Add rank to player class

Molrn:


pascalaubry:
Added Player.rank to make Player know their rank (this will be used by sets of ranking screens, work in progress on branch ranking-screen).
Simplified the print view
Fixed translations, unused imports...

src/data/tournament.py
```diff
@@ -440,14 +441,8 @@ def players_by_name_without_unpaired(self) -> list[Player]:
...
-        ranked_players = sorted(
-            self.players_by_id.values(),
-            key=lambda player: player.rank_sort_key,
-        )
-        return {
-            rank: player for rank, player in
-            enumerate(ranked_players, start=1)
-        }
+        assert self._players_by_rank is not None, 'Tournament._players_sorted_by_rank is not set, call Tournament.compute_player_ranks() before.'
+        return self._players_by_rank
```
Amaras, line 445: Maybe compute the ranks, instead of throwing an `AssertionError`?
If you're not computing anything, this method can be a property.

Amaras (changes requested):
Overall looks good.
I don't understand the point of `Tourament.players_by_rank` if you're not computing the ranks inside.
Also, the assertion message is wrong.

pascalaubry:
Branch player-class-rank merged with ranking-screen, new PR #184

PR #188, opened 2025-03-07
Initial plugin work

timothyarmes:


Amaras (commented):
Without trying to run the code, it looks good.
I'll run it when I have some time

src/plugins/chessevent/chessevent.py
```diff
@@ -0,0 +1,17 @@
+
```
pascalaubry, line 1: ~~Is there a place where all the methods of the pugins are describes?~~
~~Can Pluggy do that automagically?~~
Found.

src/plugins/ffe/ffe.py
```diff
@@ -0,0 +1,114 @@
...
+                player.ffe_licence_number = ffe_player.ffe_licence_number
+                player.club = ffe_player.club
+                player.league = ffe_player.league
+                player.comment = ffe_player.comment
+
+@hookimpl
+def get_tournament_card_block_template():
+    return "/ffe_tournament_card_block.html"
+
+def split_players_by(split_by: str, players: list[Player]):
```
pascalaubry, line 71: No @hookimpl here?

src/plugins/ffe/ffe.py
```diff
@@ -0,0 +1,114 @@
...
+        ),
+    ]
+
+@hookimpl
+def get_extra_print_view_columns(document: PrintDocument):
+    match document:
+        case PrintDocument.PLAYER_LIST | PrintDocument.RANKING | PrintDocument.CROSSTABLE:
+            return [
+                ExtraColumn(
+                    insertion_index=6,
```
pascalaubry, line 106: Probably impossible to have several plugins adding columns with an int value, but I dont see how to name columns properly.

src/web/controllers/admin/base_event_admin_controller.py
```diff
@@ -176,6 +185,11 @@ def _get_admin_event_render_context(
             case 'config':
                 pass
             case 'tournaments':
+                tournament_card_blocks = PM.plugin_manager.hook.get_tournament_card_block_template()
+                template_context |= {
+                    'paired_bye_result_options': cls._get_paired_bye_result_options(),
+                    "tournament_card_blocks": tournament_card_blocks
+                }
```
pascalaubry, line 192: Lines 193-195 do nothing.

src/plugins/ffe/ffe_database.py
```diff
@@ -30,10 +30,10 @@
     PlayerRatingType,
     PlayerGender,
     PlayerTitle,
-    PlayerFFELicence,
 )
 from database.access.ffe.ffe_access_database import FfeAccessDatabase
```
pascalaubry, line 34: I feel that keeping the same folder hierarchy in plugins would help (src/plugins/ffe/database/sqlite here).

src/plugins/hookspec.py
```diff
@@ -0,0 +1,66 @@
...
+hookimpl = pluggy.HookimplMarker(APP_NAME)
+
+class PrintSplitOption(NamedTuple):
+    name: str
+    url_name: str
+    split_fn: Callable[
+        [list[Player]], dict[str, list[Player]]
+    ]
+
+class ExtraColumn(NamedTuple):
```
pascalaubry, line 21: I wonder what will happen if two plugins add columns.

src/web/settings.py
```diff
@@ -129,8 +131,12 @@ def uptodate() -> bool:
         return contents, os.path.normpath(filename), uptodate
 
 
-template_dir: Path = BASE_DIR / 'src/web/templates'
+plugin_template_paths: list[Path] = plugin_manager.hook.get_templates_path()
```
pascalaubry, line 135: We have to keep in mind that homonymic templates may lead to undesired templates loaded.

src/web/settings.py
```diff
@@ -140,10 +146,11 @@ class PapiWebEnvironment(Environment):
 
     def __init__(
         self,
-        directory: Path,
+        directories: list[Path],
     ) -> None:
```
pascalaubry, line 150: Remove this line.

pascalaubry (changes requested):
Did not really test yet.

src/plugins/ffe/ffe.py
timothyarmes: It's not a hook, just a function

src/plugins/ffe/ffe.py
timothyarmes: Not sure what you mean, you can add several columns, they're inserted at appropriate index based on the core columns.  The function is intended to be generic for multiple views, so name would be harder.

src/plugins/hookspec.py
timothyarmes: It works perfectly.

src/web/settings.py
timothyarmes: I've used a prefix naming convention

src/plugins/ffe/ffe.py
pascalaubry: I was thinking of several plugins adding columns: in which order?
Anyway not time to think of that now ;-)

src/web/settings.py
pascalaubry: Prefixing the names of the templates by the name of the plugin is a good idea.

src/plugins/ffe/ffe_database.py
timothyarmes: I think we'll just end up with folders containing one file.  Let's tidy up the plugin folder once we know what we're working with

pascalaubry (approved):
OK for me.
As architecture is concerned, I would like @Amaras to have a look at it also.

Amaras (approved):
Code looks good, architecture seems fine

PR #195, opened 2025-03-08
Fixed result entry modal dismiss

pascalaubry:
JS code has been forgotten.


pascalaubry:
@timothyarmes The hover effect is back when removing the lstrip_blocks and trim_blocks parameters, but you probably had a good reason to add them. So I let you see ;-)

timothyarmes (approved):

PR #196, opened 2025-03-08
Fix print values

Amaras:


timothyarmes (approved):
Finding it hard t follow how all this works, but it's working under test

Amaras:
> Finding it hard t follow how all this works, but it's working under test

Yeah, I don't know why the players are not updated properly, so I simply returned them to be sure.

PR #198, opened 2025-03-09
Removed papi_web.ini

pascalaubry:
Removed papi_web.ini (now use an application-level database)
Moved the application version from PapiWebConfig to StoredConfig
The web port is not configured, a list of pre-defined ports are tried at startup
Forcing the edition of the settings is possible by setting StoredConfig.force_edit to True
Updated federation flags
@Molrn I think that some of the migration code can be factorized between event and config


Molrn (commented):
- Factoring between Config and Event migrations definitely needed
- access in the toolbar a bit excessive in my opinion (suggestion: cog icon only, aligned right of the toolbar)
image
 - for PRs with a lot of changes, better to keep the features separated into multiple PRs (ex: federation flags should not be here)

src/database/sqlite/config/config_database.py
```diff
@@ -0,0 +1,157 @@
...
+class ConfigDatabase(SQLiteDatabase):
+    """
+    The SQLite database class for Papi-web config.
+    """
+
+    # The current Papi-web version
+    papi_web_version: Version = Version('2.4.24')
+
+    # The file holding the configuration of the application.
+    config_database_path: Path = Path('.scc')
```
Molrn, line 29: why this naming ? I'm not familiar with it, and I can't find it used the same way in other projects. Why not a clearer name, like ``config.db` ?

pascalaubry: scc stands for Sharly Chess Config.

pascalaubry:
Back to draft to factorize the migration code with @Molrn

Molrn:
factorize done, it represents much more changes than I expected. Hope this is not using a bazooka to kill a fly, I prioritized having as few code duplication as possible. @pascalaubry I'd like to have your opinion on this

src/common/papi_web_config.py
```diff
@@ -70,313 +63,90 @@ def __init__(self):
...
-            print_interactive_error(
-                _('Could not write to {ini_file}: {ex}.').format(
-                    ini_file=self.config_file, ex=ex
-                )
-            )
+    def reload(self):
+        self.stored_config = self.load()
+
+    def load(self) -> StoredConfig:
+        from database.sqlite.config.config_database import ConfigDatabase
```
pascalaubry: Is there a real need to import rom the function (called at any run)?

src/database/sqlite/config/config_database.py
```diff
@@ -0,0 +1,113 @@
...
+        self.execute(
+            'UPDATE `info` SET `version` = ?',
+            (f'{version.major}.{version.minor}.{version.micro}', ),
+        )
+        self._version = version
+
+    @property
+    def migration_manager(self) -> AbstractMigrationManager:
+        return ConfigMigrationManager()
+
```
pascalaubry, line 58: I would expect this method in v2_00_00.py, don't you?

src/database/sqlite/event/event_database.py
```diff
@@ -52,9 +60,51 @@ def __init__(
...
+            'UPDATE `info` SET `version` = ?, `last_update` = ?',
+            (f'{version.major}.{version.minor}.{version.micro}', time.time()),
+        )
+        self._version = version
+
+    @property
+    def migration_manager(self) -> AbstractMigrationManager:
+        return EventMigrationManager()
+
+    def insert_creation_values(self):
```
pascalaubry, line 86: Same remark as for config_database.py

pascalaubry:
@Molrn Everything is clear for me now.
Juste have a look at my comment on papi_web_dev.py ~~and make it ready for review of OK for you~~.
Wait wait... Something must be done in utils to migrate both event and config.

src/common/papi_web_config.py
Molrn: the local import of Config Database is required to prevent circular import. The other solution I see to do this is to put back the version into its previous state (hard-coded  in PapiWebConfig) instead of reading it from the .scc DB. Version has to be hardcoded somewhere, I don't think it is that much more relevant to have it hard-coded in ConfigDatabase than in PapiWebConfig. It is in my opinion confusing, as other hard-coded values are part of PapiWebConfig

Molrn:
> access in the toolbar a bit excessive in my opinion (suggestion: cog icon only, aligned right of the toolbar)  

before moving to non-draft you mind taking care of this point we discussed ? @pascalaubry

utils/event_databases/migrate.py
```diff
@@ -37,38 +38,127 @@
...
-        with EventDatabase(event_id, True, False) as event_database:
-            if event_database.version == version:
-                print_interactive_info(
-                    f'Database [{event_id}] already at version {version}'
+    config_versions: list[Version] = []
+    event_versions: list[Version] = []
+    config_migration_manager: ConfigMigrationManager = ConfigMigrationManager(True)
+    event_migration_manager: EventMigrationManager = EventMigrationManager(True)
+    if args.version:
+        print(args.version)
```
Molrn: remaining print

utils/event_databases/migrate.py
```diff
@@ -37,38 +38,127 @@
...
+        for v in args.version:
+            if v == 'current':
+                if config:
+                    config_versions.append(SQLiteVersionedDatabase.papi_web_version)
+                if event_ids:
+                    event_versions.append(SQLiteVersionedDatabase.papi_web_version)
+            else:
+                version = Version(v)
+                max_version = max(version, SQLiteVersionedDatabase.papi_web_version)
```
Molrn: unused variable

utils/event_databases/migrate.py
```diff
@@ -37,38 +38,127 @@
...
+                if config:
+                    config_versions.append(
+                        max(
+                            min(
+                                Version(v),
+                                SQLiteVersionedDatabase.papi_web_version,
+                                config_migration_manager.last_migration_version
+                            ),
+                            config_migration_manager.first_migration_version,
```
Molrn: last_migration_version est toujours supérieur à first_migration_version

utils/event_databases/migrate.py
```diff
@@ -37,38 +38,127 @@
...
+                        max(
+                            min(
+                                Version(v),
+                                SQLiteVersionedDatabase.papi_web_version,
+                                event_migration_manager.last_migration_version
+                            ),
+                            event_migration_manager.first_migration_version,
+                        )
+                    )
+    else:
```
Molrn: calculs de version pas nécessaires peu nécessaires à mon avis: avoir un message d'erreur quand tu mets une mauvais version est déjà pris en charge par la classe, et migrer vers une autre version sans explication laisse penser à un bug. faut aller au plus simple: `.append(Version(v))`

utils/event_databases/migrate.py
```diff
@@ -37,38 +38,127 @@
...
+        if event_ids:
+            event_versions.append(
+                max(
+                    min(
+                        SQLiteVersionedDatabase.papi_web_version,
+                        event_migration_manager.last_migration_version
+                    ),
+                    event_migration_manager.first_migration_version,
                 )
+            )
```
Molrn: en amont: `versions = args.version or ['current']`, ça te permet de supprimer toute cette partie. Tout ce qui permet de réduire la complexité et les répétitions dans le code, on prend

src/common/papi_web_config.py
pascalaubry: Resolved by moving papi_web_version to SqliteVersionedDatabase

utils/event_databases/migrate.py
pascalaubry: Removed.

utils/event_databases/migrate.py
pascalaubry: Removed.

utils/event_databases/migrate.py
pascalaubry: Avant simplification, si j'exécute migrate.py --event=35-tc --config --version=0.0.0 --version=current, obtiens :
image
Après simplification j'obtiens :
image
J'ai essayé de simplifier au max avec tes remarques, n'hésite pas à rectifier directement en gardant le même comportement si tu penses que c'est mieux.

utils/event_databases/migrate.py
pascalaubry: Cf ci-dessous.

utils/event_databases/migrate.py
pascalaubry: Cf ci-dessous.

pascalaubry:
> * Factoring between Config and Event migrations definitely needed
> * access in the toolbar a bit excessive in my opinion (suggestion: cog icon only, aligned right of the toolbar)
>   image
> * for PRs with a lot of changes, better to keep the features separated into multiple PRs (ex: federation flags should not be here)

I will do it on another PR.

Molrn (dismissed):

src/data/player.py
```diff
@@ -210,7 +210,8 @@ def __init__(
         self.time_control_initial_time: int | None = None
         self.time_control_increment: int | None = None
         self.time_control_modified: bool | None = None
-        self._tournament_ref: 'ReferenceType[Tournament] | None' = tournament
+        self._tournament_ref: 'ReferenceType[Tournament] | None' = None
+        self.tournament = tournament
```
pascalaubry: Also partially fixed in #211 but I had to fix it here to test.

timothyarmes (changes requested):
I have the default federation set to FRA.

When I create an event, the dropdown is still set to FID.

timothyarmes (dismissed):

timothyarmes (approved):

PR #204, opened 2025-03-10
Allow addition of tournament exporters via plugins

Molrn:


src/plugins/hookspec.py
```diff
@@ -107,3 +108,6 @@ def get_extra_print_view_columns(self, document: PrintDocument) -> Iterable[Iter
     def get_extra_screen_columns(self, screen: ScreenType) -> Iterable[Iterable[ExtraColumn] | None]:
         """Provide extra columns for the print view"""
 
+    @hookspec
+    def get_extra_tournament_exporters(self) -> Iterable[AbstractTournamentExporter]:
```
timothyarmes: I think this should be Iterable[Iterable[AbstractTournamentExporter]].

Each plugin can return a list of AbstractTournamentExporter.  The hook returns an array of results for each plugin;:
[[plugin1_exporter1, plugin1_exporter2], [plugin2_exporter1, plugin2_exporter2]]

timothyarmes: maybe you're right though, and we should document the return type for the plugin, not the hook itself?
I thinks it's completely ignored by Pluggy

Molrn: my bad I commited before seeing your second comment ^^' I've looked up the pluggy doc and usages and encountered both, without any effect, the @hookspec decorator seems to cancel the effects of typing. I think the most important is for for it to be consistant in our code, and either choose the typing that @hookimpl should return or the one returned when calling the hook.  

I think I have a preference for  using the @hookimpl typing (i.e without the additional iterable) because we are more likely to mess up types when adding a new implementation of the hook than when first calling the hook. If you're ok with it, I'd like to modify it for every other @hookspec

Molrn: just added a commit to modify the typing on every hook as I described, if it does not suit you I'll revert it

timothyarmes (approved):

src/plugins/hookspec.py
timothyarmes: I agree, it's more useful

PR #211, opened 2025-03-11
Fix weak ref

timothyarmes:


src/data/player.py
```diff
@@ -210,7 +210,7 @@ def __init__(
         self.time_control_initial_time: int | None = None
         self.time_control_increment: int | None = None
         self.time_control_modified: bool | None = None
-        self._tournament_ref: 'ReferenceType[Tournament] | None' = tournament
+        self.tournament = tournament
```
pascalaubry: We should keep self._tournament_ref initialized in the constructor:
self._tournament_ref: 'ReferenceType[Tournament] | None' = None
And then set the tournament:
self.tournament = tournament

pascalaubry (changes requested):

pascalaubry (approved):

Amaras (approved):

PR #213, opened 2025-03-11
Move initial ratings to plugin

timothyarmes:


src/plugins/ffe/ffe.py
```diff
@@ -222,6 +222,40 @@ def augment_player_after_search(player: Player):
                 }
 
 
+@hookimpl
+def set_player_default_ratings(federation: str, player: 'Player'):
+    if federation == 'FRA':
```
Molrn: for big if statements with no else and no other actions, it is a python coding standard to prefer:
```
if not {condition}:
    return # or break, or continue
```
this reduces global indent and line length

Molrn (changes requested):
looking good :)

Molrn (approved):

PR #215, opened 2025-03-12
Plugin-specific migrations

Molrn:
Allow plugins to have migrations. In addition to the standard event migration feed, one feed is added per plugin. When auto-migrating, the standard feed upgrades the database to the latest papi-web version, then all plugins are upgraded to their latest version.  

Small work-around I had to do to move the pre-existing fields for chessevent and ffe plugins to these migrations without losing data (i.e not just delete / recreate):
- rename all fields as `deprecated_{field}` in a standard database migration
- from plugin migration, rename `deprecated_{field}` to `{field}`, or create `{field}` if `deprecated_{field}` does not exist   

This is intended to be a temporary solution, as otherwise this would mean that users not using the plugins would still have the deprecated fields in their databases. However, we can't yet delete these fields from the standard migrations, as this would mean the data would be lost.  

I suggest the following: wait until the next papi-web release, then add a migration deleting all deprecated fields. This would mean users of version 2.4.18 will keep their chessevent/ffe data if they install the next release (2.5.0?). However, users skipping from 2.4.18 to the next next release (ex: 2.6.0) will lose this data.  

N.B: the branch is based on #198, so I'm keeping it as draft until it is merged  
N.B2: to follow the guideline to prefix all plugin fields with the plugin name, I renamed the following:
- last_chessevent_download_md5 -> chessevent_last_download_md5
- last_ffe_upload -> ffe_last_upload
- last_ffe_rules_upload -> ffe_last_rules_upload


Molrn:
replaced by #219

PR #221, opened 2025-03-12
FFE specific sort key for players

timothyarmes:


src/plugins/hookspec.py
```diff
@@ -137,7 +137,11 @@ def clear_player_filters(self, request: HTMXRequest):
     @hookspec
     def filter_player(self, web_context: HTMXRequest, player: Player) -> bool:
         """Returns True if the player should be in the admin player list, False otherwise """
-
+        
+    @hookspec(firstresult=True)
+    def player_club_sort_key(self, player: Player) -> Callable[[Player], tuple]:
```
Molrn:
```suggestion
    def player_club_sort_key(self, player: Player) -> tuple:
```

src/plugins/ffe/ffe.py
```diff
@@ -227,6 +227,17 @@ def filter_player(web_context: HTMXRequest, player: Player) -> bool:
     )
 
 
+@hookimpl
+def player_club_sort_key(player: Player):
```
Molrn, line 230:
```suggestion
@hookimpl
def player_club_sort_key(player: Player) -> tuple:
```

Molrn (dismissed):
github suggestions work well, but only when the selection does not include a removed line... still interesting tho

Molrn (approved):
LGTM

PR #226, opened 2025-03-12
Add engines to plugin architecture

Molrn:
not the most useful, but at least it removes the ffe and chessevent dirs from /src


timothyarmes (approved):
LGTM

PR #227, opened 2025-03-12
Use collapsable for screen types

timothyarmes:


src/web/controllers/admin/event_admin_controller.py
```diff
@@ -337,9 +337,12 @@ async def htmx_admin_event_tab(
                 }.items():
                     if param is not None:
                         if param:
-                            screen_types.append(screen_type)
+                            screen_types.add(screen_type)
```
Molrn, line 340: screen_types typed as `list[str]` -> `set[str]`

src/web/static/css/base.css
```diff
@@ -386,4 +386,17 @@ body[data-bs-theme=light] .btn-primary {
...
+    background-color: transparent;
+}
+
+button.collapse-button .bi-chevron-down {
+    transition: all 0.3s ease;
+}
+
+button.collapse-button[aria-expanded="true"] .bi-chevron-down {
+    transform: rotate(-90deg);
 }
```
Molrn: EOF newline

Molrn (commented):
image  
with a collapse, not sure the tooltip should stay  

image

Is that normal or a bug ? caused by:  
```html
{% for screen_set in screen.screen_sets_sorted_by_order %}
                <div>
                    {{ screen_set.tournament.name }}
                    ({{ screen_set.numbers_str }})
                </div>
            {% endfor %}
```
might be a duplicate introduced in `screen_sets_sorted_by_order` when `all boards` are selected, could you look it up ?

src/web/session.py
```diff
@@ -165,16 +165,16 @@ def get_session_admin_events_show_details(cls, request: HTMXRequest) -> bool:
...
 
     @classmethod
-    def get_session_admin_screens_screen_types(cls, request: HTMXRequest) -> list[str]:
-        return request.session.get(
+    def get_session_admin_screens_screen_types(cls, request: HTMXRequest) -> set[str]:
+        return set(request.session.get(
             cls.ADMIN_SCREENS_SCREEN_TYPES_KEY,
             ['boards', 'input', 'players', 'results', 'ranking', 'image', ],
-        )
+        ))
```
pascalaubry, line 177: Maybe 
return request.session.get(
            cls.ADMIN_SCREENS_SCREEN_TYPES_KEY,
            ('boards', 'input', 'players', 'results', 'ranking', 'image', ),
)

src/web/controllers/admin/event_admin_controller.py
pascalaubry: No problem for me, lists are not needed here (and no error on my PyCharm).

pascalaubry (changes requested):
Unused import in event_admin_controller.py (PlayerFFELicence) and a little improvement in session management.
Accordions are clearly nicer that the switches, and more intuitive for users.
However I dont feel comfortable with the look, I was expecting full-width accordions (or at least accordions of the same width).

src/web/session.py
timothyarmes: I don't understand.  If you to that you no longer get a set back.

I tried it, it breaks.

pascalaubry: You are right: type-cast to set is needed, sorry.

src/web/controllers/admin/event_admin_controller.py
Molrn: fixed by @timothyarmes in https://github.com/papi-web-org/papi-web/pull/227/commits/37cfaa3af19e6a5b79e2ab814d846aa2e96d3b58. state of my pycharm before:  
image

pascalaubry (approved):
No need to wait now, the rest can be solved later by HTML/CSS changes.

PR #228, opened 2025-03-12
Put papi-web version in a sensible place

Amaras:


Molrn:
good place to put it, same should probably be done with the plugin versions. If that's ok I'm adding the remapping for plugin versions to this PR

Molrn (approved):
LGTM

PR #229, opened 2025-03-12
Add visible URLs

timothyarmes:


pascalaubry (commented):
From our previous discussions about URLs, I identified much more URLs that can be considered as 'navigation URLs'.
Can you check https://github.com/papi-web-org/papi-web/blob/dev/docs/81-urls.md ?

pascalaubry (dismissed):
Works pretty well with the URLs that are pushed now, but when navigation URLs are not pushed then the previous URL remains and it is not relevant.
To reproduce for example (you can with any 'back' button):
1. Load the home page
2. goto /user (the URL is shown)
3. go back home (the /user URL is still shown)

Amaras:
> Works pretty well with the URLs that are pushed now, but when navigation URLs are not pushed then the previous URL remains and it is not relevant. To reproduce for example (you can with any 'back' button):
> 
> 1. Load the home page
> 2. goto /user (the URL is shown)
> 3. go back home (the /user URL is still shown)

Can't reproduce, make sure you fully refreshed the clients

timothyarmes:
@pascalaubry I'd missed some /user routes, added now.
I also couldn't reproduce your "back" button issue, but then I realised that you meant the "Home" button.  Fixed.

src/web/templates/user/index.html
```diff
@@ -5,7 +5,8 @@
         class="btn btn-link link-dark fs-3 fw-bold px-1 py-0"
         hx-get="{{ url_for('index') }}"
         preload
-        hx-target="body"
+        hx-target="body"`
```
pascalaubry: Strange character at the end of line.

pascalaubry (dismissed):
Saw that you added the push-url="true" to all the 'Go back to' buttons so verything should work for me (can't test right now).
Please just take care of the character in user/index.html.
Thanks.

Molrn (approved):
LGTM

PR #230, opened 2025-03-12
Restored deleted code in event_database.py

pascalaubry:
The bug was happening when creating example databases on startup.


Molrn (approved):
LGTM

PR #233, opened 2025-03-13
Move tournament FFE fields to plugin

timothyarmes:


src/database/sqlite/event/event_database.py
```diff
@@ -466,10 +464,6 @@ def _populate(self):
                                 path=None,
                                 filename=tournament_dict.get('filename', None),
                                 name=tournament_dict.get('name', None),
-                                ffe_id=tournament_dict.get('ffe_id', None),
-                                ffe_password=tournament_dict.get(
-                                    'ffe_password', None
-                                ),
```
Molrn, line 472: just deleting this implies that the new tournaments populated through yml files will not have an ffe id or password. More than this, it seems to have broken the example database install  
image  
to reproduce: remove all .db files from your /events dir, then run

src/database/sqlite/event/event_database.py
```diff
@@ -1421,13 +1412,12 @@ def _row_to_stored_tournament(cls, row: dict[str, Any]) -> StoredTournament:
...
             last_check_in_update=row['last_check_in_update'],
-            ffe_last_upload=row['ffe_last_upload'],
-            # needed to open event databases when version < 2.4.11 before checking the version
-            ffe_last_rules_upload=row.get('ffe_last_rules_upload', 0.0),
             chessevent_last_download_md5=row['chessevent_last_download_md5'],
             # needed to open event databases when version < 2.4.23 before checking the version
             tie_breaks=cls._load_tie_breaks_from_database_field(row.get('tie_breaks', None)),
         )
+        plugin_manager.hook.augment_tournament_after_db_fetch(stored_tournament=stored_tournament, row=row)
+        return stored_tournament;
```
Molrn, line 1420:
```suggestion
        return stored_tournament
```

src/database/sqlite/event/event_database.py
```diff
@@ -1450,14 +1440,15 @@ def _write_stored_tournament(
         self,
         stored_tournament: StoredTournament,
     ) -> StoredTournament:
+        per_plugin_player_data = plugin_manager.hook.tournament_data_for_db_write(stored_tournament=stored_tournament)
+        plugin_data = { key: value for data in per_plugin_player_data for key, value in data.items() }
```
Molrn, line 1444:
```suggestion
        plugin_data = {key: value for data in per_plugin_player_data for key, value in data.items()}
```
no space after a `{` or before a `}` is a coding standard:  
image

src/plugins/ffe/ffe.py
```diff
@@ -111,6 +113,42 @@ def player_data_for_db_write(player: Player) -> dict[str, Any]:
...
+@hookimpl
+def augment_tournament_after_db_fetch(stored_tournament: 'StoredTournament', row: dict[str, Any]):
+    if not stored_tournament.plugin_data:
+        stored_tournament.plugin_data = {}
+    stored_tournament.plugin_data[PLUGIN_NAME] = {
+        'ffe_id': row.get('ffe_id', ''),
+        'ffe_password': row.get('ffe_password', ''),
+        'ffe_last_upload': row.get('ffe_last_upload', 0.0),
+        'ffe_last_rules_upload': row.get('ffe_last_rules_upload', 0.0),
+    }
```
Molrn, line 125: It seems redundant to both prefixing fields with the plugin name and store values in a separate dict of the plugin_data.  
`plugin_data` should either look like `{"ffe": {"id": 1}}` or `{"ffe_id": 1}` in my opinion. The guideline to prefix field with the plugin name is for database names. if you want to keep the same name everywhere (db field, form name, etc...) which I understand, plugin_data should be a `dict[str, Any]`

src/plugins/ffe/templates/ffe_tournament_form_fields.html
```diff
@@ -0,0 +1,23 @@
...
+    <div class="col-12 col-md-6 mb-3">
+        {{ admin_macros.password_input(
+            name='ffe_password',
+            label=_('FFE password:'),
+            placeholder=_('E.g.: %(string)s', string='KYNPRTZMHX'),
+            help_text=_('The password is made of 10 uppercase letters, it is used to connect to the FFE website.'),
+            data=data, errors=errors
+        ) }}
+    </div>
+</div>
```
Molrn: EOF newline

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -223,6 +212,10 @@ def _admin_validate_tournament_update_data(
                         )
                     )
 
+        # Have plugins validate their fields and return private plugin data
+        per_plugin_tournament_data = plugin_manager.hook.get_validated_tournament_form_fields(action=action, tournament=web_context.admin_tournament, data=data, errors=errors)
+        plugin_data = { key: value for data in per_plugin_tournament_data for key, value in data.items() }
```
Molrn:
```suggestion
        plugin_data = {key: value for data in per_plugin_tournament_data for key, value in data.items()}
```

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -376,6 +362,10 @@ def _admin_event_tournaments_render(
                             pass
                         case _:
                             raise ValueError(f'action=[{action}]')
+                    
+                    per_plugin_form_data = plugin_manager.hook.get_tournament_form_data(tournament=web_context.admin_tournament)
+                    plugin_form_data = { key: value for data in per_plugin_form_data for key, value in data.items() }
```
Molrn:
```suggestion
                    plugin_form_data = {key: value for data in per_plugin_form_data for key, value in data.items()}
```

Molrn (changes requested):
uncaught references to ffe fields:
-  macro jinja 2 `ffe_license`
- create_ffe.sql (to refactor)

Besides that looks good ;) one major point to discuss about the format of plugin_data and a few changes

src/plugins/ffe/ffe.py
timothyarmes: I thought about this, but I don't actually agree.

All data associate with the plugin is under its name.  That makes sense.

Now, if I just do this: 

```
{
  ffe: {
    id:  "xxx",
  }
}
```

Then what's the ID of?  Could be anything, we might need to store multiple IDs.  It's your FFE id, so it should be called `ffe_id`.

What about league?  Well league isn't tied to the FFE, so it isn't prefixed.

src/database/sqlite/event/event_database.py
timothyarmes: I think it's going too far to allow plugin specifics in the example yaml files.

pascalaubry: Another question is: should we limit the possibility of creating example databases to developers only?

Molrn: > I think it's going too far to allow plugin specifics in the example yaml files.  

then ffe and chessevent fields should be deleted from yaml files. as they contain credentials, they should never have been there in the first place   

> Another question is: should we limit the possibility of creating example databases to developers only?   

if we have TRF and papi import, I think examples databases would no longer be needed. In the mean time, yes, it would be good

src/plugins/ffe/ffe.py
Molrn: > What about league? Well league isn't tied to the FFE, so it isn't prefixed.  

I don't know about that, a league from the FFE plugin is definitely a ffe league... I think code consistency is important for that kind of data, so not having them all formatted the same way is not a good idea for me.   

> All data associate with the plugin is under its name. That makes sense.  

I agree, that makes sense, but is it necessary ? if all fields are prefixed, we don't need to worry about key duplication, so a format like this is already enough: 
```json
{
    "ffe_id": 1,
    "ffe_password": "password";
    "chessevent_password": "password",
}
```
If you still disagree with that I'm ok with whatever format you'll choose, but league not being prefixed if every other field is prefixed is something I'd really like to avoid

timothyarmes: We'll come back to the subject later, let's not block this PR, because teh chess event one is on its way

Molrn (approved):
LGTM, more later on the plugin_data format

PR #234, opened 2025-03-13
Solve problems identified by VSCode

Amaras:
This PR aims to reduce the number of problems identified by VSCode

Problems identified are, among others:
  1. Trailing whitespace
  2. Line too long
  3. Type problems
  4. Missing docstrings

There are 477 problems remaining in `src/data`, 43 not related to a missing docstring


src/data/family.py
```diff
@@ -175,7 +176,6 @@ def last_update(self) -> float | None:
     def last_update_str(self) -> str | None:
         return format_timestamp_date_time(self.last_update)
 
-    @cache
```
pascalaubry, line 178: IMO It is important to cache this computation.
At the opposite, maybe properties caclculted_*() may not be cached if _calculate_screens is.

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -721,8 +721,8 @@ def get_players_datasheet_columns(cls) -> list[str]:
         
         # Add plugin columns
         extra_columns = EventAdminController.get_players_datasheet_extra_columns()
-        for index, columns in extra_columns.items():
-            header_columns[index:index] = [column.title for column in columns]
+        for index, column in extra_columns.items():
+            header_columns[index:index] = column.title
```
pascalaubry: Not sure about this one, I dont understand how it could work before.
(I did the change)

pascalaubry (dismissed):
Added some more errors raised by PyCharm.
One error is not fixed in this PR but it needs another class, I keep for another PR when this one is merged.
OK for me with the change done in EventAdminController.get_players_datasheet_columns() and if you are OK with restoring the cache for Family._calculate_screens().

pascalaubry (dismissed):
Family.get_players_datasheet_columns()

src/data/family.py
Amaras: The problem is that caching keeps instances alive, so this might cause part of our memory leak (if the screens are recomputed regularly).
Either we do a `cached_property`, or make the computation a cached `staticmethod` (instances are not kept alive).

Amaras:
This is absolutely not ready for review, I intend on solving all problems not related to missing docstrings and lines too long (only when dealing with i18n strings) before putting it up for review

pascalaubry:
Did not see that it was still a draft. Waiting then.

src/data/family.py
pascalaubry: I understand, and I plan to take some time this week-end to track other possible sources of the leaks.
I thought this method used @cached_property not @cache, @cached_property should replace @cache.

src/web/controllers/admin/event_admin_controller.py
pascalaubry: Undone.

timothyarmes (approved):
LGTM

PR #235, opened 2025-03-13
Improved database tests with searches

pascalaubry:


utils/test/test_databases.py
```diff
@@ -1,19 +1,102 @@
...
+    }
+    sql_commands_list: list[dict[str, str]] = [
+        {},
+        name_sql_commands,
+        id_sql_commands,
+        name_sql_commands | id_sql_commands,
+    ]
+    for sql_commands in sql_commands_list:
+        FideDatabase().delete()
+        t: float = time.time()
```
Amaras: Don't use `time.time()` for benchmarks, use `time.perf_counter()`

utils/test/test_databases.py
```diff
@@ -1,19 +1,102 @@
...
+        print('Creating the FIDE database... ')
```
Amaras, line 40: Don't count printing inside the timing

utils/test/test_databases.py
```diff
@@ -1,19 +1,102 @@
...
+        if FideDatabase().create():
+            print(f'Done in {time.time() - t:.2f} seconds.')
```
Amaras: I'd rather you used an `end` variable and print from there, to be sure that only the creation is timed and to be less confusing

utils/test/test_databases.py
```diff
@@ -1,19 +1,102 @@
...
+                print(f'{time.time() - t:.2f} seconds.')
+            print(f'Database size: {int(FideDatabase().file.lstat().st_size / 1024 / 1024)}Mb')
+            if test_names:
+                t: float = time.time()
+                print(f'Performing {searches} name searches on the FIDE database (opening and closing each time)... ', end='')
+                for _ in range(searches):
+                    token: str = random_search_token()
+                    with FideDatabase() as fide_database:
+                        #print(f'Token [{token}]: {len(list(fide_database.search_player(random_search_token(), limit=limit)))}')
+                        fide_database.search_player(random_search_token(), limit=limit)
```
Amaras: you're not using the same tokens, why?

Amaras (dismissed):
Benchmarking is hard, I'm not sure we're properly testing our use case there, but the testing is more complete than it was

utils/test/test_databases.py
pascalaubry: Done.

utils/test/test_databases.py
pascalaubry: Done.

utils/test/test_databases.py
pascalaubry: Done.

utils/test/test_databases.py
pascalaubry: No reason... ;-)

utils/test/test_databases.py
```diff
@@ -1,19 +1,106 @@
...
+    if FfeDatabase().create():
+        print(f'Done in {int(time.time() - start)} seconds.')
+        print(f'Performing {searches} searches on the FFE database...')
+        start: float = time.time()
+        for _ in range(searches):
+            token: str = random_search_token()
+            with FfeDatabase() as ffe_database:
+                print(f'Token [{token}]: {len(list(ffe_database.search_player(random_search_token(), limit=limit)))}')
+        duration: float = time.perf_counter() - start
+        print(f'Done in {duration:.2f} seconds.')
```
Amaras: At least do the same as above with this part

Amaras (dismissed):

utils/test/test_databases.py
pascalaubry: Forgot the FFE database... Done.

utils/test/test_databases.py
```diff
@@ -1,19 +1,107 @@
...
+    start: float = time.perf_counter()
+    if FfeDatabase().create():
+        duration: float = time.perf_counter() - start
+        print(f'Done in {duration} seconds.')
+        print(f'Performing {searches} searches on the FFE database...')
+        start: float = time.perf_counter()
+        for _ in range(searches):
+            token: str = random_search_token()
+            with FfeDatabase() as ffe_database:
+                ffe_database.search_player(random_search_token(), limit=limit)
```
Amaras: why regenerate tokens?

Amaras (dismissed):

Amaras (approved):

PR #239, opened 2025-03-14
Made SqliteDatabase protected methods _fetch_*() public for migration classes

pascalaubry:
Note: _execute() has been made public recently for the same reason.


Molrn (approved):
LGTM

PR #241, opened 2025-03-14
Misc fixes after plugin extraction

pascalaubry:


src/web/controllers/admin/base_admin_controller.py
```diff
@@ -814,11 +814,13 @@ def _admin_render(
                 locale_options[''] = _('By default - {option}').format(
                     option=locale_options[DEFAULT_LOCALE]
                 )
+                plugin_form_fields_templates = plugin_manager.hook.get_event_form_fields_template() or []
```
Molrn, line 817: hook not defined in the hookspecs

src/plugins/ffe/ffe.py
```diff
@@ -23,7 +22,7 @@
 
 
 from web.controllers.admin.base_event_admin_controller import BaseEventAdminWebContext
-import web.controllers.base_controller as WebContextModule
+import web.controllers.base_controller as web_context_module
```
Molrn: I don't think the module is necessary here, I think the classes used can be directly imported without losing any clarity: 
 
```suggestion
from web.controllers.base_controller import BaseController, WebContext
```

Molrn (changes requested):

src/web/controllers/admin/base_admin_controller.py
pascalaubry: Fixed.

src/plugins/ffe/ffe.py
pascalaubry: Done.

Molrn (approved):
LGTM

PR #242, opened 2025-03-14
Plugin tie breaks

Molrn:
- Move from Enum data definition to Abstract classes
- move all tie-break papi_legacy code to ffe plugin
- add a hook to fetch extra tie-breaks from plugins


src/data/tie_break.py
```diff
@@ -1,534 +1,1604 @@
...
+    def adjusted_score(
+        player: 'Player',
+        tournament: 'Tournament',
+        max_round: int | None = None,
+        adjust_fore: bool = False,
+    ) -> float:
+        """Computes the adjusted score of the player for the purposes of their opponents' tie-breaks
+        Only adjusts them in case of requested byes followed by all VUR.
+        If *adjust_fore* is True, the adjusted score for Fore Buchholz is computed:
+        games not already determined are considered a draw."""
```
Amaras:
```suggestion
        games for the last round not determined over the board are considered as draws."""
```

src/data/tie_break.py
```diff
@@ -1,534 +1,1604 @@
...
+        if max_round is None:
+            max_round = max(player.pairings)
```
Amaras: Should probably be the tournament's max ranking round.
For unfinished tournaments, this might cause problems.

src/data/tie_break.py
```diff
@@ -1,534 +1,1604 @@
...
+            return player.points_after(max_round)
+        score = 0
+        for round_index, pairing in player.pairings.items():
+            if round_index > max_round:
+                continue
+            if adjust_fore and round_index == max_round:
+                if pairing.result in (
+                    Result.FULL_POINT_BYE,
+                    Result.PAIRING_ALLOCATED_BYE,
+                ):
```
Amaras, line 252: Games not determined over the board use their result's points
```suggestion
                if pairing.result in (
                    Result.FULL_POINT_BYE,
                    Result.PAIRING_ALLOCATED_BYE,
                    Result.HALF_POINT_BYE,
                    Result.ZERO_POINT_BYE,
                ):
```

tests/test_tiebreaks.py
```diff
@@ -2,30 +2,35 @@
 from decimal import Decimal
```
pascalaubry, line 2: Is there a reason to have a /tests directory with one file or could it move to /utils/test?

src/plugins/ffe/tie_break.py
```diff
@@ -0,0 +1,355 @@
+from abc import ABC, abstractmethod
```
pascalaubry: Rename this file ffe_tie_break.py (as well as ffe_utils.py)?

pascalaubry (dismissed):
Everything works fine on this branch, and I saw nothing wrong.
@Amaras 's review is needed anyway.

tests/test_tiebreaks.py
Molrn: having it as a dir is a good thing in my opinion, it encourages to test other parts of the app. It is common for a python app to have a test dir outside of /src

pascalaubry: I don't want to move into /src, I was wondering if it could be moved to /utils/test.

Molrn: I'm a bit opposed to that, tests are not a utils. You could however have a utils files that runs all tests:
```sh
python -m unittest discover -s tests
```
this kind of command could also be listed in a makefile, alongside commands for linting for example

src/data/tie_break.py
```diff
@@ -1,534 +1,1620 @@
...
+    def name(self) -> str:
+        return _('Sonnenborn-Berger')
+
+    @property
+    def id(self) -> str:
+        return 'SONNENBORN_BERGER'
+
+    @property
+    def papi_id(self) -> str:
+        return 'Sonnenborn-Berger'
```
Amaras, line 974: Do we need to have `papi_id` defined in the core tie-breaks?
Couldn't it be monkey-patched or defined by the plugin?

tests/test_tiebreaks.py
pascalaubry: OK

Amaras (commented):
I have not looked at the changed code in detail.
I'll need to run the tests to be sure.
For now, a few remarks (see review comments)

src/data/tie_break.py
Molrn: It needs to, because it is the id used to insert the file in the papi-database. Once we stop using papi as the tournament storage, we'll be able to remove it. otherwise, all the database insertion would need to be handled by the plugin, which does not make much sense.  
As this usage was not obvious, I documented it in the original method definition. papi_id is also the method used to determine which tie breaks can be added in the tournament form

src/data/tie_break.py
Molrn: Actually except for the tests max_round is already set for each usage of adjusted_score. I suggest making the parameter mandatory and removing this part.

src/data/tie_break.py
Molrn: I commited this suggestion, but it fails the fore_buchholz tests. I'm reverting it for now, letting you handle this if that's ok

Amaras (approved):
Tie-breaks work as intended.
I have not tested the web interface, but trusting everyone else on that

PR #244, opened 2025-03-15
Removed file reference from POT/PO files

pascalaubry:
The aim of this PR is to limit the diff when moving strings from a file to another one.


Molrn (approved):
I agree I much prefer it that way. I haven't looked into the options of I18n, but with the same objective there's one thing that bugs me: the addition of unused strings to the end of the file like this:  
```
#~ msgid "Default federation"
#~ msgstr "Fédération par défaut :"
```
Do you know if we can remove this ? if so, do you want to, or it is important for you ?

PR #251, opened 2025-03-16
Remove remaining deprecated ffe fields

Molrn:
There was an error in the migration creating the ffe plugin, leading to a two deprecated fields remaining in the DB:  
- deprecated_last_ffe_rules_upload
- deprecated_last_ffe_upload  

This PR fixes it


pascalaubry (approved):
Works perfectly.

PR #253, opened 2025-03-16
SharedUtils sytsem to override functions from plugins

Molrn:
replaces #248

### DIFF
--- a/src/data/tournament.py
+++ b/src/data/tournament.py
@@ -22,15 +22,14 @@ from data.player import Player, Federation, Club
 from data.screen import Screen
 from data.tie_break import AbstractTieBreak
 from data.util import (
-    TrfType,
     BoardColor,
-    TournamentRating,
     PlayerGender,
     PointValueType,
-    performance_bonus,
-    round_fide,
     TournamentPairing,
-    Result
+    Result,
+    SharedUtils,
+    TournamentRating,
+    TrfType,
 )
 from database.access.papi.papi_database import PapiDatabase
 from database.sqlite.event.event_database import EventDatabase
@@ -462,9 +461,8 @@ class Tournament:
         self,
         trf_type: TrfType,
         first_round_pairing: BoardColor = BoardColor.WHITE,
-        papi_legacy: bool = True,
     ) -> TrfTournament:
-        self.compute_player_ranks(after_round=self.max_ranking_round, papi_legacy=papi_legacy)
+        self.compute_player_ranks(after_round=self.max_ranking_round)
         return TrfTournament(
             name=self.name,
             city=self.location,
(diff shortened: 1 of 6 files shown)


src/data/tournament.py
```diff
@@ -743,12 +737,14 @@ def estimate_players(self, *, max_round: int | None = None, papi_legacy: bool =
             estimation = level_estimations[points]
             if estimation > 0:
                 # No need to touch a group's estimation if it already has one
-                previous_bonus = round_function(
-                    performance_bonus(points / max_possible_points, papi_legacy=papi_legacy)
+                previous_bonus = SharedUtils.performance_bonus(
```
Amaras: be careful, `performance_bonus` may return a `float` (namely for papi compatibility)

src/plugins/ffe/ffe_tie_break.py
```diff
@@ -1,12 +1,28 @@
...
 from common.i18n import _
 from data.pairing import Pairing
 from data.player import Player
 from data.tie_break import AbstractTieBreak
 from data.tournament import Tournament
-from data.util import TournamentPairing, Result, performance_bonus
+from data.util import TournamentPairing, Result, StaticUtils
+
+
+def papi_performance_bonus(fractional_score: float) -> int:
```
Amaras:
```suggestion
def papi_performance_bonus(fractional_score: float) -> int | float:
```

Amaras (commented):
Papi bonus computations return `double`s (in C# decompiled code), not always `int`s.

src/data/tournament.py
Molrn: my bad got fooled by the typing of performance_bonus ^^' fixed ;)

src/plugins/ffe/ffe_tie_break.py
```diff
@@ -1,12 +1,28 @@
...
 from common.i18n import _
 from data.pairing import Pairing
 from data.player import Player
 from data.tie_break import AbstractTieBreak
 from data.tournament import Tournament
-from data.util import TournamentPairing, Result, performance_bonus
+from data.util import TournamentPairing, Result, StaticUtils
+
+
+def papi_performance_bonus(fractional_score: float) -> int | float:
```
Amaras, line 15: Maybe this function (and all similar functions/methods) could be `@cached`.
Not sure about it, though

Amaras (commented):
Overall good code.
I want to have the conversation about caching before I approve this, however

src/plugins/ffe/ffe_tie_break.py
Molrn: never used function caching before, but it seems appropriate in this case. I think a safe usage would be to clear the cache  before and after a tournament's rankings have been computed, as this is the use case for bulk computations with similar values (switching to another usage afterwards will probably have another number of rounds so the cache would be useless).  

besides performance_bonus functions, do you have other functions in mind that uses a player's score and that could be cached ?

Molrn: tried clearing before / after ranking, it gets me slightly worst computation time. I simply set a cache with a limit

Amaras (approved):

PR #255, opened 2025-03-17
Simplify session data

timothyarmes:


pascalaubry (changes requested):
The changes force to declare session data in the controllers that use it, not taking into account the fact that some data may be shred between controllers (and making a little harder to move things between controllers).
I feel that the complexity level of session data is not such that it is needed to split it.

PR #257, opened 2025-03-17
Use Python 3.12

Amaras:
Python 3.11 doesn't have any bugfix updates (with installer) since 3.11.9 released on 2024-04-02 (nearly a year ago), see [PEP 664: Pyhton 3.11 release schedule](https://peps.python.org/pep-0664/).
Only source-only security fixes are available.

This PR simply changes the required Python version to `>= 3.12.0` in our `pyproject.toml`.
Python 3.12 is expected to have one last bugfix version on 2025-04-08.
Here is the changelog for Python 3.12: https://docs.python.org/3.12/whatsnew/3.12.html

I also expect to change Python versions every year.
I am proposing to change to version 3.n+1 every October, after the last bugfix of version 3.n is released (when version 3.n+2 is released).

For users, this should be completely transparent.
For devs (us), this means we have to recreate our virtual environments, however we will have better features.

For future reference, this means in October this year (2025), we will change from Python 3.12 to Python 3.13, once Python 3.14 is released (even if the last bugfix release will be in April 2025).

### DIFF
--- a/docs/86-i18n.md
+++ b/docs/86-i18n.md
@@ -7,6 +7,7 @@
 <!-- DO NOT EDIT! (START) -->
 | Locale | Messages | Empty | Empty mandatory | [ai_translation] | [fuzzy] | PO file | Translators |
 |--|:--:|:--:|:--:|:--:|:--:|:--:|:--:|
+<<<<<<< Updated upstream
 |<img src="../src/web/static/images/locales/en.svg" style="height: 1em;"/>&nbsp;``en``&nbsp;English | 1261 | 1172 | 0 | 0 | 0 | [messages.po](../locale/en/LC_MESSAGES/messages.po) | [Timothy ARMES](https://github.com/timothyarmes) |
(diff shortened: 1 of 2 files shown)


docs/86-i18n.md
```diff
@@ -7,6 +7,7 @@
 <!-- DO NOT EDIT! (START) -->
 | Locale | Messages | Empty | Empty mandatory | [ai_translation] | [fuzzy] | PO file | Translators |
 |--|:--:|:--:|:--:|:--:|:--:|:--:|:--:|
+<<<<<<< Updated upstream
```
Molrn: little oopsie on this file ^^

Molrn (changes requested):
Just read all the changelog, from my knowledge of what is used in papi-web we are not impacted by any deprecation. Tried running the project from a 3.12 venv, works perfectly :) big yes for me

docs/86-i18n.md
Amaras: Restored to dev

pascalaubry (approved):

Molrn (approved):

PR #259, opened 2025-03-18
Add direct links to user screens from admin

timothyarmes:


pascalaubry (changes requested):
Could you push the URL also when clicking on a menu link of a screen (in order to have the screen URL visible each time a screen is displayed)?

src/data/family.py
```diff
@@ -297,6 +297,10 @@ def screens_by_uniq_id(self) -> dict[str, Screen]:
                 screens_by_uniq_id[screen.uniq_id] = screen
         return screens_by_uniq_id
 
+    @cached_property
+    def calculated_first_screen_id(self) -> str | None:
+        return next(iter(self.screens_by_uniq_id.keys()))
```
Molrn:
```suggestion
    @cached_property
    def calculated_first_screen_id(self) -> str | None:
        return next(iter(self.screens_by_uniq_id.keys()), None)
```
Ironic that that's what we've talked about today ^^ next without default return value raises a StopIteration error when the generator is empty. You can use it without default value when the generator is never empty, but the typing and the usage of the property suggests you need a default None value

Molrn (changes requested):

pascalaubry (approved):

Molrn (approved):

PR #260, opened 2025-03-18
Remove dynamic selects (closes #258)

### LINKED ISSUE #258: Replace "dynamic selects" by normal ones
Dynamic selects have 2 issues:

1) You can't control them via the keyboard
2) They don't pop outside of the modal

timothyarmes:


pascalaubry (changes requested):
I would really prefer to keep the prefix (FIDE code) of the federations (ALG - Algeria), unless there is a good reason not to show them (I had to check them all to associate the FIDE codes and the names and I know that for some of them it is not straigh to guess the name from the FIDE code...).

pascalaubry (approved):

PR #261, opened 2025-03-18
Fixed event creation and ensure event URL unicity

pascalaubry:
image


timothyarmes (dismissed):

pascalaubry (commented):
Other URLs had to be updated, please review again.
Note: @timothyarmes other things need to be done on URLs but we have to talk about it before. In another PR.

timothyarmes (approved):

PR #264, opened 2025-03-18
Fix StopIteration error for incorrect tie-break

Molrn:


pascalaubry (approved):
Works well on all my test databases.

PR #265, opened 2025-03-18
Make fewer commits on creation

Amaras:
This PR aims to quicken database creation, or at least make it seem quicker.

This is done by making fewer transaction commits (going from once every 1000 players, to once every 100,000 players)

It might make sense to only do one commit at the end, I need some feedback on that.
The problem with that approach is that the <db>-wal file gets large relatively quickly as long as the transaction is not done.
Once the transaction is `COMMIT`-ed, a new one begins and the contents of the WAL file are overwritten from the start, limiting the size.


pascalaubry (commented):
~Tested: about 90 seconds to create the FIDE database. Much better than 180 seconds previously, but I remamber that the easiest gain was to add the indices at the very end.
Can we do both?

Amaras:
> ~Tested: about 90 seconds to create the FIDE database. Much better than 180 seconds previously, but I remamber that the easiest gain was to add the indices at the very end. Can we do both?

It wasn't the point of the PR, but I can add that in.

pascalaubry:
The maximum space occupied is ~187Mb at the very end, 151Mb one finished. 40Mb is not excessive IMO.
The gain with a single final commit is not important (a few seconds only), but the maximum occupied space is about 160Mb.

src/database/sqlite/fide/fide_database.py
```diff
@@ -203,8 +204,16 @@ def create(self) -> bool:
...
+        self.create_indexes()
         return True
 
+    def create_indexes(self) -> None:
+        self.write = True
+        with self:
+            self.execute('CREATE INDEX `player_first_name` ON `player` (`first_name` COLLATE NOCASE)')
+            self.execute('CREATE INDEX `player_last_name` ON `player` (`last_name` COLLATE NOCASE)')
+            self.execute('CREATE INDEX `player_fide_id` ON `player` (`fide_id`)')
+
```
Molrn: missing commit here I think (commited at the same place in `ffe_database.py`)

Molrn (changes requested):

src/database/sqlite/fide/fide_database.py
Amaras: done

Molrn (approved):
LGTM

PR #269, opened 2025-03-21
Only define version once

Amaras:
Using importlib.metadata.version, we can get the installed version at runtime.

This avoids version desynchronizations I have observed **several times**, especially when there are dependency version updates.

There is a workflow adaptation expected:
  1. New version is decided by one author
  2. author updates the pyproject.toml version
  3. author runs  `pip install -e .`
  4. reviewers also update their installations before reviewing
  5. reviewers can install previous versions once they go back to other branches.

> [!CAUTION]
> There is now an error on dev when the installed version and the version defined in `pyproject.toml` do not match.
> This ensures we always have the appropriate version


Molrn (approved):
Woops, I keep forgetting about the pyproject, feeling soewhat targetted by this PR ^^ anyway good idea and works well, LGTM

PR #274, opened 2025-03-21
Added FFE online search

pascalaubry:
FFE search now relies on the online FFE database, when connected to the internet:
image
When not connected, the local database is used:
image
Use script generate_ffe_sql_server_credentials.py before testing or you will get the following message:
image


src/common/network.py
```diff
@@ -0,0 +1,11 @@
+import socket
+
+
+def connected() -> bool:
+    # https://stackoverflow.com/questions/3764291/how-can-i-see-if-theres-an-available-and-active-network-connection-in-python
+    try:
+        socket.setdefaulttimeout(3)
+        socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect(('8.8.8.8', 53))
```
Amaras: You might want to try to connect to a DNS root server, instead of Google's DNS : https://www.iana.org/domains/root/servers

utils/ffe/generate_ffe_sql_server_credentials.py
```diff
@@ -0,0 +1,57 @@
...
+    FFESqlServer.dump_credentials(
+        args.host,
+        args.user,
+        args.password,
+        args.database,
+    )
+    try:
+        with FFESqlServer() as ffe_sql_server:
+            for player in ffe_sql_server.search_player('pascal aubry', limit=8):
+                print_interactive_info(f'{player=}')
```
Molrn, line 54: Even if it's for devs, deserves a little message of explanation ("credentials generated", "testing credentials", "connexion succeded/failed")

src/plugins/ffe/ffe_search_controller.py
```diff
@@ -35,14 +43,85 @@ async def htmx_search_ffe(
         template_context: dict[str, Any] = self._get_admin_event_render_context(
             web_context
         )
-        players: list['Player'] | None = None
+        search_results: list['Player'] = []
+        search_messages: list[tuple] = []
         if search_ffe:
-            with FfeDatabase() as ffe_database:
-                players: list['Player'] = [player for player in ffe_database.search_player(search_ffe, limit=8)]
+            start: float = time.perf_counter()
```
Amaras: if you're only using that in `DEVEL_ENV`, maybe don't define it for all

src/plugins/ffe/ffe_search_controller.py
```diff
@@ -35,14 +43,85 @@ async def htmx_search_ffe(
...
+            try:
+                with FFESqlServer() as ffe_sql_server:
+                    search_results: list['Player'] = [
+                        player for player in ffe_sql_server.search_player(
+                            search_ffe, limit=self.MAX_RESULTS
+                        )
+                    ]
+                    seconds: float = time.perf_counter() - start
+                    if DEVEL_ENV:
```
Amaras:
```suggestion
                    if DEVEL_ENV:
                        seconds: float = time.perf_counter() - start
```

src/plugins/ffe/ffe_search_controller.py
```diff
@@ -35,14 +43,85 @@ async def htmx_search_ffe(
...
+                    search_messages.append(
+                        (
+                            'bi-cloud-arrow-down-fill',
+                            'text-primary',
+                            message,
+                        )
+                    )
+            except PapiWebException as e:
+                search_messages.append(('bi-cloud-slash', '', str(e)))
+                start: float = time.perf_counter()
```
Amaras: same as above

utils/i18n/i18n_update.py
```diff
@@ -145,7 +145,7 @@ def compile(
     @staticmethod
     def escape_gh_md(string: str) -> str:
         """Escapes a string for GitHub markdown."""
-        return string.replace('*', '\*')
+        return string.replace('*', '\\*')
```
Amaras: This works as well:
```suggestion
        return string.replace('*', r'\*')
```

Amaras (dismissed):
Nitpicks as always, and a relatively important question

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -0,0 +1,295 @@
...
+        query: str = (
+            f'(SELECT {", ".join(self.get_player_fields() + self.get_club_fields())} FROM joueur JOIN club on joueur.ClubRef = club.Ref '
+            f'WHERE {player_and_club_condition} '
+            f'ORDER BY (CASE WHEN {order} THEN 0 ELSE 1 END), Joueur.Nom, Joueur.Prenom) '
+            f'UNION '
+            f'(SELECT {", ".join(self.get_player_fields() + self.get_empty_club_fields())} FROM joueur '
+            f'WHERE joueur.ClubRef = 0 AND {player_only_condition} '
+            f'ORDER BY (CASE WHEN {order} THEN 0 ELSE 1 END), Joueur.Nom, Joueur.Prenom) '
+        )
+        print(f'{query=}')
```
Molrn: remaining print

src/plugins/ffe/templates/ffe_search.html
```diff
@@ -6,15 +6,17 @@
...
-    {% if ffe_search_available %}
-        <ul id="ffe-search-results" class="dropdown-menu ffe-search-results p-0">
-        </ul>
-    {% endif %}
+    <ul id="ffe-search-results" class="dropdown-menu ffe-search-results p-0">
+        <li class="dropdown-item no-hover text-nowrap">
+            <span class="fst-italic">{{ _('Searching...') }}</span>
+        </li>
+    </ul>
 </span>
```
Molrn: new new line at EOF

Molrn (changes requested):
good feature, code looks good overall

src/plugins/ffe/ffe_sql_server.py
pascalaubry: It is in the unused method search_player_error(), the one who will have a look at it will probably need it, cf the point on discord :-).

src/common/network.py
pascalaubry: Well I have chosen the speedest solution (https://stackoverflow.com/questions/3764291/how-can-i-see-if-theres-an-available-and-active-network-connection-in-python) but availability is probably more important than speed here.
Pick a random server from https://www.iana.org/domains/root/servers?
Pick another one on failure to confirm?
I like this solution.

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -0,0 +1,231 @@
...
+        condition: str = ' AND '.join(map(lambda c: f'({c})', conditions))
+        order = ' OR '.join(['(UPPER(joueur.Nom) LIKE ?)', ] * len(tokens))
+        params += [f'{token}%' for token in tokens]
+        query: str = (
+            f'SELECT {", ".join(self.get_player_fields() + self.get_club_fields())} '
+            f'FROM joueur LEFT JOIN club on joueur.ClubRef = club.Ref '
+            f'WHERE {condition} '
+            f'ORDER BY (CASE WHEN {order} THEN 0 ELSE 1 END), Joueur.Nom, Joueur.Prenom'
+        )
+        print(f'{query=}')
```
Molrn: now I think it should be removed ^^

Molrn (changes requested):
one final change and we're good :)

Molrn (approved):

PR #275, opened 2025-03-21
Do not check the version and update databases when not connected

pascalaubry:


Amaras (commented):
Without running the code, looks good

Amaras (approved):

PR #284, opened 2025-03-23
Make SQLServer search async

Amaras:
Made SQLServer search async

In addition:
* simplified the player search query, removing searching by club
* if the search string looks like a FFE licence number, take a shortcut
* added a HTMX sync directive for queries to cancel in-flight queries on
  search changes

With the simplified query, this yields a 4-5x improvement in query
speed on simple tests.
The shortcut yields a 1.5-2x speed on simple tests as well.

Also add various simple fixes


Amaras:
Maybe the `raise from e` should be `raise from None` in non-dev user-facing code, I'll think about it before merging

timothyarmes (approved):

PR #298, opened 2025-03-25
Download databases in the background (closes #297)

### LINKED ISSUE #297: Use separate threads to update the offline databases
Currently the application is blocked during the update of the off-line databases. Since the updates are time consuming this is frustrating.

I'd like to have each update running in its own thread so that the application can start up immediately.

During this time we need to handle access to the old database versions for player search and player updates.  There are two options:

1. Disable access to these features until the update is ready
2. Use a thread-locking mechanism so that:
   * The updater blocks when overwriting the old database if it has an open connection
   * The app blocks when opening a connection to database at the instant that the updater is overwriting it.

Looking to the future option 2 along with an automatic download would be nice.  The simplest way of using the app would be just just launch the exe as an application without needing to know anything about the terminal, so if we can avoid asking the users questions via the terminal it'd be a step in that direction.

timothyarmes:


src/plugins/ffe/ffe.py
```diff
@@ -106,7 +106,7 @@ def migration_manager(self) -> PluginMigrationManager:
 
     @hookimpl
     def on_init(self):
-        if not FfeDatabase().check():
+        if FfeDatabase().check() == False:
```
Amaras: Don't check boolean equality with `== False`, check with their truthy value
```suggestion
        if not FfeDatabase().check():
```

src/web/server_engine.py
```diff
@@ -62,7 +62,7 @@ def __init__(self, debug: bool=False):
             )
         )
 
-        if not FideDatabase().check():
+        if FideDatabase().check() == False:
```
Amaras:
```suggestion
        if not FideDatabase().check():
```

Amaras (dismissed):
Code looks pretty good, just don't check boolean equality with `==`

Molrn:
A little abstraction between the classes FideDatabase and FfeDatabase woud be nice (like a class "AbstractPlayerDatabase"), but if you don't feel like doing it now I can do it later

timothyarmes:
> A little abstraction between the classes FideDatabase and FfeDatabase woud be nice (like a class "AbstractLocalDatabase"), but if you don't feel like doing it now I can do it later

Abstraction is your forté, I'll let you do that later :)

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -4101,8 +4104,11 @@ msgstr "oui"
 msgid "{family_type}-family"
 msgstr "famille-{family_type}"
 
-msgid "{number} players written."
-msgstr "{number} joueur·euses enregistré·es."
+msgid "{number} players written to FFE database."
+msgstr ""
+
+msgid "{number} players written to FIDE database."
+msgstr ""
```
```suggestion
msgid "{number} players written to FFE database."
msgstr "{number} joueur·euses enregistré·es dans la base de données FFE."

msgid "{number} players written to FIDE database."
msgstr "{number} joueur·euses enregistré·es dans la base de données FIDE."
```

src/database/sqlite/fide/fide_database.py
```diff
@@ -50,15 +52,16 @@ class FideDatabase(SQLiteDatabase):
 
     def __init__(self, write: bool = False):
         super().__init__(TMP_DIR / f'fide.{PapiWebConfig.federation_database_ext}', write)
-
-    def check(self) -> bool:
+        self.stop_event = Event()
+        
+    def check(self):
         """Check if the database exists and proposes to create it if not, or update it if too old,
         returns True if the database is available after the call, False otherwise."""
```
Molrn:
```suggestion
        """Checks if the database exists and is up to date and proposes to create it if not"""
```

src/plugins/ffe/ffe_database.py
```diff
@@ -55,15 +57,16 @@ class FfeDatabase(SQLiteDatabase):
 
     def __init__(self, write: bool = False):
         super().__init__(TMP_DIR / f'ffe.{PapiWebConfig.federation_database_ext}', write)
+        self.stop_event = Event()
 
-    def check(self) -> bool:
+    def check(self):
         """Check if the database exists and proposes to create it if not, or update it if too old,
         returns True if the database is available after the call, False otherwise."""
```
Molrn:
```suggestion
        """Checks if the database exists and is up to date and proposes to create it if not"""
```

Molrn (changes requested):

Molrn (approved):

PR #305, opened 2025-03-26
Added the plugins to the EXE

pascalaubry:


src/web/settings.py
```diff
@@ -162,7 +161,7 @@ def join_path(self, template: str, parent: str) -> str:
 
 template_dirs: list[Path] = [
     BASE_DIR / 'src/web/templates',
-    *[path for path in plugin_manager.template_paths],
+    *[path for path in plugin_manager.templates_paths if path],
```
Molrn:
```suggestion
    *[path for path in plugin_manager.templates_paths],
```
path never is not set

src/common/papi_web_config.py
```diff
@@ -65,7 +62,35 @@ def __init__(self):
...
+        # offline databases download, ...
+        if not self.stored_config.locale:
+            set_locale(get_locale())
+            print_interactive_input(_('The following languages are available:'))
+            locale_range = range(1, len(self.locales) + 1)
+            for num in locale_range:
+                locale: str = self.locales[num - 1]
+                print_interactive_input(
+                    f'  - [{num}] {locale} ({locale_localized_name(locale)})'
+                )
```
Molrn, line 79: I thought we were trying to get rid of interactive inputs to more move towards a no console application ?

src/papi_web.py
```diff
@@ -1,34 +1,35 @@
...
 
 # undocumented feature to start from a different folder and work with different configurations
 # Has to be executed before plugin_manager to avoid initializing from the wrong path
 path_parser = argparse.ArgumentParser(add_help=False)
 path_parser.add_argument('--path', default='.')
+path_parser.add_argument('--server', action='store_true')
 args, remaining_args = path_parser.parse_known_args()
 os.chdir(args.path)
+if args.server:
+    print_interactive_warning(_('Argument --server is deprecated, ignored.'))
```
Molrn: should not be checked by the path_parser, its usage before imports is exceptional and documented, unrelated code should not be added to it

Molrn (dismissed):
missing translation for "The following languages are available:". export runs smoothly, no appearing problems in the exported app

Molrn (approved):
LGTM

PR #307, opened 2025-03-26
Gather abstract base classes as IdentifiableEntity

Molrn:
Tools for Abstract entities both present in the DB / forms and in the UI. Application for existing such classes. usages were kept as much as possible as they were, and impacted features were thouroughly retested:
- Player update from FFE and FIDE database
- document printing
- tie-breaks


timothyarmes (approved):
Approving the code.  Class renames and comment would be nice in a future PR.

PR #310, opened 2025-03-28
Fix types

timothyarmes:


src/common/papi_web_config.py
```diff
@@ -285,14 +285,14 @@ def local_url(self) -> str:
     default_record_illegal_moves_number: int = 0
 
     # The default colors for the timers.
-    default_timer_colors: dict[int, str] = {
+    default_timer_colors: dict[int, str | None] = {
```
Amaras: Defaults are always defined, overrides are not.
The typing problem is not here, it's in the calling code.
```suggestion
    default_timer_colors: dict[int, str] = {
```

src/common/papi_web_config.py
```diff
@@ -285,14 +285,14 @@ def local_url(self) -> str:
...
         1: '#00FF00',
         2: '#FF7700',
         3: '#FF0000',
     }
 
     # The default delays for the timers.
-    default_timer_delays: dict[int, int] = {
+    default_timer_delays: dict[int, int | None] = {
```
Amaras:
```suggestion
    default_timer_delays: dict[int, int] = {
```

src/data/board.py
```diff
@@ -54,6 +54,7 @@ def __lt__(self, other):
         # the highest-scoring players
         self_player_1: Player
         self_player_2: Player
+        assert self.white_player is not None, 'The white player is not defined.'
```
Amaras: This should raise a `ValueError` if `self.white_player` is not None, as boards have at least a white player

src/data/board.py
```diff
@@ -63,13 +64,18 @@ def __lt__(self, other):
         # Here self_player_1 is the strongest player of this board
         other_player_1: Player
         other_player_2: Player
+        assert other.white_player is not None, 'The white player is not defined.'
```
Amaras: Same as above, comparing boards where the white player is not defined should be a `ValueError`, as all proper boards have at least a white player

src/data/board.py
```diff
@@ -63,13 +64,18 @@ def __lt__(self, other):
...
             other_player_1 = other.black_player
             other_player_2 = other.white_player
         else:
             other_player_1 = other.white_player
             other_player_2 = other.black_player
         # Here other_player_1 is the strongest player of the other board
+        assert self_player_1.vpoints is not None, 'Self Player 1 has no vpoints.'
+        assert other_player_1.vpoints is not None, 'Other Player 1 has no vpoints.'
+        assert self_player_2.vpoints is not None, 'Self Player 2 has no vpoints.'
+        assert other_player_2.vpoints is not None, 'Other Player 2 has no vpoints.'
```
Amaras, line 81:
```suggestion
        if TYPE_CHECKING:
            assert self_player_1.vpoints is not None
            assert other_player_1.vpoints is not None
            assert self_player_2.vpoints is not None
            assert other_player_2.vpoints is not None
```

src/data/board.py
```diff
@@ -97,6 +103,7 @@ def __eq__(self, other):
             return False
         self_player_1: Player
         self_player_2: Player
+        assert self.white_player is not None, 'The white player is not defined.'
```
Amaras: Same remark as with `__lt__`

src/data/board.py
```diff
@@ -105,6 +112,7 @@ def __eq__(self, other):
             self_player_2 = self.black_player
         other_player_1: Player
         other_player_2: Player
+        assert other.white_player is not None, 'The white player is not defined.'
```
Amaras: same remark as with `__lt__`

src/data/event.py
```diff
@@ -76,10 +76,11 @@ def formatted_text(self) -> str:
                 text=self.text,
             )
         elif self.timer:
+            assert self.timer_hour is not None
             return _('Timer [{timer_uniq_id}]: {text}').format(
                 timer_uniq_id=self.timer_hour.timer.uniq_id, text=self.text
             )
```
Amaras: Since `self.timer_hour` is considered false, this means that it is `None` (an object is considered True unless its class implements a `__bool__()` method that returns `False` or a `__len__()` method that returns zero, when called with the object, see https://docs.python.org/3.12/library/stdtypes.html#truth-value-testing)
```suggestion
            return _('Timer [{timer_uniq_id}]: {text}').format(
                timer_uniq_id=self.timer.uniq_id, text=self.text
            )
```

src/data/event.py
```diff
@@ -341,19 +342,19 @@ def rules(self) -> str | None:
...
-    def timer_colors(self) -> dict[int, str]:
+    def timer_colors(self) -> dict[int, str | None]:
         return {
-            i: self.stored_event.timer_colors[i]
-            if i in self.stored_event.timer_colors and self.stored_event.timer_colors[i]
+            i: str(self.stored_event.timer_colors[i])
+            if self.stored_event.timer_colors and i in self.stored_event.timer_colors and self.stored_event.timer_colors[i]
             else PapiWebConfig.default_timer_colors[i]
             for i in range(1, 4)
         }
```
Amaras: this can probably be simplified for the type checker (and our brains), like:
```py
    def timer_colors(self) -> dict[int, str]:
        colors = PapiWebConfig.default_timer_colors
        stored_colors = self.stored_event.timer_colors
        if stored_colors is not None:
            for i in range(1, 4):
                if i in stored_colors and (color := stored_colors[i]) is not None:
                    colors[i] = color
        return colors
```

src/data/event.py
```diff
@@ -341,19 +342,19 @@ def rules(self) -> str | None:
...
-    def timer_delays(self) -> dict[int, int]:
+    def timer_delays(self) -> dict[int, int | None]:
         return {
-            i: self.stored_event.timer_delays[i]
-            if i in self.stored_event.timer_delays and self.stored_event.timer_delays[i]
+            i: int(self.stored_event.timer_delays[i] or 0)
+            if self.stored_event.timer_delays and i in self.stored_event.timer_delays and self.stored_event.timer_delays[i] is not None
             else PapiWebConfig.default_timer_delays[i]
             for i in range(1, 4)
         }
```
Amaras: Same as above:
```py
    def timer_delays(self) -> dict[int, int]:
        delays = PapiWebConfig.default_timer_delays
        stored_delays = self.stored_event.timer_delays
        if stored_delays is not None:
            for i in range(1, 4):
                if i in stored_delays and (delay := stored_delays) is not None:
                    delays[i] = delay 
        return delays
```

src/data/family.py
```diff
@@ -215,7 +218,7 @@ def last_update(self) -> float | None:
     def last_update_str(self) -> str | None:
         return format_timestamp_date_time(self.last_update)
 
-    @cache
+    @cached_property
```
Amaras: This is not a property and thus should not be a `cached_property`.
It should either be a cached `staticmethod` that is passed all necessary parameters and returns all necessary things, or the resulting properties should be cached.

This is just confusing, because of the (cached) side effects

src/data/loader.py
```diff
@@ -58,7 +58,8 @@ def clear_cache(self, event_uniq_id: str | None = None):
 
     def load_stored_event(self, uniq_id: str) -> StoredEvent:
         try:
-            return self._loaded_stored_events_by_id[uniq_id]
+            event = self._loaded_stored_events_by_id[uniq_id] 
+            return event
```
Amaras: what wrong with the previous version?

src/data/player.py
```diff
@@ -152,14 +152,15 @@ def points_after(
             pairing.result.points(self.point_values)
             for round_index, pairing in self.pairings.items()
             if round_index <= after_round and
+            pairing.result and
```
Amaras:
```suggestion
            pairing.result is not None and
```

src/data/player.py
```diff
@@ -152,14 +152,15 @@ def points_after(
...
             (pairing.played or not only_played)
         )
 
     def total_points(self, only_played: bool = False) -> float:
         return sum(
             pairing.result.points(self.point_values)
             for pairing in self.pairings.values()
-            if pairing.played or not only_played
+            if (pairing.played or not only_played) and pairing.result
```
Amaras:
```suggestion
            if (pairing.played or not only_played) and pairing.result is not None
```

src/data/player.py
```diff
@@ -252,27 +253,32 @@ def player_tournament_id_from_papi_web_id(player_id: int) -> int:
     @property
     def ref_id(self) -> int:
         """Returns the Unique ID of the player in the Papi file (needed while using the Papi storage)."""
+        assert self.id is not None
```
Amaras, line 256: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -252,27 +253,32 @@ def player_tournament_id_from_papi_web_id(player_id: int) -> int:
...
         return self.player_papi_id_from_papi_web_id(self.id)
 
     @property
     def tournament_id(self) -> int:
+        assert self.id is not None
```
Amaras, line 261: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -252,27 +253,32 @@ def player_tournament_id_from_papi_web_id(player_id: int) -> int:
...
         return self.player_tournament_id_from_papi_web_id(self.id)
 
     @property
     def estimation(self) -> int:
         if not self.estimated:
-            return self.ratings[self.tournament.rating]
+            assert self.tournament is not None
```
Amaras, line 267: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -252,27 +253,32 @@ def player_tournament_id_from_papi_web_id(player_id: int) -> int:
...
+            return self.ratings[self.tournament.rating] or 0
         return self._estimation or 0
 
     @estimation.setter
     def estimation(self, value: int):
         if not self.estimated:
-            self._estimation = self.ratings[self.tournament.rating]
+            assert self.tournament is not None
```
Amaras, line 274: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -252,27 +253,32 @@ def player_tournament_id_from_papi_web_id(player_id: int) -> int:
...
+            self._estimation = self.ratings[self.tournament.rating] or 0
         else:
             self._estimation = value
 
     @property
     def estimated(self) -> bool:
+        assert self.tournament is not None
```
Amaras, line 281: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -285,10 +291,12 @@ def category(self) -> PlayerCategory:
 
     @property
     def rating(self) -> int:
-        return self.ratings[self.tournament.rating]
+        assert self.tournament is not None
```
Amaras, line 294: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -285,10 +291,12 @@ def category(self) -> PlayerCategory:
...
+        return self.ratings[self.tournament.rating] or 0
 
     @property
     def rating_type(self) -> PlayerRatingType:
+        assert self.tournament is not None
```
Amaras, line 299: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -304,7 +312,11 @@ def compute_points(
         self.points = self.points_before(before_round)
 
     def points_total(self) -> float:
-        return sum(pairing.result.points(self.point_values) for pairing in self.pairings.values())
+        return sum(
+            pairing.result.points(self.point_values)
+            for pairing in self.pairings.values()
+            if pairing.result
```
Amaras:
```suggestion
            if pairing.result is not None
```

src/data/player.py
```diff
@@ -327,6 +339,7 @@ def to_trf(
         *,
         after_round: int,
     ) -> TrfPlayer:
+        assert self.id is not None
```
Amaras, line 342: add `if TYPE_CHECKING`

src/data/player.py
```diff
@@ -396,6 +409,7 @@ def has_real_pairings(self) -> bool:
 
     @cached_property
     def can_check_in_out(self) -> bool:
+        assert self.tournament is not None
```
Amaras, line 412: add `if TYPE_CHECKING`

src/data/print.py
```diff
@@ -341,6 +341,7 @@ def validate_options(self):
         ranking_round = self._get_option(RoundPrintOption).value
         if ranking_round is None:
             return
+        assert self.tournament is not None
```
Amaras, line 347: add `if TYPE_CHECKING`

src/data/print.py
```diff
@@ -408,6 +409,7 @@ def show_results(self) -> bool:
 
     @property
     def boards(self) -> list[Board]:
+        assert self.tournament is not None
```
Amaras, line 415: add `if TYPE_CHECKING`

src/data/print.py
```diff
@@ -423,6 +425,7 @@ def template_context(self) -> dict[str, Any]:
 
     @property
     def at_round(self) -> int:
+        assert self.tournament is not None
```
Amaras, line 431: add `if TYPE_CHECKING`

src/data/rotator.py
```diff
@@ -33,10 +33,13 @@ def __init__(
 
     @property
     def event(self) -> 'Event':
-        return self._event_ref()
+        event = self._event_ref()
+        assert event is not None, "Event reference has been garbage collected"
```
Amaras: do the same as previously: throw a `RuntimeErorr`

src/data/screen.py
```diff
@@ -81,34 +80,35 @@ def screen_sets_by_id(self) -> dict[int | None, ScreenSet]:
...
     @property
     def type(self) -> ScreenType:
-        return (
-            ScreenType(self.stored_screen.type)
-            if self.stored_screen
-            else self.family.type
-        )
+        if self.stored_screen:
+            return ScreenType(self.stored_screen.type)
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` if `self.family is None`

src/data/screen.py
```diff
@@ -81,34 +80,35 @@ def screen_sets_by_id(self) -> dict[int | None, ScreenSet]:
...
+        return self.family.type
 
     @property
     def public(self) -> bool:
-        return self.stored_screen.public if self.stored_screen else self.family.public
+        if self.stored_screen:
+            return self.stored_screen.public
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeErorr` instead

src/data/screen.py
```diff
@@ -81,34 +80,35 @@ def screen_sets_by_id(self) -> dict[int | None, ScreenSet]:
...
     @property
     def uniq_id(self) -> str:
-        return (
-            self.stored_screen.uniq_id
-            if self.stored_screen
-            else f'{self.family.uniq_id}:{self.family_part:03}'
-        )
+        if self.stored_screen:
+            return self.stored_screen.uniq_id
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -81,34 +80,35 @@ def screen_sets_by_id(self) -> dict[int | None, ScreenSet]:
...
+        return f'{self.family.uniq_id}:{self.family_part:03}'
 
     @property
-    def name(self) -> str:
+    def name(self) -> str | None:
```
Amaras, line 117: you can do better, unless callers specifically check for `None`

src/data/screen.py
```diff
@@ -134,23 +134,24 @@ def columns(self) -> int:
             else:
                 return 1
         else:
+            assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -217,6 +218,7 @@ def menu_label(self) -> str | None:
             return None
         match self.type:
             case ScreenType.BOARDS | ScreenType.INPUT | ScreenType.PLAYERS | ScreenType.RANKING:
+                assert self.event is not None
```
Amaras, line 221: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -265,18 +267,20 @@ def menu_label(self) -> str | None:
                         text = text.replace('%l', str(screen_set.last_board.id))
                 return text
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
```
Amaras, line 270: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -265,18 +267,20 @@ def menu_label(self) -> str | None:
...
                 return self.stored_screen.menu_text or _('Last results')
             case _:
                 raise ValueError(f'type=[{self.type}]')
 
-    def _menu_screens(self, admin: bool) -> list[Self]:
-        menu_screens: list[Self] = []
+    def _menu_screens(self, admin: bool) -> list['Screen']:
+        menu_screens: list['Screen'] = []
+        assert self.event is not None
```
Amaras, line 277: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -389,21 +393,27 @@ def _menu_screens(self, admin: bool) -> list[Self]:
...
-    def admin_menu_screens(self) -> list[Self]:
+    def admin_menu_screens(self) -> list['Screen']:
         return self._menu_screens(True)
 
     @property
     def menu(self) -> str:
-        return self.stored_screen.menu if self.stored_screen else self.family.menu
+        if self.stored_screen:
+            return self.stored_screen.menu or ''
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -389,21 +393,27 @@ def _menu_screens(self, admin: bool) -> list[Self]:
...
+        return self.family.menu or ''
 
     @property
     def timer(self) -> Timer | None:
+        assert self.event is not None
```
Amaras, line 412: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -389,21 +393,27 @@ def _menu_screens(self, admin: bool) -> list[Self]:
...
         timer_id: int | None = (
-            self.stored_screen.timer_id if self.stored_screen else self.family.timer_id
+            self.stored_screen.timer_id if self.stored_screen
+            else self.family.timer_id if self.family
+            else None
```
Amaras: chained ternary checks are confusing, please break out into `if... elif... else` checks

src/data/screen.py
```diff
@@ -430,6 +440,7 @@ def input_exit_button(self) -> bool:
                     else:
                         return PapiWebConfig.default_input_exit_button
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -447,6 +458,7 @@ def players_show_unpaired(self) -> bool:
                     else:
                         return PapiWebConfig.default_players_show_unpaired
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -479,6 +491,8 @@ def type_str(self) -> str:
     def results_limit(self) -> int:
         match self.type:
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
+                assert self.event is not None
```
Amaras, line 495: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -504,6 +518,7 @@ def results_limit(self) -> int:
     def results_max_age(self) -> int:
         match self.type:
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
```
Amaras, line 521: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -515,6 +530,8 @@ def results_max_age(self) -> int:
     def results_tournament_ids(self) -> list[int]:
         match self.type:
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
+                assert self.event is not None
```
Amaras, line 534: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -525,6 +542,7 @@ def results_tournament_ids(self) -> list[int]:
 
     @cached_property
     def results_tournament_names(self) -> str:
+        assert self.event is not None
```
Amaras, line 545: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -536,6 +554,7 @@ def results_tournament_names(self) -> str:
 
     @cached_property
     def _results(self) -> list[Result]:
+        assert self.event is not None
```
Amaras, line 557: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -554,8 +573,9 @@ def ranking_crosstable(self) -> bool:
         match self.type:
             case ScreenType.RANKING:
                 if self.stored_screen:
-                    return self.stored_screen.ranking_crosstable == True
+                    return bool(self.stored_screen.ranking_crosstable)
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -554,8 +573,9 @@ def ranking_crosstable(self) -> bool:
         match self.type:
             case ScreenType.RANKING:
                 if self.stored_screen:
-                    return self.stored_screen.ranking_crosstable == True
+                    return bool(self.stored_screen.ranking_crosstable)
```
Amaras, line 579: isn't the property a `bool` already?

src/data/screen.py
```diff
@@ -567,45 +587,48 @@ def ranking_round(self) -> int | None:
                 if self.stored_screen:
                     return self.stored_screen.ranking_round
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -567,45 +587,48 @@ def ranking_round(self) -> int | None:
...
 
     @property
-    def ranking_min_points(self) -> int | None:
+    def ranking_min_points(self) -> float | None:
         match self.type:
             case ScreenType.RANKING:
                 if self.stored_screen:
                     return self.stored_screen.ranking_min_points
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -567,45 +587,48 @@ def ranking_round(self) -> int | None:
...
 
     @property
-    def ranking_max_points(self) -> int | None:
+    def ranking_max_points(self) -> float | None:
         match self.type:
             case ScreenType.RANKING:
                 if self.stored_screen:
                     return self.stored_screen.ranking_max_points
                 else:
+                    assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -567,45 +587,48 @@ def ranking_round(self) -> int | None:
...
     @property
     def last_update(self) -> float:
-        return (
-            self.stored_screen.last_update
-            if self.stored_screen
-            else self.family.last_update
-        )
+        if self.stored_screen:
+            return self.stored_screen.last_update or 0.0
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

src/data/screen.py
```diff
@@ -567,45 +587,48 @@ def ranking_round(self) -> int | None:
...
+        return self.family.last_update or 0.0
 
     @property
     def background_image(self) -> str:
         if self.stored_screen and self.stored_screen.background_image:
             return self.stored_screen.background_image
         else:
+            assert self.event is not None
```
Amaras, line 631: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -617,26 +640,25 @@ def background_color(self) -> str:
         if self.stored_screen and self.stored_screen.background_color:
             return self.stored_screen.background_color
         else:
+            assert self.event is not None
```
Amaras, line 643: add `if TYPE_CHECKING`

src/data/screen.py
```diff
@@ -617,26 +640,25 @@ def background_color(self) -> str:
...
     @property
     def message_default(self) -> bool:
-        return (
-            self.stored_screen.message_default
-            if self.stored_screen
-            else self.family.message_default
-        )
+        if self.stored_screen:
+            return self.stored_screen.message_default
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw `RuntimeError` instead

src/data/screen.py
```diff
@@ -617,26 +640,25 @@ def background_color(self) -> str:
...
             return self.event.message_text
-        else:
-            return (
-                self.stored_screen.message_text
-                if self.stored_screen
-                else self.family.message_text
-            )
+        if self.stored_screen:
+            return self.stored_screen.message_text
+        assert self.family is not None, "Family reference has been garbage collected"
```
Amaras: throw a `RuntimeError` instead

Amaras (dismissed):
stopping after 15 files for now, there are way too many changes to do already

src/data/event.py
```diff
@@ -343,17 +343,17 @@ def rules(self) -> str | None:
...
     @cached_property
     def timer_delays(self) -> dict[int, int]:
         return {
-            i: self.stored_event.timer_delays[i]
-            if i in self.stored_event.timer_delays and self.stored_event.timer_delays[i]
+            i: int(self.stored_event.timer_delays[i] or 0)
+            if self.stored_event.timer_delays and i in self.stored_event.timer_delays and self.stored_event.timer_delays[i] is not None
             else PapiWebConfig.default_timer_delays[i]
             for i in range(1, 4)
         }
```
Amaras: The comprehensions are too hard to follow for me, please replace them by an explicit iteration

src/data/event.py
```diff
@@ -604,7 +607,7 @@ def get_unused_screen_uniq_id(
         """Returns the first unused screen uniq_id looking like base_uniq_id:
         base_uniq_id, or base_uniq_id-2, or base_uniq_id-n+1...
         screen_type is used when the given ID is empty to set an ID that corresponds to the screen type."""
-        assert base_uniq_id is not None or screen_type is not None
+        assert screen_type is not None
```
Amaras: If you assert this, then `screen_type` should not be `None`-able

src/data/event.py
```diff
@@ -650,6 +654,7 @@ def get_unused_family_uniq_id(
         """Returns the first unused family uniq_id looking like base_uniq_id:
         base_uniq_id, or base_uniq_id-2, or base_uniq_id-n+1...
         family_type is used when the given ID is empty to set an ID that corresponds to the family type."""
+        assert family_type is not None
```
Amaras: `family_type` should not be `None-able`

src/data/family.py
```diff
@@ -280,9 +286,9 @@ def _calculate_screens(self) -> bool:
                             player
                             for player in self.tournament.players_by_rank.values()
                             if (
-                                       self.ranking_min_points is None or player.points >= self.ranking_min_points
+                                    self.ranking_min_points is None or (player.points or 0) >= float(self.ranking_min_points)
```
Amaras: why is the `float` cast necessary?

src/data/family.py
```diff
@@ -280,9 +286,9 @@ def _calculate_screens(self) -> bool:
...
                                ) and (
-                                       self.ranking_max_points is None or player.points <= self.ranking_max_points
+                                    self.ranking_max_points is None or (player.points or 0) <= float(self.ranking_max_points)
```
Amaras: why is the `float`cast necessary?

src/data/family.py
```diff
@@ -338,8 +344,8 @@ def _calculate_screens(self) -> bool:
     @cached_property
     def screens_by_uniq_id(self) -> dict[str, Screen]:
         screens_by_uniq_id: dict[str, Screen] = {}
-        if self._calculate_screens():
-            for family_index in range(1, self.calculated_parts + 1):
+        if self._calculate_and_cache_screens():
+            for family_index in range(1, (self.calculated_parts or 1) + 1):
```
Amaras: maybe `self._calculated_parts` should default at 1.
Only `self._calculate_screens()` sets it, and it is always set to an `int`, never to `None`

src/data/screen.py
```diff
@@ -134,23 +137,25 @@ def columns(self) -> int:
             else:
                 return 1
```
Amaras: This inner `if` should probably be a simple `return self.stored_screen.columns or 1`

src/data/screen.py
```diff
@@ -217,6 +222,7 @@ def menu_label(self) -> str | None:
             return None
         match self.type:
             case ScreenType.BOARDS | ScreenType.INPUT | ScreenType.PLAYERS | ScreenType.RANKING:
+                assert self.event is not None
```
Amaras: `self.event` should never be `None`, logically.
Either it is set, or it's not set and we have a floating screen.
The property should reflect that by throwing a `RuntimeError` if the object has been unexpectedly garbage collected.

src/data/screen.py
```diff
@@ -265,18 +271,20 @@ def menu_label(self) -> str | None:
...
+                assert self.stored_screen is not None
                 return self.stored_screen.menu_text or _('Last results')
             case _:
                 raise ValueError(f'type=[{self.type}]')
 
-    def _menu_screens(self, admin: bool) -> list[Self]:
-        menu_screens: list[Self] = []
+    def _menu_screens(self, admin: bool) -> list['Screen']:
+        menu_screens: list['Screen'] = []
+        assert self.event is not None
```
Amaras: not relevant with recommendation above

src/data/screen.py
```diff
@@ -389,21 +397,28 @@ def _menu_screens(self, admin: bool) -> list[Self]:
...
-        return self.stored_screen.menu if self.stored_screen else self.family.menu
+        if self.stored_screen:
+            return self.stored_screen.menu or ''
+        if self.family is None:
+            raise RuntimeError("Family reference unexpectedly None")
+        return self.family.menu or ''
 
     @property
     def timer(self) -> Timer | None:
+        assert self.event is not None
```
Amaras: not relevant if the property throws

src/data/screen.py
```diff
@@ -479,6 +498,8 @@ def type_str(self) -> str:
     def results_limit(self) -> int:
         match self.type:
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
+                assert self.event is not None
```
Amaras: not relevant if the property throws

src/data/screen.py
```diff
@@ -515,6 +537,8 @@ def results_max_age(self) -> int:
     def results_tournament_ids(self) -> list[int]:
         match self.type:
             case ScreenType.RESULTS:
+                assert self.stored_screen is not None
+                assert self.event is not None
```
Amaras: not relevant once the property throws

src/data/screen.py
```diff
@@ -525,6 +549,7 @@ def results_tournament_ids(self) -> list[int]:
 
     @cached_property
     def results_tournament_names(self) -> str:
+        assert self.event is not None
```
Amaras: not relevant

src/data/screen.py
```diff
@@ -536,6 +561,7 @@ def results_tournament_names(self) -> str:
 
     @cached_property
     def _results(self) -> list[Result]:
+        assert self.event is not None
```
Amaras: not relevant

src/data/screen.py
```diff
@@ -567,45 +595,52 @@ def ranking_round(self) -> int | None:
...
+        if self.family is None:
+            raise RuntimeError("Family reference unexpectedly None")
+        return self.family.last_update or 0.0
 
     @property
     def background_image(self) -> str:
         if self.stored_screen and self.stored_screen.background_image:
             return self.stored_screen.background_image
         else:
+            assert self.event is not None
```
Amaras: not relevant

src/data/screen.py
```diff
@@ -617,26 +652,27 @@ def background_color(self) -> str:
         if self.stored_screen and self.stored_screen.background_color:
             return self.stored_screen.background_color
         else:
+            assert self.event is not None
```
Amaras: not relevant

src/data/screen.py
```diff
@@ -617,26 +652,27 @@ def background_color(self) -> str:
...
-        )
+        if self.stored_screen:
+            return self.stored_screen.message_default
+        if self.family is None:
+            raise RuntimeError("Family reference unexpectedly None")
+        return self.family.message_default
 
     @property
     def message_text(self) -> str | None:
+        assert self.event is not None
```
Amaras: not relevant

src/data/screen_set.py
```diff
@@ -132,6 +132,7 @@ def order(self) -> int | None:
 
     @property
     def event(self) -> 'Event':
+        assert self.screen.event is not None
```
Amaras: not relevant once `Screen.event` raises a `RuntimeError` when the `Event` object is garbage collected

src/data/tie_break.py
```diff
@@ -124,8 +124,9 @@ def entity_types() -> list[type[AbstractTieBreak]]:
     @classmethod
     def type_by_papi_id(cls) -> dict[str, type[AbstractTieBreak]]:
         return {
-            entity_type.static_papi_id(): entity_type
+            str(entity_type.static_papi_id()): entity_type
```
Amaras, line 127: technically not needed, but I understand why the type checker might be confused

src/data/tie_break.py
```diff
@@ -326,12 +328,13 @@ def compute_player_value(
...
         if after_round is None:
             after_round = max(player.pairings)
+        assert player.tournament is not None
         point_values = player.tournament.point_values
         return sum(
             pairing.result.points(point_values) == Result.GAIN.points(
                 point_values
             ) for round_index, pairing in player.pairings.items()
-            if round_index <= after_round
+            if pairing.result and round_index <= after_round
```
Amaras:
```suggestion
            if pairing.result is not None and round_index <= after_round
```

src/data/timer.py
```diff
@@ -42,7 +42,10 @@ def __init__(
 
     @property
     def timer(self) -> 'Timer':
-        return self._timer_ref()
+        timer = self._timer_ref()
+        if timer is None:
+            raise RuntimeError("Timer reference has been garbage collected")
+        return timer
```
Amaras: I don't want to have any trouble with shadowing names
```suggestion
        inner_timer = self._timer_ref()
        if inner_timer is None:
            raise RuntimeError("Timer reference has been garbage collected")
        return inner_timer
```

src/data/timer.py
```diff
@@ -105,19 +108,19 @@ def text_after(self) -> str:
 
     @property
     def timestamp_1(self) -> int:
-        return self.timestamp - self.timer.delays[1] * 60 - self.timer.delays[2] * 60
+        return (self.timestamp or 0) - self.timer.delays[1] * 60 - self.timer.delays[2] * 60
```
Amaras: I'm not convinced I want negative timestamps

src/data/timer.py
```diff
@@ -105,19 +108,19 @@ def text_after(self) -> str:
...
 
     @property
     def timestamp_2(self) -> int:
-        return self.timestamp - self.timer.delays[2] * 60
+        return (self.timestamp or 0) - self.timer.delays[2] * 60
```
Amaras: Are you sure you want negative timestamps?

src/database/sqlite/event/event_database.py
```diff
@@ -2197,6 +2198,8 @@ def clone_stored_screen_set(
         screen_id: int,
     ) -> StoredScreenSet:
         stored_screen_set = self.get_stored_screen_set(screen_set_id)
+        assert stored_screen_set is not None
+        assert stored_screen_set.id is not None
         stored_screen_set.id = None
```
Amaras: Why do you need to assert that `stored_screen_set.id is not None` when it set to `None` immediately after?

src/plugins/chessevent/chessevent.py
```diff
@@ -4,6 +4,7 @@
 from packaging.version import Version
 
 from common.i18n import _
+from data.event import Event
```
Amaras: put it in the `if TYPE_CHECKING` bloc if you don't specifically need it

src/plugins/ffe/engine/ffe_session.py
```diff
@@ -239,12 +239,13 @@ def _ffe_auth(self) -> bool:
...
             'ctl00$CmdLogin.y': '6',
         }
-        html: str = self._read_url(url=url, data=post_data, files=None)
+        html = self._read_url(url=url, data=post_data, files=None)
         if not html:
             return False
         parser, error = self._parse_html_content(html)
         if error:
             return False
+        assert parser is not None
```
Amaras: alright, this is a post-condition check, it has its place in the code

src/plugins/ffe/ffe_access_database.py
```diff
@@ -39,14 +38,14 @@ def read_player_dicts(self) -> Iterator[dict[str, str | int | datetime | None]]:
             'ffe_licence': 'joueur.AffType',  # str
         }
         # the fields for players with a club
-        club_fields: dict[str, tuple[str, types.FunctionType]] = common_fields | {
+        club_fields: dict[str, str | int | datetime | None] = common_fields | {
```
Amaras: looks like it should be:
```suggestion
        club_fields: dict[str, str] = common_fields | {
```

src/plugins/ffe/ffe_access_database.py
```diff
@@ -39,14 +38,14 @@ def read_player_dicts(self) -> Iterator[dict[str, str | int | datetime | None]]:
...
             # '': 'club.NrFFE',
             'club': 'club.Nom',  # str
             'league': 'club.Ligue',  # str
             'city': 'club.Commune',  # str
         }
         # the fields for players without a club
-        no_club_fields: dict[str, tuple[str, types.FunctionType]] = common_fields | {
+        no_club_fields: dict[str, str | int | datetime | None] = common_fields | {
```
Amaras: Same as above?
```suggestion
        no_club_fields: dict[str, str] = common_fields | {
```

src/plugins/ffe/ffe_access_database.py
```diff
@@ -20,7 +19,7 @@ def __init__(self, file: Path, write: bool = False):
 
     def read_player_dicts(self) -> Iterator[dict[str, str | int | datetime | None]]:
         # the fields common to players with a club or not
-        common_fields: dict[str, tuple[str, types.FunctionType]] = {
+        common_fields: dict[str, str | int | datetime | None] = {
```
Amaras: Looks like it should be 
```suggestion
        common_fields: dict[str, str] = {
```

src/plugins/hookspec.py
```diff
@@ -31,7 +32,8 @@
 hookspec = pluggy.HookspecMarker(APP_NAME)
 hookimpl = pluggy.HookimplMarker(APP_NAME)
 
-
+# pylint: disable=empty-body
+# mypy: disable-error-code=empty-body
```
Amaras, line 36: Yeah, better than adding `#type: ignore` at every line

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -283,6 +282,7 @@ def _admin_validate_event_update_data(
         errors: dict[str, str] = {}
         uniq_id: str | None = WebContext.form_data_to_str(data, 'uniq_id')
         if action == 'delete':
+            assert admin_event is not None
```
Amaras: This is probably a good case for an explicit `ValueError` or `RuntimeError`: if you want to delete a non-existant event, there's gonna be a problem

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -305,6 +305,7 @@ def _admin_validate_event_update_data(
                                 'Event [{uniq_id}] already exists.'
                             ).format(uniq_id=uniq_id)
                     case 'update':
+                        assert admin_event is not None
```
Amaras: Same as above, there should probably be an explicit raise

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -440,15 +441,21 @@ def _admin_validate_event_update_data(
         per_plugin_tournament_data = plugin_manager.hook.get_validated_event_form_fields(action=action, event=admin_event, data=data, errors=errors)
         plugin_data = {key: value for data in per_plugin_tournament_data for key, value in data.items()}
 
+        assert uniq_id is not None
+        assert name is not None
+        assert federation is not None
+        assert start is not None
+        assert stop is not None
```
Amaras, line 457: Alright with the post-condition checks, but there might be code paths that don't return when those are `None`.
Maybe we should focus on that

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -534,14 +541,16 @@ def _prepare_event_modal_data(
...
+    ) -> dict[str, Any]:
         uniq_id: str | None = None
         name: str | None = None
         match action:
             case 'update':
+                assert admin_event is not None
                 name = admin_event.stored_event.name
                 uniq_id = admin_event.stored_event.uniq_id
             case 'clone':
+                assert admin_event is not None
```
Amaras: shouldn't you also raise on `action == 'delete'`?

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -559,6 +568,7 @@ def _prepare_event_modal_data(
         stop: float | None = None
         match action:
             case 'update' | 'clone':
+                assert admin_event is not None
```
Amaras, line 582: there might be a way to combine both `match` statement so we avoid this repetition

src/web/controllers/admin/base_admin_controller.py
```diff
@@ -591,6 +601,7 @@ def _prepare_event_modal_data(
         message_background_color: str | None = None
         match action:
             case 'update' | 'clone':
+                assert admin_event is not None
```
Amaras: Could this `match` statement also be combined with the previous ones?

src/web/controllers/admin/base_event_admin_controller.py
```diff
@@ -56,6 +55,7 @@ def template_context(self) -> dict[str, Any]:
         }
 
     def get_tournament_options(self) -> dict[str, str]:
+        assert self.admin_event is not None
```
Amaras: Maybe this should be an explicit `raise`

src/web/controllers/admin/base_event_admin_controller.py
```diff
@@ -68,6 +68,7 @@ def _get_admin_event_render_context(
         cls,
         web_context: BaseEventAdminWebContext,
     ) -> dict[str, Any]:
+        assert web_context.admin_event is not None
```
Amaras: Maybe this should be an explicit `raise` (or `redirect_error`)?

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -54,6 +54,7 @@ def _admin_event_config_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: Maybe there should be a check to add an error, and it should be put before the `if web_context.error` check

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -68,6 +69,7 @@ def _admin_event_config_render(
             case None:
                 pass
             case 'event':
+                assert action is not None
```
Amaras: Maybe an explicit `raise` is better here?

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -153,6 +155,7 @@ async def htmx_admin_event(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: Same as previously: explicitly check earlier, add an error if not, and then return `web_context.errors`

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -246,6 +250,7 @@ def _admin_event_update(
         event_loader = EventLoader.get(request=request)
         match action:
             case 'update':
+                assert web_context.admin_event is not None
```
Amaras: There are three repeated `assert`s on all three relevant code paths.
The other one raises a `ValueError` anyway.
Maybe combine them before the `match` block?
There could also be a case to check earlier as I suggested previously

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -372,36 +379,38 @@ async def htmx_admin_event_print(
...
-    ) -> Template | ClientRedirect:
+    ) -> Template | ClientRedirect | Redirect:
         web_context: BaseEventAdminWebContext = BaseEventAdminWebContext(
             request,
             event_uniq_id=event_uniq_id,
             data=data,
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: Same suggestion as previously

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -372,36 +379,38 @@ async def htmx_admin_event_print(
...
         try:
             document_type = PrintDocumentManager.get_type(
-                WebContext.form_data_to_str(data, field)
+                WebContext.form_data_to_str(data, field) or ''
             )
         except KeyError:
             errors[field] = _('Please choose the document.')
-        if tournament and document_type:
+        assert document_type is not None
+        if tournament:
```
Amaras: Why split them apart?
There are cases where a user will enter an invalid document type in the URL, the code should handle this gracefully, not raise an `AssertionError`
At most, the `assert` should be before the final `return`, because we use it there, not if `errors` is not empty

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -644,9 +653,10 @@ async def htmx_admin_event_download_players(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier, add an error, return early anyway

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -45,6 +45,7 @@ def __init__(
             data=data,
             event_uniq_id=event_uniq_id,
         )
+        assert self.admin_event is not None
```
Amaras: Could it be made sure that it is not `None` earlier (e.g. in `BaseEventAdminController`)?

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -80,8 +81,10 @@ def _admin_validate_family_update_data(
         type_: str
         match action:
             case 'create':
+                assert web_context.family_type is not None
```
Amaras: could the `family_type` be made not `None` in `__init__()`?

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -80,8 +81,10 @@ def _admin_validate_family_update_data(
...
                 type_ = web_context.family_type
             case 'update' | 'clone' | 'delete':
+                assert web_context.admin_family is not None
```
Amaras: I feel this could be made not `None` in the type system

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -124,11 +127,14 @@ def _admin_validate_family_update_data(
             else:
                 match action:
                     case 'create' | 'clone':
+                        assert web_context.admin_event is not None
```
Amaras: This check can probably be combined with the identical one in the next `case` block

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -309,6 +322,7 @@ def _admin_event_families_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -371,6 +388,7 @@ def _admin_event_families_render(
                             raise ValueError(f'action=[{action}]')
                     match action:
                         case 'update' | 'clone':
+                            assert web_context.admin_family is not None
```
Amaras: is there a way to combine asserts with the previous ones?

src/web/controllers/admin/family_admin_controller.py
```diff
@@ -567,6 +586,7 @@ def _admin_family_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         super().__init__(
             request, event_uniq_id=event_uniq_id, data=data
         )
+        assert self.admin_event is not None
```
Amaras: This could probably be an explicit `raise`

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -292,15 +305,16 @@ def _admin_event_players_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -401,68 +416,36 @@ def _admin_event_players_render(
...
-                    return plugin_manager.hook.player_club_sort_key(player=player) or (
-                        player.club,
-                        player.last_name,
-                        player.first_name,
-                    )
-
-            case 'tournament':
-                def sort_key(player: Player): # pylint: disable=function-redefined
+        sort_type = SessionHandler.get_session_admin_players_sort(web_context.request)
+        def get_sort_key(player: Player) -> tuple:
```
Amaras, line 420: okay, fair enough

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -471,72 +454,72 @@ def sort_key(player: Player): # pylint: disable=function-redefined
...
+            and all(plugin_manager.hook.filter_player(
+                web_context=web_context,
+                template_context=template_context,
+                player=player
+            ))
+        ]
         
+        players: dict[int, Player] = {}
+        for player in sorted(filtered_players, key=get_sort_key):
+            if player.id is not None:
```
Amaras, line 520: hasn't the `filtered_players` already filtered out the players with no id?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1006,6 +993,7 @@ def _admin_player_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1114,7 +1105,10 @@ async def htmx_admin_player_move(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_player is not None
+        assert web_context.admin_tournament is not None
```
Amaras: check earlier for both?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1291,6 +1288,8 @@ async def htmx_admin_player_record(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_player is not None
+        assert web_context.admin_player.tournament is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1345,6 +1344,7 @@ async def htmx_admin_tournament_open_check_in(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_tournament is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1401,6 +1401,7 @@ def _admin_tournament_close_check_in(
...
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1479,7 +1480,9 @@ def _admin_player_check_in_out(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_player is not None
         admin_player: Player = web_context.admin_player
+        assert admin_player.tournament is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1549,6 +1552,7 @@ async def htmx_admin_update_event_players(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1623,6 +1627,7 @@ async def htmx_admin_event_players_diff_modal(
...
```
Amaras: check earlier?

src/web/controllers/admin/rotator_admin_controller.py
```diff
@@ -172,6 +177,7 @@ def _admin_event_rotators_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/rotator_admin_controller.py
```diff
@@ -355,6 +364,7 @@ def _admin_rotator_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/screen_admin_controller.py
```diff
@@ -408,6 +428,7 @@ def _admin_event_screens_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/screen_admin_controller.py
```diff
@@ -791,6 +820,7 @@ def _admin_screen_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/screen_admin_controller.py
```diff
@@ -1027,6 +1063,7 @@ def _admin_screen_sets_update(
                     screen_set_id=screen_set_id,
                     data=data,
                 )
+                assert web_context.admin_screen is not None
```
Amaras: add an error instead?

src/web/controllers/admin/screen_admin_controller.py
```diff
@@ -1043,6 +1080,7 @@ def _admin_screen_sets_update(
             case _:
                 raise ValueError(f'action=[{action}]')
         next_screen_set_id: int | None = None
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/timer_admin_controller.py
```diff
@@ -571,9 +595,11 @@ def _admin_timer_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/timer_admin_controller.py
```diff
@@ -665,10 +696,13 @@ def _admin_timer_update(
                         request, event_uniq_id=event_uniq_id
                     )
                 case 'clone':
+                    assert web_context.admin_timer is not None
                     stored_timer = event_database.add_stored_timer(stored_timer)
+                    assert stored_timer and stored_timer.id is not None
```
Amaras:
```suggestion
                    assert stored_timer is not None and stored_timer.id is not None
```

src/web/controllers/admin/timer_admin_controller.py
```diff
@@ -838,9 +872,11 @@ def _admin_timer_hours_update(
                 raise ValueError(f'action=[{action}]')
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -261,8 +271,9 @@ def _admin_event_tournaments_render(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.admin_event is not None
```
Amaras: check earlier?

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -506,6 +523,7 @@ async def admin_tournament_trf_export(
             request, event_uniq_id, tournament_id, None
         )
         tournament = context.admin_tournament
+        assert tournament is not None
```
Amaras: an explicit `raise` feels more warranted here.
That, or an error redirect

src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -816,7 +840,7 @@ async def htmx_tournament_print_view(
         )
         if web_context.error:
             return web_context.error
-
+        assert web_context.admin_tournament is not None
```
Amaras: check earlier?

src/web/controllers/user/event_user_controller.py
```diff
@@ -265,6 +267,7 @@ def _user_event(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.user_event is not None
```
Amaras: check earlier?

src/web/controllers/user/screen_user_controller.py
```diff
@@ -73,6 +73,7 @@ async def htmx_user_login(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.user_event is not None
```
Amaras: check earlier?

src/web/controllers/user/tournament_user_controller.py
```diff
@@ -234,15 +239,19 @@ async def htmx_user_toggle_check_in(
...
-        if web_context.error:
-            return web_context.error
-        web_context.tournament.check_in_player(
-            web_context.player, not web_context.player.check_in
+        if player_web_context.error:
+            return player_web_context.error
+        assert player_web_context.player is not None
+        assert player_web_context.player.id is not None
+        assert player_web_context.user_event is not None
+        assert player_web_context.tournament is not None
```
Amaras, line 247: maybe check those earlier?

src/web/controllers/user/tournament_user_controller.py
```diff
@@ -275,24 +284,29 @@ def _delete_or_add_illegal_move(
...
             tournament_started=True,
         )
-        if web_context.error:
-            return web_context.error
+        if player_web_context.error:
+            return player_web_context.error
+        assert player_web_context.tournament is not None
+        assert player_web_context.player is not None
+        assert player_web_context.player.id is not None
+        assert player_web_context.user_event is not None
```
Amaras, line 292: check those earlier?

src/web/controllers/user/tournament_user_controller.py
```diff
@@ -391,43 +405,47 @@ def _user_update_result(
...
         )
-        if web_context.error:
-            return web_context.error
-        if round_ not in range(1, web_context.tournament.rounds + 1):
+        if board_web_context.error:
+            return board_web_context.error
+        assert board_web_context.tournament is not None
+        assert board_web_context.board is not None
+        assert board_web_context.board.id is not None
+        assert board_web_context.user_event is not None
```
Amaras, line 421: check those earlier?

src/web/controllers/user/tournament_user_controller.py
```diff
@@ -506,6 +524,7 @@ async def htmx_user_download_event_tournaments(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.user_event is not None
```
Amaras: check earlier?

src/web/controllers/user/tournament_user_controller.py
```diff
@@ -546,6 +565,7 @@ async def htmx_user_download_tournament(
         )
         if web_context.error:
             return web_context.error
+        assert web_context.tournament is not None
```
Amaras: check earlier?

src/web/settings.py
```diff
@@ -44,17 +45,16 @@
...
 static_files_folders = [
     static_files_base_dir,
-    *[path for path in plugin_manager.static_paths],
+    *plugin_manager.static_paths,
 ]
 
 static_files_router: Router = create_static_files_router(
     path='/static',
-    directories=static_files_folders,
+    directories=list(static_files_folders),
```
Amaras, line 53: that's a list already, why is this cast needed?

Amaras (changes requested):
Alright, so a lot of questions and remarks.
I had to change to use my editor in the middle, so I'm not sure I have all comments in this review.

src/data/screen.py
timothyarmes: No: ranking_crosstable: bool | None

src/data/event.py
timothyarmes: Well spotted

src/web/controllers/admin/base_admin_controller.py
timothyarmes: We're asserting that we checked those

src/web/controllers/admin/base_admin_controller.py
timothyarmes: How I wish mypy were that good

src/web/controllers/admin/base_admin_controller.py
timothyarmes: Not if we don't need to access it

src/web/controllers/admin/event_admin_controller.py
timothyarmes: I'll remove for now, leaving the type error as a reminder to handle this case.

src/web/controllers/admin/family_admin_controller.py
timothyarmes: Not easily, perhaps a future PR can look at this

src/web/controllers/admin/family_admin_controller.py
timothyarmes: It's called be htmx_admin_family_create, which has a None family ID, thus no family

src/web/controllers/admin/family_admin_controller.py
timothyarmes: Oh for a better mypy

src/web/controllers/admin/player_admin_controller.py
timothyarmes: The type system can't cope here

src/web/settings.py
timothyarmes: I'll let you try to figure that one out, but it's unhappy without it

src/data/event.py
```diff
@@ -604,10 +609,14 @@ def get_unused_screen_uniq_id(
         """Returns the first unused screen uniq_id looking like base_uniq_id:
         base_uniq_id, or base_uniq_id-2, or base_uniq_id-n+1...
         screen_type is used when the given ID is empty to set an ID that corresponds to the screen type."""
-        assert base_uniq_id is not None or screen_type is not None
+        screen_uniq_id = base_uniq_id
+        if screen_uniq_id is None:
+            if screen_type is None:
+                raise ValueError('Either screen_type or base_uniq_id must be provided.')
```
Amaras, line 615: This can be:
```suggestion
        if screen_uniq_id is None and screen_type is None:
            raise ValueError('Either screen_type or base_uniq_id must be provided.')
```

src/data/event.py
```diff
@@ -650,9 +660,13 @@ def get_unused_family_uniq_id(
         """Returns the first unused family uniq_id looking like base_uniq_id:
         base_uniq_id, or base_uniq_id-2, or base_uniq_id-n+1...
         family_type is used when the given ID is empty to set an ID that corresponds to the family type."""
+        family_uniq_id = base_uniq_id
+        if family_uniq_id is None:
+            if family_type is None:
+                raise ValueError('Either family_type or base_uniq_id must be provided.')
```
Amaras, line 666: simplify
```suggestion
        if family_uniq_id is None and family_type is None:
            raise ValueError('Either family_type or base_uniq_id must be provided.')
```

src/data/tie_break.py
```diff
@@ -167,8 +169,10 @@ def adjusted_score(
                 ):
                     score += Result.DRAW.points(tournament.point_values)
                 else:
+                    assert pairing.result is not None
                     score += pairing.result.points(tournament.point_values)
             else:
+                assert pairing.result is not None
```
Amaras, line 174: I think it should be a saparate case:
if `pairing.result` is `None`, ignore this `Pairing` and continue

src/data/timer.py
```diff
@@ -75,7 +78,7 @@ def order(self) -> int | None:
     @cached_property
     def round(self) -> int:
         try:
-            return max(int(self.uniq_id), 0)
+            return max(int(self.uniq_id or 0), 0)
```
Amaras, line 81: weird typing, I think it should be
```suggestion
            return max(int(self.uniq_id) or 0, 0)
```

src/data/tournament.py
```diff
@@ -583,8 +587,8 @@ def read_papi(self, update: bool = False):
             self._rounds = 0
             self._players_by_id = {}
             self._current_round = 0
-            self._rating_limit1 = None
-            self._rating_limit2 = None
+            self._rating_limit1 = 0
+            self._rating_limit2 = 0
```
Amaras, line 591: seems slightly dangerous

src/database/access/access_database.py
```diff
@@ -52,20 +52,30 @@ def __enter__(self) -> Self:
...
+                self.cursor = None
             self.database.close()
             del self.database
             self.database = None
 
+    def _check_cursor(self):
+        """Check that the cursor is available."""
+        if self.cursor is None:
+            raise RuntimeError("Database connection not established")
+
```
Amaras, line 67: This looks like a `TypeGuard` use case: https://docs.python.org/3.12/library/typing.html#typing.TypeGuard
This will be able to be changed to `TypeIs` when we upgrade to Python 3.13: https://docs.python.org/3.12/library/typing.html#typing.TypeGuard

src/database/sql_server/sql_server.py
```diff
@@ -139,15 +140,23 @@ async def connect_to_server():
...
-            self.cursor = None
+            if self.cursor is not None:
+                await self.cursor.close()
+                del self.cursor
+                self.cursor = None
             await self.database.close()
             del self.database
             self.database = None
 
+    def _check_cursor(self):
```
Amaras, line 151: Same as for `AccessDatabase`: consider `TypeGuard`

src/web/controllers/admin/event_admin_controller.py
```diff
@@ -372,32 +385,34 @@ async def htmx_admin_event_print(
...
-        field: str = 'tournament_id'
+        field = 'tournament_id'
+        
         try:
-            tournament = web_context.admin_event.tournaments_by_id[
-                WebContext.form_data_to_int(data, field)
-            ]
+            tournament_id = WebContext.form_data_to_int(data, field)
+            if not tournament_id:
+                raise ValueError()
```
Amaras: please add a message

src/data/event.py
timothyarmes: No it can't, otherwise screen_uniq_id = ... is no longer in an if statement

src/data/event.py
timothyarmes: Same problem

src/data/tie_break.py
timothyarmes: This PR is just for type checking.  I shall remove the assertion and the code can be revisited later.

Amaras (changes requested):
A few things to consider.
I feel this is finally close to merging

src/data/timer.py
timothyarmes: `max(int(self.uniq_id) or 0, 0)` fails since int can't be passed None.

src/data/tournament.py
timothyarmes: @pascalaubry this is your change, can you confirm that it's ok?

src/data/timer.py
Amaras: Alright then:
```suggestion
            return max(int(self.uniq_id or '0'), 0)
```
Ugly, but it's consistent (not that I recommend it)

Amaras (approved):
I'm okay with this now

PR #311, opened 2025-03-28
Add a contribution guide

Amaras:


pascalaubry (approved):
Everything clear to me.
Note: we (developers) also have to stick to these rules, sometimes we miss a clear description of the motivations that lead to the PR, how the change solve problems, ...

PR #312, opened 2025-03-29
Moved the initial language selection from the CLI to the UI

pascalaubry:


src/web/templates/admin/index/config_modal.html
```diff
@@ -2,102 +2,140 @@
     <div class="modal-content">
         <div class="modal-header">
             <h2 class="modal-title">
-                {{ _('Edition of papi-web settings') }}
+                {% if papi_web_config.stored_config.locale %}
+                    {{ _('Edition of papi-web settings') }}
+                {% else %}
+                    {{ _('Choose your language') }}
```
Molrn, line 8: missing from the translation file, as it is pre-locale selection should probably not be translated anyway

src/web/controllers/admin/index_admin_controller.py
```diff
@@ -7,13 +7,13 @@
...
-from litestar.contrib.htmx.response import HTMXTemplate, ClientRedirect
-from litestar.contrib.htmx.response import ClientRedirect
+from litestar.contrib.htmx.response import ClientRedirect, HTMXTemplate
 from litestar.enums import RequestEncodingType
 from litestar.params import Body
 from litestar.response import Template, Redirect
 
-from common.i18n import _, EXPERIMENTAL_FEATURES, locale_localized_name, trusted_locales, untrusted_locales, DEFAULT_LOCALE
+from common.i18n import _, EXPERIMENTAL_FEATURES, locale_localized_name, trusted_locales, untrusted_locales, \
+    DEFAULT_LOCALE
```
Molrn:
```suggestion
from common.i18n import (
    _, 
    DEFAULT_LOCALE,
    EXPERIMENTAL_FEATURES, 
    locale_localized_name, 
    trusted_locales, 
    untrusted_locales,
)
```

Molrn (dismissed):
Good code-wise, just one point feature-wise: the force_edit now does not work, the settings modal does not open once you've selected the language. if this is intended, no problem, does not necessarly have too for me ;)

Other small thing, not super important but would be better without IMO: the empty modal footer looks a bit weird

image

src/web/templates/admin/index/config_modal.html
pascalaubry: I feel that even this message should be translated in a world where we get the default language from the system, added it to the PO files.

Molrn (dismissed):
LGTM

Molrn (approved):

PR #316, opened 2025-03-31
Added support for SVG background images

pascalaubry:
Also added BA custom files


src/custom/FFSU.svg
Molrn, line 1: should be in the custom/FRA dir IMO, feels weird to have at the root with the FIDE logo

Molrn (dismissed):
works well, looks weird though...
image  
maybe it would be better to lower the opacity of the image, and to add the same logo but without text ?

Amaras:
> maybe it would be better to lower the opacity of the image, and to add the same logo but without text ?

The logo without the text is the the correct logo any more.
There should probably be a styling option for images with a format that support it (namely transparency)

pascalaubry:
> > maybe it would be better to lower the opacity of the image, and to add the same logo but without text ?
> 
> The logo without the text is the the correct logo any more. There should probably be a styling option for images with a format that support it (namely transparency)

In an ideal world the transparency percentage would be a configuration property but I suggest that we do this on a future PR.

Molrn (approved):
LGTM

PR #317, opened 2025-04-01
Version independant migrations

Molrn:


src/database/sqlite/migration.py
```diff
@@ -9,274 +9,328 @@
...
+                f'{self.latest_version}.'
+            )
+        status = True
+        if version < self.latest_version:
+            status = False
+        migration = self.get_migration()
+        if migration not in self.migrations + [self.MIGRATION_ZERO]:
+            if migration > self.migrations[-1]:
+                message = (
+                    'Database can only be opened by a next version of Papi-Web'
```
Amaras: Could this use `common.APP_NAME`, somehow?
It would make sure we don't have to change all the instances of "Papi-Web" (and variants) when we change name.

Amaras (commented):
Looks pretty good, provided conflicts can be resolved easily.
Haven't run the code

src/common/papi_web_config.py
```diff
@@ -10,7 +10,7 @@
 import uvicorn
 from packaging.version import Version
 
-from common import BASE_DIR, EXPERIMENTAL_FEATURES, EVENTS_DIR
+from common import TMP_DIR, BASE_DIR, EXPERIMENTAL_FEATURES, EVENTS_DIR, PAPI_WEB_VERSION
```
Amaras: Is `TMP_DIR` still used?

Amaras (commented):
Thanks, code is good, there are just a few warnings, we'll be able to correct that later :)

timothyarmes (approved):

PR #318, opened 2025-04-01
Fix database badges

Molrn:
database badges were broken by the style property refactor to the css class `database-badge`  
image
I'm convinced to have tested it and that it worked fine. Reverting it to before the refactor fixes it, @timothyarmes sorry for modifyng this part in the first place


Molrn:
still a mystery to me, but fix does not seem to be necessary

PR #319, opened 2025-04-02
Test the FFE password from the UI

timothyarmes:


src/plugins/ffe/ffe_event_controller.py
```diff
@@ -53,3 +59,42 @@ async def htmx_ffe_event_tab(
             request,
             event_uniq_id=event_uniq_id,
         )
+    
+    @post(
+        path='/ffe/test-password',
+        name='ffe-admin-tournament-test-auth',
+    )
+    async def htmx_admin_timer_clone(
```
Molrn: Path, name and function name not matching
```suggestion
    @post(
        path='/admin/ffe/test-auth',
        name='admin-ffe-test-auth',
    )
    async def htmx_admin_ffe_test_auth(
```

src/plugins/ffe/ffe_event_controller.py
```diff
@@ -53,3 +59,42 @@ async def htmx_ffe_event_tab(
...
+
+        if NetworkMonitor.connected():
+            ffe_id = data['ffe_id']
+            ffe_password = data['ffe_password']
+            
+            if ffe_id and ffe_password:
+                ffe_auth_valid = FFESession(tournament=None, debug=False).test_auth(ffe_id=ffe_id, ffe_password=ffe_password)
+
+        errors = {}
+        # Compare to False, None means 'unable to check'
```
Molrn, line 85: this doc should be in the `test_auth` method

Molrn (changes requested):
Good code-wise :) did not know morphdom, really think it could be useful elsewhere in the code, thanks for introducing it !  

Feature-wise, I don't have valid FFE credentials to test it ^^' I've still got a question: it is possible to input a `FFE certification number` without a `FFE password` without failing the form. Is there a use-case for this ? I was expecting for it to fail

Molrn (approved):
LGTM

PR #320, opened 2025-04-02
Don't show tooltip when closing modal (close #217)

### LINKED ISSUE #217: Tooltip opens when closing a modal
When a modal is closed and the page not reloaded, the tooltip of the button opening the modal stays open even if the mouse is not on it.  

Image

timothyarmes:


Molrn (commented):
great idea on the fix, but tooltips are not reenabled when a modal is closed by canceling or clicking outside of it

Molrn (approved):
LGTM

PR #326, opened 2025-04-03
Remove status check from database opening

Molrn:


timothyarmes (approved):
LGTM.  I can still use the app :)

PR #328, opened 2025-04-03
Fix running of tests + some tie-break types

timothyarmes:


src/common/papi_web_config.py
```diff
@@ -259,17 +259,17 @@ def lan_ip(self) -> str | None:
...
 
     @property
-    def lan_url(self) -> str:
+    def lan_url(self) -> str | None:
         """The URL of the application on the LAN/WAN."""
         return self._url(self.lan_ip)
 
     @property
-    def local_url(self) -> str:
+    def local_url(self) -> str | None:
```
Molrn, line 272: why ?

src/data/event.py
```diff
@@ -631,7 +631,7 @@ def get_unused_screen_name(
         screen_type is used when the given name is empty to set a default name that corresponds to the screen type."""
         return self._get_unused_item_name(
             base_name or screen_type.name,
-            [screen.name for screen in self.basic_screens_by_id.values() if screen.name is not None],
+            [str(screen.name) for screen in self.basic_screens_by_id.values() if screen.name is not None],
```
Molrn, line 634: I don't get it, most names are typed as str | None but digging into the function it turns out it actually never is None... it is too deep to change here, but it really is a pain

Amaras (commented):
Code looks good, not approving till I run the tests

Amaras (approved):

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -171,10 +171,11 @@ async def search_player(
                     f'FROM joueur LEFT JOIN club on joueur.ClubRef = club.Ref '
                     f'WHERE joueur.NrFFE = ?'
                 )
-                await self.execute(query, string)
+                await self.execute(query, (string,))
                 return (
-                    self._get_player_from_row(row)
+                    player
                     async for row in self.fetchall()
+                    if (player := self._get_player_from_row(row)) is not None
```
Molrn, line 178: this should be fixed th other way around. _get_player_from_row should not accept None and always return a Player

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -211,8 +212,9 @@ async def search_player(
             params += [limit, ]
         await self.execute(query, tuple(params), )
         return (
-            self._get_player_from_row(row)
+            player
             async for row in self.fetchall()
+            if (player := self._get_player_from_row(row)) is not None
```
Molrn, line 217: same

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -254,6 +256,7 @@ async def get_players_by_ffe_licence_number(
         )
         await self.execute(query, tuple(player_ffe_licence_numbers))
         return (
-            self._get_player_from_row(row)
+            player
             async for row in self.fetchall()
+            if (player := self._get_player_from_row(row)) is not None
```
Molrn, line 261: +1

src/plugins/ffe/ffe_tie_break.py
```diff
@@ -102,14 +105,15 @@ def _papi_buchholz_cut(tournament_rounds: int) -> int:
...
-        player: Player,
+        player: TournamentPlayer,
         *,
         after_round: int | None,
         use_cut_top: bool = False,
         use_cut_btm: bool = False,
     ) -> float:
         if after_round is None:
             after_round = max(player.pairings)
+        assert player.tournament is not None
```
Molrn, line 116: Wow that's a big no... we had tournament as a parameter of the tie-break functions, now in each one we've got this:
```py
        assert player.tournament is not None
        tournament: 'Tournament' = player.tournament
```
that's a lot of useless code duplication... I'd really like these removed, if you don't feel like it in this PR I'll do it later

PR #330, opened 2025-04-03
The pre-commit run PR

Amaras:
This PR contains the full changes recommended by the pre-commit hooks.

I had to adjust the pre-commit configuration to exclude the vendored web libraries, although the exclude directive could be better off at the top level.


pascalaubry (approved):
Works fine:
image

PR #334, opened 2025-04-04
Refactor utils and entity architecture

Molrn:
Not to be included in 2.5.0
- src
  - utils
    - \_\_init__
      - StaticUtils
      - SharedUtils
  - enum
  - entity (Abstract classes of the entity architecture)
  
@register decorators removed  
Entitiy managers gathered in one file  
Abstract class prefixes removed


src/data/entity_managers.py
```diff
@@ -0,0 +1,135 @@
...
+            tie_break.AveragePerfectPerformanceTieBreak,
+            tie_break.DirectEncounterTieBreak,
+        ] + list(
+            itertools.chain.from_iterable(
+                plugin_manager.hook.get_extra_tie_break_classes()
+            )
+        )
+
+
+class PapiTieBreakManager(EntityManager[TieBreak]):
```
Amaras: shouldn't this class be put in the ffe plugin?

Amaras (commented):
Before I go further, why did you remove type checking imports?
It's raising errors in the type checker

src/data/entity_managers.py
Molrn: As long as we are relying on papi as data storage, this should be in the core, as it indicates the tie-breaks that are compatible with our data storage. Some of them are not papi specific tie-breaks (ex: number of wins, koya)

Molrn:
> Before I go further, why did you remove type checking imports?

I did not, my guess is that it comes from the pre-commit hook, as they are used like `'{type} | None'` they count as unused imports... I think we should use `Optional['{type}']` for these

Amaras (commented):
I don't know if I want to look at it now, but maybe in a future PR: change `XTieBreak` into `X`.

src/plugins/ffe/ffe_tie_break.py
```diff
@@ -43,7 +44,7 @@ def static_papi_id() -> str:
 class AbstractPapiBuchholzTieBreak(AbstractPapiTieBreak, ABC):
     @staticmethod
     def _papi_adjusted_score(
-        player: 'Player',
+        player: 'TournamentPlayer',
```
Amaras: why the need for quotes if you can directly import the class?

src/utils/entity.py
```diff
@@ -0,0 +1,174 @@
...
+    @property
+    def id(self) -> str:
+        return self.static_id()
+
+    @property
+    def name(self) -> str:
+        return self.static_name()
+
+
+class EntityManager[IdentifiableEntity](ABC):
```
Amaras: This creates a type that is bound to `EntityManager`.
You might want to change to `T: IdentifiableEntity` (not sure it will do exactly the same, though)

Amaras (commented):
Code looks alright.
Not approving now since it's supposed to be merged after 2.5 is released

timothyarmes (approved):
LGTM. The folder hierarchy helps a lot.
Tested.

PR #336, opened 2025-04-04
Fix more types

timothyarmes:


src/web/controllers/admin/player_admin_controller.py
```diff
@@ -536,7 +536,7 @@ def get_sort_key(player: Player) -> tuple:
         for player in sorted(filtered_players, key=get_sort_key):
             if player.id is not None:
                 players[player.id] = player
-
+        print(PlayerUpdaterManager.options())
```
Molrn:
```suggestion
```

src/data/input_output.py
```diff
@@ -72,19 +72,28 @@ def route_parameters(self) -> dict[str, Any]:
 
 
 @dataclass
-class PlayerMatch:
+class PlayerMatcher(ABC):
```
Molrn: I don't get this renaming. A matcher would be an object that creates matches. Here, we've got an object containing two players that are matching, so it makes sense to name it this way. The action that this class does is not the matching, it's to handle the match once it's done

src/data/input_output.py
```diff
@@ -121,9 +135,21 @@ def diff_field_ids(self) -> list[str] | None:
...
+                    and match_date
+                    and (
+                        (src_date.year != match_date.year)
+                        or (
+                            (match_date.month, match_date.day) != (1, 1)
+                            and (src_date.month, src_date.day)
+                            != (match_date.month, match_date.day)
+                        )
+                    )
+                )
```
Molrn: sub function for that

src/data/input_output.py
```diff
@@ -109,8 +118,13 @@ def diff_field_ids(self) -> list[str] | None:
         field_id: str = 'club'
         if field_id in self.field_ids:
             if (
-                not self.player.club
-                or self.player.club.name != self.match_player.club.name
+                (not self.player.club and self.match_player.club)
+                or (self.player.club and not self.match_player.club)
```
Molrn: this was not included on purpose, if the data is missing in one database we'd still want to keep it

src/data/input_output.py
```diff
@@ -257,13 +294,12 @@ def _create_player_matches(
...
-    ) -> list[PlayerMatch]:
-        if match_type is None:
-            match_type = PlayerMatch
-        player_matches: list[match_type] = []
+        matcher: type[PlayerMatcher],
+    ) -> list[PlayerMatcher]:
+        player_matches: list[PlayerMatcher] = []
         for player in players:
-            player_match = match_type(
+            player_match = matcher(
```
Molrn: if the class is Matcher, the objects are `_matcher`

Molrn (changes requested):

src/data/input_output.py
```diff
@@ -72,19 +73,42 @@ def route_parameters(self) -> dict[str, Any]:
...
+        src_date is not None
+        and match_date is not None
+        and (
+            (src_date.year != match_date.year)
+            or (
+                (match_date.month, match_date.day) != (1, 1)
+                and (src_date.month, src_date.day) != (match_date.month, match_date.day)
+            )
+        )
+    )
```
Molrn: sorry I was not clear: doing a sub function is mainly to avoid having a deep nested boolean statement. If you could switch it to simple if-else statements it would be much clearer.

Also if it could be a method of PlayerComparator it would be better, trying to avoid functions laying around like that

Amaras: could you separate the conditions, the expression is quite hard to follow?

Molrn (changes requested):

src/data/tournament.py
```diff
@@ -962,18 +967,21 @@ def build_boards(
                     key=attrgetter('rating'),
                     reverse=True,
                 )
-                weak_time = self.time_control_initial_time
+                weak_time = self.time_control_initial_time or 0
                 rating_diff = strong_player.rating - weak_player.rating
-                penalties = rating_diff // self.time_control_handicap_penalty_step
+                penalties = rating_diff // (
+                    self.time_control_handicap_penalty_step or 0
```
Amaras: not a good idea, it will raise a `ZeroDivisionError` if the code path is somehow executed with a `None` value.
The best case would be to separate the cases and set `penalties = 0` if `self.time_control_handicap_step is None`

Amaras (dismissed):
Code looks pretty good, apart from the two remarks

Molrn (dismissed):
~LGTM~ division by 0 to fix first

Molrn (approved):
LGTM

src/data/input_output.py
```diff
@@ -72,19 +72,43 @@ def route_parameters(self) -> dict[str, Any]:
...
+            return match_date is not None
+        if match_date is None:
+            return False
+        if src_date.year != match_date.year:
+            return True
+        if (match_date.month, match_date.day) == (1, 1):
+            return False
+        if (src_date.month, src_date.day) == (match_date.month, match_date.day):
+            return False
+        return True
```
Amaras, line 108: Could be simplified into:
```suggestion
        return (src_date.month, src_date.day) != (match_date.month, match_date.day)
```

PR #339, opened 2025-04-04
Fix tournament Rating and Pairing with None value

Molrn:


src/data/tournament.py
```diff
@@ -277,6 +277,8 @@ def stored_tie_breaks(self) -> list[AbstractTieBreak] | None:
             TieBreakOptionManager.type_by_id()
         )
         for tie_break_dict in self.stored_tournament.tie_breaks:
+            assert isinstance(tie_break_dict['type'], str)
+            assert isinstance(tie_break_dict['options'], dict)
```
Amaras, line 281: Are you sure `'type'` and `'options'` are keys in `tie_break_dict`?

Molrn: yes, it's the format of the tie-breaks, event with no option you've got a `'options': {}`

Amaras (approved):

PR #344, opened 2025-04-05
Fix SQL Server error message

pascalaubry:
image


src/database/sql_server/sql_server.py
```diff
@@ -116,9 +116,14 @@ async def connect_to_server():
             else:
                 error: str = _('Connection to the FFE server failed.')
             logger.error(error)
-            raise PapiWebException(
-                error or _('Connection to the FFE server failed.')
-            ) from e
+            raise PapiWebException(error) from e
+        except TimeoutError as e:
+            NetworkMonitor.set_connected(False)
+            error: str = _('Connection to the FFE server failed: {error}.').format(
```
Molrn: Would be cool to remove the FFE specific logs from the core... would you be ok with messages like `Connection to the server failed.` ?

Amaras: +1 on that

timothyarmes (commented):
Look like you need to re-run the translations script?

timothyarmes (approved):

PR #346, opened 2025-04-05
Fix FFE database update

Molrn:
25000 records with `"'{fide_id}'"` instead of `"{fide_id}"`


src/plugins/ffe/ffe_database.py
```diff
@@ -124,7 +124,7 @@ def _populate_from_source_file(self, database: SQLiteDatabase) -> bool:
             'standard_rating_type': PlayerRatingType.from_papi_value,
             'rapid_rating_type': PlayerRatingType.from_papi_value,
             'blitz_rating_type': PlayerRatingType.from_papi_value,
-            'fide_id': lambda s: int(s.strip()) if s else 0,
+            'fide_id': lambda s: int(s.strip('\' ')) if s else 0,
```
Amaras:
```suggestion
            'fide_id': lambda s: int(s.strip("' ")) if s else 0,
```

pascalaubry (approved):

PR #349, opened 2025-04-06
Fix all remaining type issues

timothyarmes:


src/data/tie_breaks/tie_breaks.py
```diff
@@ -493,6 +492,7 @@ def compute_player_value(
                 else:
                     scores.append(dummy_points)
                 continue
+            assert pairing.opponent_id is not None
```
Amaras, line 495: This is not always the case.
It should never happen, but in unusual tournament formats (e.g knock-out), this doesn't always work.
They aren't supported yet, but it could be interesting in the future.

src/data/tie_breaks/tie_breaks.py
```diff
@@ -975,7 +975,7 @@ def compute_player_value(
             Result.FORFEIT_LOSS: 0,
             Result.DOUBLE_FORFEIT: 0,
         }
-        return sum(pairing.result.points(score_by_result) for pairing in pairings)
+        return int(sum(pairing.result.points(score_by_result) for pairing in pairings))
```
Amaras: Is this typing correct?
You're returning an `int` when the method is supposed to return a `float`

src/data/tie_breaks/tie_breaks.py
```diff
@@ -1243,15 +1245,16 @@ def compute_player_value(
         while not isclose(
             actual_score,
             mid_score := self._expected_score(
-                (mid := (low + high) / 2), ratings, tournament.point_values
+                (mid := StaticUtils.round_ranking((low + high) / 2)),
```
Amaras, line 1250: I'm not sure this loop won't be infinite.
I don't know how much 1 rating point can affect the predicted score (`mid_score`)

src/data/pairing.py
```diff
@@ -100,11 +100,12 @@ def requested_bye(self) -> bool:
     def color_papi_value(self) -> str:
         if self.color:
             return self.color.to_papi_value
-        return 'F' if self.result.is_bye else 'R'
+        return 'F' if self.result and self.result.is_bye else 'R'
 
     def to_trf(
         self, round_number: int, player_id_to_trf_id: Callable[[int], int]
     ) -> TrfGame:
+        assert self.result is not None
```
Amaras: You don't need that any more

src/data/pairing.py
```diff
@@ -121,4 +122,5 @@ def to_trf(
         )
 
     def __repr__(self):
+        assert self.result is not None
```
Amaras: You don't need that any more

src/data/pairing.py
```diff
@@ -100,11 +100,12 @@ def requested_bye(self) -> bool:
     def color_papi_value(self) -> str:
         if self.color:
             return self.color.to_papi_value
-        return 'F' if self.result.is_bye else 'R'
+        return 'F' if self.result and self.result.is_bye else 'R'
```
Amaras: maybe we could implement a `__bool__` method on Result, but this does not seem needed any more

Amaras (changes requested):
The changes make sense.
Just one nitpick and one doubt

src/data/tie_breaks/tie_breaks.py
timothyarmes: I'm not sure what to do then since if it isn't the code following that line will break.

src/data/tie_breaks/tie_breaks.py
timothyarmes: Oops

src/data/tie_breaks/tie_breaks.py
timothyarmes: We're currently asserting that in a few places.  We'll change it if and when it's no longer true

Amaras: fair enough

Amaras (approved):

PR #352, opened 2025-04-07
Log to files

Molrn:


src/common/logger.py
```diff
@@ -1,46 +1,99 @@
...
+            'class': 'logging.StreamHandler',
+            'level': logging.INFO,
+            'formatter': 'colored',
+            'stream': 'ext://sys.stdout',
+        },
+        'file': {
+            'class': 'logging.handlers.RotatingFileHandler',
+            'level': logging.DEBUG,
+            'formatter': 'standard',
+            'filename': str(LOGS_DIR / f'{APP_NAME}.log'),
```
timothyarmes, line 45: Mypy error: src\common\logger.py:45: error: Cannot determine type of "LOGS_DIR"  [has-type]

timothyarmes (changes requested):

src/common/logger.py
Molrn: does not show up on my MyPy... don't really know what to do about it. maybe `LOGS_DIR = Path('logs')` --> `LOGS_DIR: Path = Path('logs')` in the common.__init__ ? if you detect it can you please handle it ?

timothyarmes (dismissed):

pascalaubry (approved):

PR #354, opened 2025-04-07
2.5 colorlog ModuleNotFoundError fix

pascalaubry:


Amaras (approved):
Works better for me.
Bypassing code owner just this once (because the exported version is unusable)

PR #360, opened 2025-04-08
Default all modal targets to body

timothyarmes:


Molrn (changes requested):
please put back the 4 spaces tab so I can have a look at what's actually been changed. I guess it's only the hx-targets

Molrn (approved):
LGTM

PR #366, opened 2025-04-08
Fix version problems that break dev

Amaras:
`dev` was broken because of an version format check that did not follow our new versioning scheme.

Versions can now in one of 4 formats:
1. "classic" versions: X.Y.Z
2. alpha versions: X.Y.ZaN
3. beta versions: X.Y.ZbN
4. release candidates: X.Y.ZrcN

This means, in particular, that X.YrcN is no longer valid.

I can launch the server with this code.


Molrn (approved):
LGTM

timothyarmes (approved):

PR #370, opened 2025-04-10
Recover configuration databases from previous releases

pascalaubry:
Recover the configuration database from previous releases
Updated the project name, main URL, API URL and contact email
Do not load the configuration database from export.py
Version 2.5.3

Note: changes must also be merged to dev.


src/common/papi_web_config.py
```diff
@@ -153,7 +154,7 @@ def copyright(self) -> str:
     @property
     def project(self) -> str:
         """The project of the application."""
-        return _('Papi-web project')
+        return _('Sharly Chess project')
```
Molrn, line 157: IDK about that, not a fan of changing the name of the project (visible on every page) in a patch. I think we'd better keep it as is with papi-web everywhere on stable-2.5, and have one full `Papi-Web to Sharly-Chess` PR on dev

docs/03-change-log.md
```diff
@@ -2,6 +2,9 @@
 
 # Papi-web - ChangeLog
 
+## Version 2.5.3 - 10 avril 2025
+- Récupération de la configuration de l'application à partir d'une version antérieure
+
```
Molrn:
```suggestion
## Version 2.5.3 - 10 avril 2025
- Récupération de la configuration de l'application à partir d'une version antérieure
- Modification de l'encodage des fichiers de log
```

Molrn (dismissed):

src/common/papi_web_config.py
pascalaubry: Did not change any occurence of Papi-web except this one because the name of the product did not change yet, but the name of the project did.

Molrn (approved):
LGTM

PR #376, opened 2025-04-10
Add papi Info to database and forms

Molrn:
Add fields from the papi Info table  to the database and forms. Same as tie-breaks, papi stays the value of reference.  
Location / arbiter --> Event level  
image  

Rounds / Rating --> Tournament level  
image

### DIFF
--- a/locale/fr/LC_MESSAGES/messages.po
+++ b/locale/fr/LC_MESSAGES/messages.po
@@ -38,8 +38,8 @@ msgid "%(set_name)s (round %(round)d)"
 msgstr "%(set_name)s (ronde %(round)d)"
 
 #, python-format
-msgid "%(title)s:"
-msgstr "%(title)s :"
+msgid "%(string)s:"
+msgstr "%(string)s :"
 
 #, python-format
 msgid "%f to %l"
@@ -537,9 +537,6 @@ msgstr "Chessevent : %(chessevent_user_id)s / %(chessevent_event_id)s"
 msgid "Chessevent: %(chessevent_user_id)s / %(chessevent_event_id)s / %(chessevent_tournament_name)s"
 msgstr "Chessevent : %(chessevent_user_id)s / %(chessevent_event_id)s / %(chessevent_tournament_name)s"
 
-msgid "Chief arbiter"
-msgstr "Arbitre en chef"
-
 msgid "Choose a custom image:"
 msgstr "Choisissez une image personnalisée :"
 
@@ -1021,9 +1018,6 @@ msgstr "Exemple : Pointage commencé depuis %%s"
 msgid "E.g.: DOE"
 msgstr "Exemple : DUPONT"
 
-msgid "E.g.: DOE John"
-msgstr "Exemple: DUPONT Pierre"
-
 msgid "E.g.: John"
 msgstr "Exemple : Pierre"
 
@@ -1041,6 +1035,9 @@ msgstr "Exemple : Mon écran"
 
 msgid "E.g.: My screen family"
 msgstr "Exemple : Ma famille d'écrans"
+
+msgid "E.g.: Paris"
+msgstr "Exemple : Paris"
 
 #, python-format
 msgid "E.g.: Players %%f to %%l"
@@ -1205,6 +1202,9 @@ msgstr "Activer/désactiver l'affichage des détails des tournois sur les cartes
 
 msgid "End of upload (Ctrl-C)"
 msgstr "Fin de mise en ligne (Ctrl-C)"
+
+msgid "End time needs to be after start time."
+msgstr "L'horaire de fin doit être après l'horaire de début."
 
 msgid "End:"
 msgstr "Fin :"
@@ -2348,9 +2348,6 @@ msgstr "Nombre de parties gagnées"
 msgid "Number of players : %(num)d"
 msgstr "Nombre de joueur·euses : %(num)d"
 
-msgid "Number of rounds in the tournament."
-msgstr "Nombre de rondes dans le tournoi."
-
 msgid "Number of wins"
 msgstr "Nombre de victoires"
 
@@ -3326,6 +3323,10 @@ msgstr "Le répertoire du fichier Papi ([%(dir)s] par défaut)."
 msgid "The end date and time of the event."
 msgstr "La date et l'heure de fin de l'évènement."
 
+#, python-format
+msgid "The end time of the tournament (defaults to %(datetime)s)."
+msgstr "L'horaire de fin du tournoi (par défaut %(datetime)s)."
+
 msgid "The federation that organizes the event."
 msgstr "La fédération qui organise l'évènement."
 
@@ -3355,6 +3356,13 @@ msgstr "Le dernier écran ne peut être supprimé."
 
 msgid "The location of the event."
 msgstr "Le lieu de l'évènement."
+
+#, python-format
+msgid "The location of the tournament (defaults to [%(location)s])."
+msgstr "Le lieu du tournoi (par défaut [%(location)s])."
+
+msgid "The location of the tournament."
+msgstr "Le lieu du tournoi."
 
 msgid "The log level allows you to trace the behaviour of the application."
 msgstr "Le niveau de log permet de trancer le comportement de l'application."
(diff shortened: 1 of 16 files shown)


Amaras:
Ideally, Chief Arbiter should be a valid person in FIDE mode and with the FFE plugin enabled.
We still need to allow for unregistered players being arbiters, so it's not that urgent (maybe split it off to an issue if you want)

Chief arbiters could also theoretically change between tournaments of the same event, so maybe also add it to the tournament as well (default to event CA as tournament CA).

Also, a circuit could change locations between tournaments of the same event, so think about putting it to the tournament level (default to event-level location)

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -1014,6 +1021,9 @@ msgstr "Exemple : Pointage commencé depuis %%s"
 msgid "E.g.: DOE"
 msgstr "Exemple : DUPONT"
 
+msgid "E.g.: DOE John"
+msgstr "Exemple: DUPONT Pierre"
```
pascalaubry: Exemple : (toujours un espace avant les double ponctuations :;!? insécable quand on peut).

src/data/event.py
```diff
@@ -285,6 +285,14 @@ def path(self) -> Path:
             self.add_error(_('[{path}] is not a directory.').format(path=path))
         return path
 
+    @property
+    def location(self) -> str | None:
+        return self.stored_event.location
+
+    @property
+    def arbiter(self) -> str | None:
```
pascalaubry: Should we have one or many arbiters? 
One or many chief arbiters?
Can we have one arbiter now and see later the needs?
Can the arbiter be defined with any ``str`` or should we use official FIDE or FFE ID?
Note: we must have also organizers but we can see this later.
As you see I have more questions than answers, we must discuss this on Discord.

src/database/access/papi/papi_database.py
```diff
@@ -353,6 +367,26 @@ def reset_player_result(self, player_papi_id: int, round_: int):
...
+            ),
+        )
+
+    def set_round_unused(self, round_: int):
+        prefix = f'Rd{round_:0>2}'
+        self._execute(
+            f'UPDATE `joueur` SET `{prefix}Res` = ?, `{prefix}Cl` = ? WHERE `Ref` > 1',
+            (
+                None,
+                None,
```
pascalaubry: I don't remember that these fields can be null in the Papi database.

src/web/templates/admin/event/event_modal.html
```diff
@@ -81,6 +81,26 @@ <h4>
                             ) }}
                         </div>
                     </div>
+                    <div class="row">
+                        <div class="col-12 col-md-6 mb-3">
+                            {{ admin_macros.text_input(
+                                name='location',
+                                label=_('%(title)s:', title=_('Location')),
```
pascalaubry: Why not ``label=_('Location:')``?
If it is to factorize with ``tab.html`` then I think is it a false good idea (we may use different typos, for instance truncate on the card, and forget that the string is also used in another place).

src/web/templates/admin/event/event_modal.html
```diff
@@ -81,6 +81,26 @@ <h4>
...
+                                placeholder=_('E.g.: %(string)s', string='Paris'),
+                                help_text=_('The location of the event.'),
+                                data=data, errors=errors
+                            ) }}
+                        </div>
+                        <div class="col-12 col-md-6 mb-3">
+                            {{ admin_macros.text_input(
+                                name='arbiter',
+                                label=_('%(title)s:', title=_('Chief arbiter')),
```
pascalaubry: Why not ``label=_('Chief arbiter:')``?

src/web/templates/admin/tournaments/tournament_modal.html
```diff
@@ -79,6 +79,45 @@ <h4>
...
+                                macros.tooltip_attributes(
+                                    'info',
+                                    _('Fields relying on the Papi database can\'t be updated if the file does not exist.'),
+                                    'bottom'
+                                )
+                            }}
+                        >
+                            {{ admin_macros.number_input(
+                                name='rounds',
+                                label=_('%(title)s:', title=_('Rounds')),
```
pascalaubry: Same as previously.

Molrn:
@Amaras adding location / arbiter at tournament level, creating an issue for arbiter search

src/web/templates/admin/tournaments/tournament_modal.html
```diff
@@ -79,6 +79,45 @@ <h4>
...
+                                help_text=_('Number of rounds in the tournament.'),
```
pascalaubry: The number of rounds of the tournament.

pascalaubry (changes requested):
A few details and a few things to discuss.

locale/fr/LC_MESSAGES/messages.po
Molrn: exactly the reason why I wanted to factorise ^^'

src/data/event.py
Amaras: Currently, as we use the papi database as the source of truth, we only have the tournament's chief arbiter (there is always only one chief arbiter for a tournament), which is an arbitrary `str`.
Whenever possible, use the FIDE and FFE databases to get all relevant information (currently, only FFE ID and FIDE ID)

src/database/access/papi/papi_database.py
Molrn: That's the case for values from `rounds` to 25 currently. Haven't questioned it tbh, just reproduced it

src/web/templates/admin/event/event_modal.html
Molrn: I think it would be good to have this kind of factorization to not make the job of translators a nightmare with most of the strings being duplicated. If we need to specify usage, it would be much clearer through a `Location **** TOURNAMENT CARDS`.  
I'm open about this, but I really don't think it's a false good idea

src/database/access/papi/papi_database.py
pascalaubry: I see RdXXCl='R', RdXXAdv=NULL, RdXXRes=0 everywhere for not-plaued rounds, don't you?

Molrn: for not played round yes, for out of tournament scope rounds I get NULL / NULL / NULL. Example with domloup-fide-a (5 rounds, not played):
image
round 5 is not played, round 6 is out of scope

locale/fr/LC_MESSAGES/messages.po
pascalaubry: As discussed, just let you change ``title`` by ``string`` everywhere.

src/database/access/papi/papi_database.py
pascalaubry: Discussed yesterday: RdXXCl='R', RdXXAdv=NULL, RdXXRes=0 are the correct values ;-)

pascalaubry (changes requested):
Just a few changes needed, thanks.

src/database/sqlite/event/event_database.py
```diff
@@ -901,6 +894,7 @@ def _row_to_stored_event(self, row: dict[str, Any]) -> StoredEvent:
             stop=row['stop'],
             public=self.load_bool_from_database_field(row['public']),
             path=row['path'],
+            location=row['location'],
```
pascalaubry, line 897: Since this field has been added to the database, I bet that checking the version will fail if you don't use ``row.get('location', None)``.

src/database/sqlite/event/event_database.py
```diff
@@ -1327,11 +1323,16 @@ def _row_to_stored_tournament(cls, row: dict[str, Any]) -> StoredTournament:
             check_in_open=cls.load_bool_from_database_field(
                 row.get('check_in_open', None)
             ),
+            rounds=row['rounds'],
+            rating=row['rating'],
```
pascalaubry, line 1325: Same as previously.

src/database/sqlite/event/event_database.py
```diff
@@ -1327,11 +1323,16 @@ def _row_to_stored_tournament(cls, row: dict[str, Any]) -> StoredTournament:
...
             last_update=row['last_update'],
             last_result_update=row['last_result_update'],
             last_illegal_move_update=row['last_illegal_move_update'],
             last_check_in_update=row['last_check_in_update'],
             tie_breaks=cls.load_json_from_database_field(row.get('tie_breaks', None)),
+            start=row['start'],
+            stop=row['stop'],
+            location=row['location'],
```
pascalaubry, line 1333: Same as previously.

pascalaubry (changes requested):

src/database/sqlite/event/event_database.py
Molrn: I think this issue was a lot related to how and when the version was checked, I'm pretty sure this is not an issue today. Do you have a way to reproduce it ? I can't, app starts well with databases from previous version / previous migrations

pascalaubry: Lost my bet since the version number is now read from table ``metadata`` (nice!).
So I let you removed the now unneeded get() calls.

pascalaubry (approved):

PR #379, opened 2025-04-10
Fix players screen issue

Amaras:
There are two "alert message" fields and checkboxes, which made Litestar fail to validate the query.

I removed the additional alert message, and made the "unpaired players" selection full-width.


Molrn (approved):
LGTM

PR #381, opened 2025-04-10
Initial pairing tab work

timothyarmes:


Molrn:
the definition of FIDE, safe and unsafe modes are missing, so currently all the operations are allowed without warning modal. I'm developing  it in another PR and undrafting so the review of this one can move forward

src/data/tournament.py
```diff
@@ -460,7 +447,10 @@ def boards(self) -> list[Board]:
     @property
```
pascalaubry: Use @cache_property to prevent from sorting twice?

pascalaubry (commented):
I am confused by the fact that the two buttons Unpair and Options are not on the same line as the tournament/round selectors.
image
Is there a reason not to have put them on the same line?

pascalaubry (changes requested):
The tournament below is finished (the 7 rounds have been played).
For any round selected, the check-in status of unpaired players is shown, and this status has no sense but for the current round when the check-in is open.
image
We should not find this information here, except maybe when the current round is selected.

src/web/templates/admin/pairings/tab.html
```diff
@@ -0,0 +1,379 @@
...
+            {% if has_unpaired %}
+                <div class="col-12 col-md-3 mb-3">
+                    {% if admin_unpaired|length > 0 %}
+                        {% with title=_('Unpaired'), unpaired=admin_unpaired %}
+                            {% include '/admin/pairings/unpaired_table.html' %}
+                        {% endwith %}
+                    {% endif %}
+
+                    {% if admin_forfeit_players|length > 0 %}
+                    {% with title=_('Bye-players'), unpaired=admin_forfeit_players %}
```
pascalaubry: Why not simply 'Not-paired players' (which groups ZPB, HPB, FPB and NO_RESULT)?

pascalaubry (changes requested):
The tournament below is finished (the 7 rounds have been played).
When round 1 to 6 is selected, the round dropdown does not show round 7.
image
The next round button is available and works well.
image
When round 7 is selected then round 7 appears in the dropdown list.
image

pascalaubry (changes requested):
This tournament has 6 rounds.
Rounds 1 to 6 are finished.
Round 7 is missing one result.
When clicking on the pairings tab, round 6 is selected; I was expecting round 7 (the current round) by default.
image
Enclosed the Papi file if you want to test.
[35-tc-2025.zip](https://github.com/user-attachments/files/19763394/35-tc-2025.zip)

src/web/static/lib/htmx/htmx-2.0.4/htmx.min.js
pascalaubry, line 1: I think that we should not store HTMX core and extensions to the repository, I added issue #417 to make developers download them as we do for bbpPairings.

pascalaubry (changes requested):
When I select the last board and type ``0`` to remove the result, the board goes to red.
image
The board result has noot been changed in the Papi database.
One line in the logs:
```
25/04/15 20:33:43 INFO      127.0.0.1:56756 - "PUT /admin/pairing/set-result-hotkey/35-tc/1/7 HTTP/1.1" 200
```

src/web/templates/admin/pairings/tab.html
timothyarmes: It's nice to separate unpaired players from those who have a bye

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -267,6 +291,10 @@ msgstr "Évènement archivé"
 msgid "Archived events ({num})"
 msgstr "Évènements archivés ({num})"
 
+#, python-format
+msgid "Are you sure you want to unpair all boards for round %(round_number)s?"
+msgstr "Êtes vous sûr de vouloir désapparier tous les échiquier de la ronde %(round_number)s?"
```
pascalaubry: échiquier > échiquiers

timothyarmes:
> When I select the last board and type `0` to remove the result, the board goes to red. image The board result has noot been changed in the Papi database. One line in the logs:
> 
> ```
> 25/04/15 20:33:43 INFO      127.0.0.1:56756 - "PUT /admin/pairing/set-result-hotkey/35-tc/1/7 HTTP/1.1" 200
> ```

You're in validate mode.

This is what some arbiters do with Papi, and it's the behaviour that i'm trying to maintain:
You get the paper list of results, and without even looking at the screen you enter all the results using the 1, 2, 3 hotkeys.  Then, to easily validate that you haven't made a mistake you go back to the top and start again.  As long as the result is the same you move to the next line, but if it differs it blocks and highlights the row so that you can check your input.  Papi brings up an error and beeps, but the red row is nicer.

src/web/static/lib/htmx/htmx-2.0.4/htmx.min.js
timothyarmes: Not relavent for this PR

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -2463,9 +2559,15 @@ msgstr "Appariements par ordre alphabétique"
 msgid "Pairings by player ({num})"
 msgstr "Appariements par ordre alphabétique ({num})"
 
+msgid "Pairings disabled while check-in is open."
+msgstr "Appariements désactivés lorsque le pointage est ouvert."
```
pascalaubry: Appariements désactivés lorsque le pointage est ouvert.
>
Les appariements sont désactivés lorsque le pointage est ouvert.

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -2699,12 +2801,18 @@ msgstr "Veuillez entrer le nom du tournoi."
 msgid "Please enter valid date and time."
 msgstr "Veuillez entrer une date et une heure valides."
 
+msgid "Please use Papi to generate pairings."
+msgstr "Utiliser Papi pour générer les appariements."
```
pascalaubry: Veuillez utiliser Papi pour générer les appariements.

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3123,6 +3237,12 @@ msgstr "Envoi du tournoi [{ffe_id}] ({file}) sur le site web FFE..."
 msgid "Sharly Chess project"
 msgstr "Projet Sharly Chess"
 
+msgid "Shortcuts check if the inputted result and the result of the selected row are a match."
+msgstr "Les raccourcis vérifient si le résultat saisi et le résultat du rang séléctionné sont identiques."
+
+msgid "Shortcuts set the result of the selected row to the inputted result."
+msgstr "Les raccourcis mettent à jour le résultat du rang séléctionné."
```
pascalaubry: Les raccourcis mettent à jour le résultat du rang séléctionné.
>
Les raccourcis mettent à jour le résultat de l'échiquier sélectionné.

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3123,6 +3237,12 @@ msgstr "Envoi du tournoi [{ffe_id}] ({file}) sur le site web FFE..."
 msgid "Sharly Chess project"
 msgstr "Projet Sharly Chess"
 
+msgid "Shortcuts check if the inputted result and the result of the selected row are a match."
+msgstr "Les raccourcis vérifient si le résultat saisi et le résultat du rang séléctionné sont identiques."
```
pascalaubry: Les raccourcis vérifient si le résultat saisi et le résultat du rang séléctionné sont identiques.
>
Les raccourcis vérifient si le résultat saisi et le résultat de l'échiquier sélectionné sont identiques.
(one accent only)

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -426,6 +461,10 @@ msgstr "Par défaut : %(minutes)d (minutes)"
 msgid "By player"
 msgstr "Par ordre alpha."
 
+#, python-format
+msgid "Bye limit (%(limit)d) is already reached for this tournament."
+msgstr "La limite de points jokers (%(limit)d) est déjà atteinte pour ce tournoi."
```
pascalaubry: points jokers > points joker
Notes: 
- "points jokers" is accepted but "points joker" is nicer..
- Can you also update string "Forfaits et points jokers" > "Forfaits et points joker" at line 4359.
- Consistent with "Trop de points joker pour le joueur [{player_name}]." at line 3866.

pascalaubry (changes requested):

pascalaubry:
> You're in validate mode.
> 
> This is what some arbiters do with Papi, and it's the behaviour that i'm trying to maintain: You get the paper list of results, and without even looking at the screen you enter all the results using the 1, 2, 3 hotkeys. Then, to easily validate that you haven't made a mistake you go back to the top and start again. As long as the result is the same you move to the next line, but if it differs it blocks and highlights the row so that you can check your input. Papi brings up an error and beeps, but the red row is nicer.

``<troll>``Paper? You mean that egyptian thing guys used to write to when they had no rock?``</troll>``
I understand whats you mean but I was not trying to change the result, I used ``0`` to remove the result.

src/data/loader.py
```diff
@@ -54,16 +55,29 @@ def clear_cache(self, event_uniq_id: str | None = None):
...
+        with EventDatabase(event_uniq_id) as database:
+            stored_tournament = database.get_stored_tournament(tournament_id)
+        if not stored_tournament:
+            if tournament_id in event.tournaments_by_id:
+                del event.tournaments_by_id[tournament_id]
+        else:
+            event.tournaments_by_id[tournament_id] = Tournament(
+                event, stored_tournament
+            )
+        self.clear_cache()
```
pascalaubry, line 80: I don't understand why you reload the tournament, and then you clear the whole loader cache (all the events).
Why not simply reload the event as we do everywhere else?

pascalaubry (dismissed):

src/data/loader.py
Molrn: This does one DB query, the one to get the tournament. Fetching an event does one per table, plus a bunch of computation.

- clear_cache(None) only deletes the cached properties, which are computed again from the stored variables _loaded_events_by_id and _loaded_stored_events_by_id once they are accessed again.
- clear_cache(event_id) on the other hand deletes the event from memory, and it has to be fetched again from start.

There are many improvements to do to the Event loader to improve the performances, allowing to reload a tournament is one of them.

Molrn (dismissed):
LGTM (finally ^^)

pascalaubry (approved):

PR #387, opened 2025-04-11
Removed the version number from README.md

pascalaubry:
This PR should also be applied to the dev branch.

### DIFF
--- a/docs/01-summary.md
+++ b/docs/01-summary.md
@@ -67,12 +67,12 @@ Pour être prévenu·e des nouvelles versions de Papi-web, il suffit de s'abonne
 | AUBRY Pascal        | AFO2  | BRE   | 35  | Échiquier Domloupéen                          |
 | AVENIA Adrien       |       | BFC   | 25  | Tour Prends Garde! Besançon                   |
 | BELEY Huong         | AFO1  | IDF   | 77  | Tours et Detours Club d'Échecs Centre Brie    |
-| BELLILI Bachir      | AFO1  | HDF   | 60  | Echiquier Beauvaisien                         |
 | BENETEAU Hubert     | AFE1  | NOR   | 50  | Agneaux Saint-Lô                              |
 | BENADDOU Khaled     | AFE2  | IDF   | 94  | Club d'Échecs de Saint-Maur                   |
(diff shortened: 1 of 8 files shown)


docs/01-summary.md
```diff
@@ -67,6 +67,7 @@ Pour être prévenu·e des nouvelles versions de Papi-web, il suffit de s'abonne
 | AUBRY Pascal        | AFO2  | BRE   | 35  | Échiquier Domloupéen                          |
 | AVENIA Adrien       |       | BFC   | 25  | Tour Prends Garde! Besançon                   |
 | BELEY Huong         | AFO1  | IDF   | 77  | Tours et Detours Club d'Échecs Centre Brie    |
+| BELLILI Bachir      | AFO1  | HDF   | 60  | Echiquier Beauvaisien                         |
```
Amaras:
```suggestion
| CHELLILI Bachir      | AFO1  | HDF   | 60  | Echiquier Beauvaisien                         |
```

Amaras (dismissed):
I'm against removing the version from the readme, we simply need better version management

Molrn:
> I'm against removing the version from the readme, we simply need better version management

That's also why I suggested to only have one PR per patch to bring back the fixes to dev. Still seems like the best solution to me. Exceptions can of course be made for app breaking fixes and fixes required for dev

pascalaubry:
> I'm against removing the version from the readme, we simply need better version management

No problem to throw this PR but please at least remove the version number from the README until we have a better solution (we can leave the README as-is).

Molrn:
> That's also why I suggested to only have one PR per patch to bring back the fixes to dev

Alternatively, we can simply just only implement changelog and version numbers in the last PR before the release.

Amaras:
> Alternatively, we can simply just only implement changelog and version numbers in the last PR before the release.

Yeah, that's what I'm thinking about.
I don't understand why the stable branch got to version 2.5.3 instead of the real stable version

Amaras:
Okay, actually, it's in PR #370.

Just... Don't change versions on stable unless you publish a new release!

pascalaubry:
Well that was THE specific case for which I had to change the release number before pushing because it obviously was needed to test.
If you want to wait for better version handling then we can forget this PR, keep the version number in the README and I change the release number only before publishing a new release.

Molrn (approved):

PR #280, opened 2025-03-22
Simplify i18n

pascalaubry:
By default, do not update untrusted locales
Sort the i18n strings by alphabetical order to limit the changes when moving code

PR #394, opened 2025-04-11
Fix a hole in contribution policy

Amaras:
Currently, we get inconsistent PRs in the i18n.

Some don't update the experimental locales and thus remove lines from `docs/86-i18n.md`, some update them and add them back.

This fixes this hole, and makes us update all locales before proposing a change

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,6 +34,9 @@ Whenever possible, please add tests to ensure your changes do not cause regressi
 
 Make sure to run `pytest` and solve problems before sending your code to review.
 
+If you're changing translation files, make sure you have the `translate` dependency group installed, run `python scripts/i18n/i18n_update.py`, and update all locales.
+A PR with a partial `docs/86-i18n.md` will get changes requested immediately.
+
 Once all the changes are made, create a Pull Request with a descriptive name, and why you made the changes.
 If you're solving an issue, link to it in when opening your PR.
 


pascalaubry (approved):
OK for that.
Is it just the reviewer's job to think of it or can make it automatic with pre-commit in a simple way?

Molrn:
this reverts the previous policy set up with #280. Not updating all untrusted locales was initially made to reduce PR sizes, and as I understood it at the time it was decided to only translate untrusted locales for releases. 

I kinda liked the previous policy, and I don't get the need for that one, the content of `docs/86-i18n.md` not containing untrusted locals does not strike me as this important. I also know it was important to @timothyarmes too

Maybe we could instead make it such that if untrusted locales are not modified the lines in the docs file are not changed ? I would really prefer such a solution to this problem

Amaras:
> this reverts the previous policy set up with #280. Not updating all untrusted locales was initially made to reduce PR sizes, and as I understood it at the time it was decided to only translate untrusted locales for releases.

No explicit policy was set in that PR.
Not updating the untrusted locales is precisely what increases the number of changes to the i18n monitor document.

> I kinda liked the previous policy, and I don't get the need for that one, the content of `docs/86-i18n.md` not containing untrusted locals does not strike me as this important.

The main problem is PRs adding the removing, then adding, then removing them.

> Maybe we could instead make it such that if untrusted locales are not modified the lines in the docs file are not changed ? I would really prefer such a solution to this problem

I agree that this is something to do, or maybe simply move to a more manual approach for that file?

The one thing that should be discussed is whether we should have those unused experimental translations in the first place...

CONTRIBUTING.md
```diff
@@ -34,6 +34,9 @@ Whenever possible, please add tests to ensure your changes do not cause regressi
 
 Make sure to run `pytest` and solve problems before sending your code to review.
 
+If you're changing translation files, make sure you have the `translate` dependency group installed, run `python scripts/i18n/i18n_update.py`, and update all locales.
+A PR with a partial `docs/86-i18n.md` will get changes requested immediately.
+
```
pascalaubry, line 39: Please see #396, I think these two lines are not needed anymore.
Note: installing the ``translate`` dependency group is needed only when you translate empty messages of experimental locales, you do not need it when simply running ``i18n_update.py``.

pascalaubry (changes requested):
I think this PR can be closed with #396.

PR #396, opened 2025-04-12
Removed experimental locales

pascalaubry:
Experimental locales have been removed, new translations can still be added with the translate script.


timothyarmes (approved):
LGTM.

I've pushed changes to the paths to make them work outside of PyCharm

PR #410, opened 2025-04-13
Added option --experimental

pascalaubry:
image
@Amaras Please have a look at i18n_update.py:27

PR #411, opened 2025-04-14
Restyle messages

timothyarmes:


pascalaubry (approved):
Nice, maybe have a look at #410 to see if the look of the bottom experimental warning needs to be also updated.

PR #421, opened 2025-04-17
Improve nav bar

timothyarmes:


pascalaubry (approved):
Really nice, I was just wondering if it was needed or not to keep the number of players in the nav bar since it is not always relevant (especially for tournaments on several FIDE periods).
Please remove it if you aggree with this.

PR #431, opened 2025-04-17
Use datetime format closer to ISO

Amaras:
Logging datetime format was %y/%m/%d %H:%M:%S

This makes it confusing for everyone, with date & times like: 25/04/17 21:26:31.
This is April 7th 2025, but looks like April 25th 2017.
This would be worse for dates fortunately in the past, like: 12/04/17.
In US-like format, this looks like December 4th 2017; in a more usual format (in the rest of the world), it's April 12th 2017; but it's supposed to be April 17th 2012.

Changed to %Y-%m-%d %H:%M:%S (ISO-like) to make sure we have no such problem in the future.
This makes the date 2025-04-17.


Molrn (dismissed):
That's only changing it in the console, but not in the logs. I don't have much opinion about which date format to use, but it should be the same in log files and console

pascalaubry:
Hi
I have changed my mind concerning the format of the logs: actually there is no need anymore to print the date on the console since this information is available in the log files.
Does everybody agree on that?

Molrn:
> there is no need anymore to print the date on the console

fine by me

src/common/logger.py
```diff
@@ -13,8 +13,8 @@
     'formatters': {
         'colored': {
             '()': 'colorlog.ColoredFormatter',
-            'fmt': '%(log_color)s%(asctime)s %(levelname)-10s%(message)s%(reset)s',
-            'datefmt': '%y/%m/%d %H:%M:%S',
+            'fmt': '%(log_color)s%(levelname)-10s%(message)s%(reset)s',
```
pascalaubry:
```suggestion
            'fmt': '%(log_color)s%(message)s%(reset)s',
```
The color is now enough to distinguish levels:
image

pascalaubry (changes requested):

src/common/logger.py
Amaras: Not convinced the colours can be distinguished properly when you are colourblind.
I won't remove the level.
At most, I will put it at the start of the line.

Amaras: téléchargement
With deuteranopia (green-blind colourblindness, simulated above), you can barely differentiate warning and error

Also, people who don't see colour exist.
Removing the level makes it inaccessible to them

pascalaubry (approved):
You are right, I will propose in another PR to make configurable:
- the log level (already done but not used yet I think);
- usage of colours;
- the format of the output (very simple: level and datetime).
No hurry for that.

pascalaubry:
The change has been applied to 2.5.7 in #442.

pascalaubry (approved):

Molrn (approved):

PR #438, opened 2025-04-19
Remove `Tournament.read_papi()`

Molrn:
The entire Tournament class relies on the `read_papi` function to set its data, which makes it hard to interact with data and makes the class heavily dependant from Papi. This makes it easier to evolve to our SQLite only format.  

 In addition, it allows for less useless DB queries and operations with the ability to clear the cache of a tournament, instead of having to delete the entire event (or tournament) from the cache. This has been used on pairings as an exemple: when adding a result (or doing any other operation), there is no more reading queries either to the papi DB or the SQLite DB, only the one to insert the result.  
 
 I've tested as much as I could, nothing semt impacted

### DIFF
--- a/src/data/tournament.py
+++ b/src/data/tournament.py
@@ -514,26 +514,14 @@ class Tournament:
         If *clear_papi_tournament_info* or *clear_players_by_id*,
         the values will need to be fetched again from the Papi Database."""
         cached_property_names = [
-            'boards',
-            'unpaired_players',
-            'current_round',
-            'playing',
-            'players_by_check_in_status',
-            'check_in_counts',
-            'players_by_fide_id',
-            'players_by_trf_id',
-            'players_by_name_with_unpaired',
-            'players_by_name_without_unpaired',
-            'gender_counts',
-            'federation_counts',
-            'club_counts',
-            'dependent_families',
-            'dependent_screens',
+            name
+            for name in dir(self)
+            if isinstance(getattr(type(self), name, None), cached_property)
         ]
-        if clear_players_by_id:
-            cached_property_names.append('players_by_id')
-        if clear_papi_tournament_info:
-            cached_property_names.append('papi_tournament_info')
+        if not clear_players_by_id:
+            cached_property_names.remove('players_by_id')
+        if not clear_papi_tournament_info:
+            cached_property_names.remove('papi_tournament_info')
         for property_name in cached_property_names:
             if property_name in self.__dict__:
                 del self.__dict__[property_name]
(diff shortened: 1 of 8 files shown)


src/web/controllers/admin/pairings_admin_controller.py
```diff
@@ -77,12 +76,12 @@ def __init__(
...
-            )
-            self.admin_boards, unpaired = self.admin_tournament.build_boards(
-                self.admin_round
-            )
+            for player in self.admin_tournament.players:
+                self.admin_tournament.set_player_points(
+                    player, before_round=self.admin_round
+                )
+            self.admin_boards = self.admin_tournament.build_boards(self.admin_round)
+            unpaired = self.admin_tournament.get_unpaired_players(self.admin_boards)
```
pascalaubry, line 117: I don't understand why you do not just just call the properties.

src/data/tournament.py
```diff
@@ -504,17 +505,39 @@ def playing(self) -> bool:
...
+            'check_in_counts',
+            'players_by_fide_id',
+            'players_by_trf_id',
+            'players_by_name_with_unpaired',
+            'players_by_name_without_unpaired',
+            'gender_counts',
+            'federation_counts',
+            'club_counts',
+            'dependent_families',
+            'dependent_screens',
```
pascalaubry: I feel that you clear all the cached properties but ``players_by_id`` and ``papi_tournament_info``; is there a way to look for all of them to be sure not to forget one on future evolutions?

src/data/tournament.py
```diff
@@ -1046,23 +1028,21 @@ def delete_result(self, board: Board):
...
 
     def read_player_dict(
         self,
         player_papi_id: int,
     ) -> dict[str, str | int | float | None]:
         """Reads a player from the Papi database and returns it as a dict
         (used to move players from one tournament to another one)."""
-        with PapiDatabase(self.file, write=True) as papi_database:
+        with PapiDatabase(self.file) as papi_database:
             return papi_database.read_player_dict(player_papi_id)
```
pascalaubry: This method is not used anymore, remove it if it is OK for you.

pascalaubry (dismissed):
Everything seems correct and tested successfully.
However I feel that, most of the time, reading a tournament will open the database twice (``papi_tournament_info`` and ``players_by_id``) and I wonder if not doing some computations is worth.
Have you thought of reading systematically all the needed data and do no computation on players until you need it?
This problem will disappear with the Access storage, but it may be important until then.
Let's talk about it, maybe we have to measure the time spent on some parts of the code.

src/web/controllers/admin/pairings_admin_controller.py
Molrn: the properties are bound to the current round, we need to set these for the admin round (i.e. the consulted / displayed round)

pascalaubry (approved):

PR #441, opened 2025-04-19
Release 2.5.7

pascalaubry:
Disabled buttons of the tournaments cards when the Papi file is not found
Updated a few i18n strings
Removed experimental locales from 2.5
Updated the download link and the ChangeLog
Release 2.5.7


src/common/logger.py
```diff
@@ -28,7 +28,7 @@
         },
         'standard': {
             'format': '%(asctime)s %(levelname)-10s%(message)s',
-            'datefmt': '%y/%m/%d %H:%M:%S',
+            'datefmt': '%y-%m-%d %H:%M:%S',
```
Molrn:
```suggestion
            'datefmt': '%Y-%m-%d %H:%M:%S',
```
#431 started as "it needs to have year on 4 chars instead of 2", 'Y' is needed in the format

src/common/logger.py
```diff
@@ -14,7 +14,7 @@
         'colored': {
             '()': 'colorlog.ColoredFormatter',
             'fmt': '%(log_color)s%(asctime)s %(levelname)-10s%(message)s%(reset)s',
-            'datefmt': '%y/%m/%d %H:%M:%S',
+            'datefmt': '%y-%m-%d %H:%M:%S',
```
Molrn: I don't get it, wasn't date format removed from console logs in #431 ? maybe we should agree on the format beofre propagating it on both stable and dev

Molrn (dismissed):

src/common/logger.py
pascalaubry: My fault, forgot to apply ALL the changes.

src/common/logger.py
pascalaubry: @Amaras put the date back to the logs, I will do another PR on the log format (see the last comment in #431).

pascalaubry (commented):
Should now be compliant to #431.

Molrn (approved):
LGTM

PR #442, opened 2025-04-19
Apply 2.5.7 last changes to 2.6

pascalaubry:
Updated the download link (2.5.7)
Disabled buttons of the tournaments cards when the Papi file is not found
Updated a few i18n strings
Updated the ChangeLog
Fixed installation checker lib name


Molrn (approved):
LGTM

PR #447, opened 2025-04-20
Make the app 1000 times faster

Molrn:
### TODO  

- [x]  reload a tournament when the `last_update`  field does not match
- [x] store the modified time of each papi file, and clear the tournament cache if it is different (i.e. if the user has modified the tournament in Papi)
- [x] make sure to not reload the papi_file when we the modification is made by SC
- [x] restrict the written fields of the scripts to `uploaded_at`
- [x] remove written fields from the global tournament update
- [x] set `last_update` to now when the tournament is modified from scripts
- [x] add an expire time of 30 minutes for each event

### Outside of PR scope: handle 2 servers running on different ports  
Add server config via a `sharly-chess.ini` file in which the port can be configured

### DIFF
--- a/src/data/event.py
+++ b/src/data/event.py
@@ -28,12 +28,11 @@ from data.screen import Screen
 from data.screen_set import ScreenSet
 from data.timer import Timer, TimerHour
 from data.tournament import Tournament
-from database.sqlite.event.event_database import EventDatabase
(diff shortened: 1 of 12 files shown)


timothyarmes (approved):
Amazing.

PR #450, opened 2025-04-21
Moved i18n_babel.py from /scripts to /src

pascalaubry:


scripts/i18n/i18n_translate.py
```diff
@@ -21,7 +21,7 @@
     )
 )
 
-from scripts.i18n.i18n_babel import run_babel_command
+from utils.i18n_babel import run_babel_command
```
Molrn: `Cannot find reference 'run_babel_command' in 'i18n_babel.py' `

scripts/i18n/i18n_translate.py
```diff
@@ -21,7 +21,7 @@
     )
 )
```
Molrn: you can also remove the ugly `sys.path.extend` at the path of the file now that there is no import from `scripts`.

scripts/i18n/i18n_check.py
```diff
@@ -20,7 +20,7 @@
     )
 )
```
Molrn: `sys.path.extend` to remove

scripts/i18n/i18n_check.py
```diff
@@ -20,7 +20,7 @@
...
 
-from scripts.i18n.i18n_babel import BabelWrapper
+from utils.i18n_babel import BabelWrapper
 
 from common.i18n import (  # Noqa: E402
```
Molrn: all `# Noqa E402` can also be removed

Molrn (changes requested):

timothyarmes (changes requested):
It still doesn't run when using --path:

```
  File "C:\Users\timothyarmes\Development\papi-web\src\utils\i18n_babel.py", line 58, in extract_i18n_strings
    with open(extract_config_file, 'r') as f:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'scripts\\i18n\\babel.cfg'
```

Molrn:
hm... I think the following refactor would be in order then:
- src/i18n
  - __init__.py (current `src/common/i18n.py`)
  -  babel.cfg
  - babel.py (current `src/utils/i18n_babel.py`)

timothyarmes (approved):
Working for me now

Molrn (approved):

PR #471, opened 2025-04-28
Hides empty modals

timothyarmes:
Ensures that empty modals are not displayed to the user, to avoid seeing a flash of emptiness ;)


Molrn (approved):
Works as expected, LGTM :)

PR #478, opened 2025-04-29
Pairings tests

Molrn:
Removed all the byes and forfeits, so papi not respecting the floater rule is not a problem anymore.  
- Haley is working
- Soft-Haley fails on round 5
- SAD fails on round 5
- Nicois fails on round 3


Amaras (approved):
Tests fail, but:
1. At least we have a testing infrastructure
2. we can investigate them later (I suspect it's papi not following the Swiss Dutch rules)

PR #482, opened 2025-04-30
Fix htmx deprecation

Amaras:


src/web/controllers/admin/tournament_admin_controller.py
```diff
@@ -101,7 +99,7 @@ def _admin_validate_tournament_update_data(
         rating: int | None = None
         start: float | None = None
         stop: float | None = None
-        if action == 'delete':
+        if action ==':
```
Amaras: Don't know why the merge has done that, I'll check later

Amaras (commented):
Alright, doing something else before that

Molrn (approved):
LGTM

PR #492, opened 2025-05-04
Fix resized modals preventing modal dismiss (closes #491)

### LINKED ISSUE #491: Resized modals prevent modal dismiss
For modals that are sized like this:
```html
<div style="--bs-modal-width: min(90vw, 400px);">
      [modal content]
</div>
```
Clicking outside of the modal at the same level does not dismiss the modal, clicking under the modal does.  
only the pairing modals are impacted

timothyarmes:


Molrn (changes requested):
image
pretty sure that's not intended ^^

Molrn (approved):
LGTM

PR #495, opened 2025-05-05
Hide timer once there are not more hours (closes #457)

### LINKED ISSUE #457: The timer 3 delay is ignored when there are no more hours
Once there are not more hours in the timer set, the timer should stop being displayed after #3 minutes.
Also, it would be nice to be able to set this to 0

timothyarmes:


src/web/templates/user/screen/timer.js
```diff
@@ -80,23 +82,25 @@ function update_timer() {
 		update_timer_values(clock_html, text_html, color);
 		return;
 	}
-	{% if not timer_hour.last_valid %}
 	if (time < {{ timer_hour.timestamp_next }}) { // {{ timer_hour.datetime_str_next }} color_3 {{ timer_hour.text_after }}
-	{% endif %}
+		console.log('color_3');
```
pascalaubry:
```suggestion
```

pascalaubry (changes requested):
Works well, just a log to remove.

pascalaubry (approved):

PR #496, opened 2025-05-05
Translate timer strings (closes #465)

### LINKED ISSUE #465: Localize the displayed timer
The timer text isn't localized, the text is hardcoded in French the timer.js

pascalaubry:
@timothyarmes Not nice but it works, can be improved on this PR or another one (nothing more can be done using ``_()`` in Jinja, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl is probably the solution as you suggested).


timothyarmes:
The Javascript Intl API might need a lots of polyfills depending on the client (browsers will have everything, but I'm concerned about other headless clients), and since we're having to send the JS code ourselves and can't rely on an internet connection this isn't a viable solution.

However, we still need robust plural handling for languages that have more than 2 plural forms.  I found a tiny JS library called Polyglot.js, written by AirBnB that does this.  I had to Browserify it myself since it's along supplied as an rpm library. I've included the source alongside timer.js.

Using this library simplifies timer.js and allows for many more languages in the future.

timothyarmes (approved):

PR #497, opened 2025-05-05
Updated sharly-chess.ico

pascalaubry:


timothyarmes:
We need to wait for the name change before updating the .ico, no?

pascalaubry:
> We need to wait for the name change before updating the .ico, no?

No need to wait for me.

timothyarmes (commented):
Do we need all the different sizes of each version of the icon in the repository?

pascalaubry:
> Do we need all the different sizes of each version of the icon in the repository?

Well I hate doing things twice, so if we want to change the ICO we won't have to recreate the images.
Du you prefer that I store them on our framaspace?

PR #501, opened 2025-05-06
Current round for Berger

Molrn:


docs/technical-appendices/i18n.md
Amaras, line 1: Ah, line 8 of `CODEOWNERS` is wrong.
please remove the `86-` part at the end of the line

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -2199,6 +2202,9 @@ msgstr "Homme"
 msgid "Mandatory for the first hour, optional for next hours."
 msgstr "Obligatoire pour le premier horaire, facultative pour les horaires suivants."
 
+msgid "Mark this round as over and define the next one as active."
+msgstr "Marquer cette ronde comme terminée et définir la suivante comme active."
```
```suggestion
msgid "Mark this round as over and define the next one as the current round."
msgstr "Marquer cette ronde comme terminée et définir la suivante comme la ronde courante."
```
Or at least update https://github.com/Sharly-Chess/sharly-chess/blob/dev/docs/sandbox/pairing-actions.md to state that the 'active round' and the 'current round' are the same.

locale/en/LC_MESSAGES/messages.po
```diff
@@ -2199,6 +2202,9 @@ msgstr "Male"
 msgid "Mandatory for the first hour, optional for next hours."
 msgstr ""
 
+msgid "Mark this round as over and define the next one as active."
```
```suggestion
msgid "Mark this round as over and define the next one as the current round."
```
Cf next comment on FR strings.

src/data/pairings/systems.py
```diff
@@ -181,10 +196,13 @@ def permission_handler(self) -> PermissionHandler[PairingAction]:
...
-            for round_ in range(1, tournament.rounds + 1)
+    def default_current_round(self, tournament: 'Tournament') -> int:
+        """Last round with played results."""
+        return next(
+            (
+                round_
+                for round_ in reversed(range(1, tournament.rounds + 1))
+                if tournament.round_has_played_result(round_)
+            ),
+            0,
```
pascalaubry, line 207: If the current round is the one displayed on user screens, then shouldn't it be the last round with no results or at least one missing result (or the current round will not change after the last result of a round has been entered)?

src/data/tournament.py
```diff
@@ -397,18 +396,24 @@ def arbiter(self) -> str:
...
     def players_by_id(self) -> dict[int, Player]:
         _, players_by_id = self.read_papi()
-        players = players_by_id.values()
-        current_round = self._get_current_round(players)
-        illegal_moves: Counter[int] = self.get_illegal_moves(current_round)
-        for player in players:
+        # The computation of the property `current_round` needs to access a players' iterable.
+        # Calling `self.players` in the context of this function creates a circular dependency
+        # as it itself calls `self.players_by_id`.
+        # To avoid this dependency, `self._players` is temporarily allocated.
```
pascalaubry, line 402: The introduction of _players is understandable with your comment but still complex (no better proposal but I wonder if it does not reveal a conception issue).

src/web/templates/admin/pairings/round_robin_pairing_buttons.html
```diff
@@ -24,4 +24,18 @@
             </button>
         </div>
     {% endif %}
+
+    {% if round_status == 'CURRENT' and not admin_tournament.playing and not admin_tournament.finished %}
+        <div {{ macros.tooltip_attributes('info', _('Mark this round as over and define the next one as active.'), 'top') }}>
```
pascalaubry:
```suggestion
        <div {{ macros.tooltip_attributes('info', _('Mark this round as over and define the next one as the current round.'), 'top') }}>
```

pascalaubry (commented):
Did not test, only comments (I let @timothyarmes who already started to test).

src/data/pairings/systems.py
Amaras: > If the current round is the one displayed on user screens, then shouldn't it be the last round with no results or at least one missing result (or the current round will not change after the last result of a round has been entered)?

The last round without results will unconditionally be the last round.
It should either be the last round with played results, or the first round with at least one missing result.

Amaras (dismissed):

docs/technical-appendices/i18n.md
Amaras: NOTE: I wasn't supposed to be code owner for this file, so that's a syntax mistake on my part

src/data/tournament.py
Molrn: It is, `read_papi` had a lot of underlying issues. That's the best I could come up with to:
- not propagate the misconception to further function
- have it clearly documented
- not modify usages of the `Tournament` class, which would cause a great risk of breaking something

This is bound to evolve once we get rid of Papi, now the part to evolve is much more restricted.

src/data/pairings/systems.py
Molrn: > It should either be the last round with played results, or the first round with at least one missing result.  

Problem with `last round with played results`: if you enter a played result rounds in advance it becomes your current round.  
Problem with `first round with missing results`: If you remove a result from rounds ago your current round changes.  

I'm ok with both, I have a slight preference with the first on because it is less likely to happen.  
This is counterbalanced by the manual selection of the round done by clicking `End round` (this is only the default current_round when no manual `current_round`) has been selected  

Updating `pairing-actions.md` with this usage feels relevant, doing so

locale/fr/LC_MESSAGES/messages.po
Molrn: I feel that `active` is more user-friendly and representative of what it is (a round can be active without being `en cours`), current feels more dev-oriented. I'm up to add active to the doc

Molrn: discussed with tim, we both agree that `current` feels right in english, I'm more under the impression that it is some kind of `faux-ami` in that context, and I feel nobody actually would say `ronde courante`. I'm putting it back to `ronde courante / current_round` for consistency

timothyarmes (approved):

PR #506, opened 2025-05-08
2.5 Fix ChessEvent players' title and rating types

pascalaubry:


Molrn (changes requested):
Same as #507, I think the utils is not on `2.5-stable` so I think you'd need to add it

pascalaubry:
Back to draft because I found this morning a similar problem on the rating types.

Molrn (approved):
All good for me, @pascalaubry letting you merge and release

pascalaubry (commented):
Cleaner this way indeed, works fine with your last changes.

PR #507, opened 2025-05-08
Fixed ChessEvent players' title and rating types

pascalaubry:


src/plugins/chessevent/data/chessevent_player.py
```diff
@@ -1,19 +1,77 @@
...
     PlayerRatingType,
-    PlayerTitle,
 )
 from plugins.chessevent.data.chessevent_field_reader import ChessEventFieldReader
 from plugins.ffe.utils import PlayerFFELicence
 
 logger: Logger = get_logger()
 
 
+class ChessEventPlayerTitle(IntEnum):
```
Molrn: Il faudrait vraiment éviter autant que possible de recopier des classes. J'ai rajouté l'utils `PluginCoreMapper` pour ça, il faudrait faire pareil (comme ce qui a été fait pour `PapiPairingVariation`)

Molrn (changes requested):

src/plugins/chessevent/data/chessevent_player.py
pascalaubry: Classes are not copied, the numbers are different.
Back to draft because I found this morning a similar problem on the rating types.

Molrn: with the mapper, you can directly map the chessevent number to the core object, so it does not matter if the numbers don't match. ex with `PapiPairingSystem`:
```py
class PapiPairingSystem(PluginCoreMapper[str, PairingSystem]):
    @classmethod
    def _core_object_by_plugin_value(cls) -> dict[str, PairingSystem]:
        return {
            'Suisse': systems.SwissPairingSystem(),
            'ToutesRondes': systems.RoundRobinPairingSystem(),
        }
```

Molrn (approved):
LGTM

PR #510, opened 2025-05-09
2.5 Fixed bug on tournament creation

pascalaubry:
Same as #508 for the dev branch
Moved blur to input
Added debounce
Release 2.5.10

### DIFF
--- a/docs/03-change-log.md
+++ b/docs/03-change-log.md
@@ -2,11 +2,11 @@
 
 # Papi-web - ChangeLog
 
-## Version 2.5.10 - 9 mai 2025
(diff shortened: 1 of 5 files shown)


Molrn:
That's an assertion error, it is not present on the prod env. We can fix it if we want, but I'm a bit opposed to doing a release specifically for that, especially when we've already done a release for something else the same day

pascalaubry:
This is obviously more than an issue than happens in DEVEL_ENV only, something fails event with the changes I did in #508.
Back to draft.

Molrn (approved):
LGTM

PR #512, opened 2025-05-10
2.5 Really force users to edit the configuration when force_edit is True

pascalaubry:
Still draft, missing something to prevent users from closing the config modal when force_edit is True.


pascalaubry:
Works well on 2.5 too.

pascalaubry:
2.6.0 will be there before 2.5.10, closing this PR.

PR #513, opened 2025-05-10
Really force users to edit the configuration when force_edit is True

pascalaubry:
Still draft, missing something to prevent users from closing the config modal when force_edit is True.


timothyarmes (dismissed):
LGTM

pascalaubry:
Works fine for me.

timothyarmes (approved):

PR #520, opened 2025-05-10
Pairings UI improvements

Molrn:


Amaras (dismissed):
Summarizing my message on Discord:
General usability problems (as per my message on Discord): settings should not be valid when re-pairing (so must be invalidated on unpairing).

Also, the pairings are wrong, somehow.

Amaras (dismissed):
All good now!

Amaras (approved):

PR #540, opened 2025-05-12
Open screens from admin in separate tab

timothyarmes:


src/web/templates/admin/screens/screen_card.html
Amaras, line 1: huh, why remove the preload?

timothyarmes: There's no HTMX left, I now open it in another tab

Amaras: My bad, I thought `a` had a `preload` attribute

Amaras (approved):

PR #544, opened 2025-05-12
Fixed error messages

pascalaubry:


Amaras:
This should not be on your branch, it should be solved on `dev`

PR #553, opened 2025-05-13
Improve pairings UI for swiss tournaments

### LINKED ISSUE #547: Fix the pairing settings dialog
The following improvements of the pairings settings dialog would be nice:
- the initial values of the rating limits should be the recommended values.
- ~~the text of the 'Save' button should be 'Pair' (as now we do not only save the settings, we also generate the pairings).~~
Image

### LINKED ISSUE #552: [BUG] Fix the pairings display when using progressive acceleration
## Bug description
Steps to reproduce the bug:
1. Create a Swiss tournament with a progressive acceleration
2. Add players
3. Pair the first round
4. See the displays

On the admin pages, only the virtual points are displayed.

Image

The virtual points are correct:

Image

On the print view, only the real points are displayed.

Image

The Papi export is correct:

Image

## Environment
- Sharly Chess version: 2.6.0

## Additional context

This also affects other accelerations.

Molrn:
closes #547 closes #552 
- reverse upper / lower limit input fields
- for rating limits, fill recommanded values instead of 0
- pair round 1 from the settings modal
- improve recommanded rating limits for SAD
- display settings modal when pairing round 1
- show real points / vpoints in print view / pairing tab

### DIFF
--- a/src/plugins/pairing_acceleration/pairing_settings.py
+++ b/src/plugins/pairing_acceleration/pairing_settings.py
@@ -188,8 +188,8 @@ class DualRatingLimitsSetting(PairingSetting[tuple[int, int]]):
         player_count = len(ratings)
         if player_count < 3:
             return 0, 0
-        if player_count < 7:
-            # Min ideal repartition: A(4), B(2), C(1)
+        if player_count < 11:
+            # Min ideal repartition: A(4), B(4), C(3)
             first_c = player_count // 3 - 1
             first_b = 2 * player_count // 3 - 1
         else:
--- a/src/plugins/pairing_acceleration/pairing_settings.py
+++ b/src/plugins/pairing_acceleration/pairing_settings.py
@@ -54,8 +54,9 @@ class RatingLimitSetting(PairingSetting[int]):
 
     @classmethod
     def default_value(cls, tournament: 'Tournament') -> int:
-        # TODO replace by recommended value once papi dependency has been removed
-        return tournament.papi_tournament_info.rating_limit1
+        if papi_limit := tournament.papi_tournament_info.rating_limit1:
+            return papi_limit
+        return cls.recommended_value(tournament)
 
     @classmethod
     def recommended_value(cls, tournament: 'Tournament') -> int:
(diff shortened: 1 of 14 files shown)


src/plugins/pairing_acceleration/pairing_settings.py
```diff
@@ -172,16 +173,30 @@ def get_value(cls, tournament: 'Tournament') -> tuple[int, int]:
...
-        if len(ratings) < 3:
+        player_count = len(ratings)
+        if player_count < 3:
             return 0, 0
-        first_c = len(ratings) // 3 - 1
-        first_b = 2 * len(ratings) // 3 - 1
+        if player_count < 7:
+            # Min ideal repartition: A(4), B(2), C(1)
+            first_c = player_count // 3 - 1
+            first_b = 2 * player_count // 3 - 1
```
Amaras: All groups must have at least 25% and at most 50% of the players.
4-2-1 does not leave enough players for group C (1/7 < 1/4)

Amaras (dismissed):
Just one question: do you validate the values are correct before pairing (i.e. each pairing group has at least 25% and at most 50% of the players)?

src/plugins/pairing_acceleration/pairing_settings.py
Molrn: > do you validate the values are correct before pairing  

image
it it validated in the form, so you can't pair round 1 if they are incorrect. For other rounds, if limits are invalid, the modal is opened again.  

Right ^^' modified the threshold to 11 (4-4-3), everything above that appears to be correct

Amaras (commented):
Didn't test, looks good

pascalaubry:
I am testing with the last changes.

pascalaubry (changes requested):
Everything OK for me (all changes tested), just ont detail should be changed:
when the tournament is started, the user should not be able to change the pairing system at all. We should disable the fields on the form and the controller should not handle the values passed (if any).

image

(not part of your PR, just saw that).

pascalaubry (approved):
As discussed, my comment can wait for the totatle removal of the Access databases.
Creating an issue.

PR #560, opened 2025-05-13
2.6 Fixed a string on family cards

pascalaubry:


locale/fr/LC_MESSAGES/messages.po
```diff
@@ -2820,8 +2820,8 @@ msgid "Players from #%(first)d to #%(last)d"
 msgstr "Joueur·euses de %(first)d à %(last)d"
 
 #, python-format
-msgid "Players from #%(first)s to #%(last)s"
-msgstr "joueur·euses du n°%(first)s au n°%(last)s"
+msgid "Players from %(first)s to %(last)s"
+msgstr "joueur·euses de %(first)s à n°%(last)s"
```
Molrn, line 2824:
```suggestion
msgstr "joueur·euses de %(first)s à %(last)s"
```
same as #559, I suggest we only duplicate a PR once it has been approved (or merged) to avoid duplicating work. In the mean time, adding `To duplicate on dev` or `To duplicate on stable-2.6` should do the trick

Molrn (changes requested):

PR #563, opened 2025-05-14
Families fix

timothyarmes:


src/data/tournament.py
```diff
@@ -632,6 +632,8 @@ def clear_cache(self, clear_papi_cache: bool = False):
         self._players_by_rank = None
         self.event.clear_screen_cache(self.id)
```
pascalaubry: I wonder if we should not simplify this way:
```suggestion
        for screen in self.dependent_screens:
            screen.clear_cache()
```

pascalaubry (approved):
Nice to have changed also the links of the rotators and families cards.
Works fine.

PR #568, opened 2025-05-14
Fix GitHub Actions export workflow

Amaras:
As we noticed during the publishing of one of our pre-releases, the pre-release check was not correct.

This was due to the use of the `contains` function incorrectly:
I tried to use it as `contains(needle, haystack)`, but the syntax is `contains(haystack, needle)`.

Also made sure to use the correct tag for `actions/checkout` (the tag is `vx.y.z`, not `x.y.z`).<


pascalaubry (approved):
Can not test but seems correct.

PR #573, opened 2025-05-15
Berger grid print document

Molrn:


src/utils/__init__.py
```diff
@@ -78,6 +78,14 @@ def round_ranking(num: float | Decimal) -> int:
             return lowest_int + 1
         return lowest_int
 
+    @staticmethod
+    def points_str(points: float | None) -> str:
+        if points is None:
+            return ''
+        if points == 0.5:
+            return '½'
+        return f'{points:.1f}'.replace('.0', '').replace('.5', '½')
```
Amaras: Please add 1/4 and 3/4 cases for Sonneborn-Berger.

src/data/print_documents/documents.py
```diff
@@ -296,3 +308,122 @@ def static_id() -> str:
...
+        assert self.tournament is not None
+        pairing_engine = self.tournament.pairing_variation.engine
+        assert isinstance(pairing_engine, RoundRobinPairingEngine)
+        result_grids: list[dict[int, list[Result | None]]] = [
+            {
+                player.id: [None] * self.grid_length
+                for player in sorted(
+                    self.tournament.players,
+                    key=lambda p: self.berger_nb_by_player_id[p.id],
+                )
```
Amaras: Shouldn't the arbiter be able to define the order?
Most likely, the first grid should be either by rating, by initial number, or by name, but the final grid should definitely be by result

Amaras (dismissed):
N° | Nom | Féd. | Elo | 1 | 2 | 3 | 4 | 5 | 6 | Pts
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | AUBRY Pascal | FRA | 1673 F | X |   |   |   |   | < | 1
2 | ARMES Timothy | FRA | 1608 F |   | X |   |   |   | < | 1
3 | AUBRY Youri | FRA | 1941 F |   |   | X |   |   | < | 1
4 | LAPEYRE Pierre | FRA | 2006 F |   |   |   | X |   | < | 1
5 | PLAT Sammy | FRA | 1556 F |   |   |   |   | X | < | 1
6 | Exempt·e |   |   | > | > | > | > | > | X |  

The PAB should never be shown, nor taken into account in Round-Robin results.
It is only useful in Swiss tournaments

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3985,6 +3994,12 @@ msgstr "Ce pilote d'afficheurs est privé (non visible des utilisateur·ices)."
 msgid "This display controller is public (visible by the users)."
 msgstr "Ce pilote d'afficheurs est public (visible des utilisateur·ices)."
 
+msgid "This document is not available for unpaired tournaments."
+msgstr "Ce document n'est pas disponible pour les tournois non appariés."
```
```suggestion
msgstr "Ce document n'est pas disponible pour les tournois non-appariés."
```

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3985,6 +3994,12 @@ msgstr "Ce pilote d'afficheurs est privé (non visible des utilisateur·ices)."
...
+
+msgid "This document is only available for Round-Robin tournaments."
+msgstr "Ce document est seulement disponible pour les tournois Toutes-rondes."
```
pascalaubry: In other strings, 'toutes rondes' is used.
My preferred is 'toutes rondes' but both are acceptable (we should always use the same form).

pascalaubry (commented):
Nice, just a few remarks on strings.

Molrn:
> The PAB should never be shown, nor taken into account in Round-Robin results.  

So far, we have PAB in the pairings, which means we have PAB in the results. should it be otherwise, with the player not being paired at all ? that is Papi's solution, but we'd have to do a lot more than just keeping it out of the Berger grid doc.  

In order to make RR work without PAB, I suggest the following: having an option for RR tournaments that adds unpaired players to the board lists, and have it in all the pairings representations (user, print and pairings tab) as `player  -  UNPAIRED` instead of `Player  1 - F  EXEMPT`

src/data/print_documents/documents.py
Molrn: So far I've put it with no option, but we could definitely add some. I think "rounds" does not feel appropriate, but the sorting option you described definitely is. 

any other relevant options / missing columns? I for example thought of adding tie-breaks, not sure it is relevant tho (could be a checkbox option `Display tie-breaks`, or could also be displayed only when you sort the players by rank)

Amaras: Tie breaks are relevant when you want to have the crosstable.

Molrn:
PAB removed from the document, doing another PR to remove PAB from berger tournaments in general

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -4791,3 +4815,9 @@ msgstr "ecran-{screen_type}"
 msgid "{string}: {value}"
 msgstr "{string} : {value}"
 
+#~ msgid "First encounter"
+#~ msgstr "Première rencontre"
+
+#~ msgid "Second encounter"
+#~ msgstr "Deuxième rencontre"
+
```
```suggestion
```

locale/en/LC_MESSAGES/messages.po
```diff
@@ -4791,3 +4815,9 @@ msgstr ""
 msgid "{string}: {value}"
 msgstr ""
 
+#~ msgid "First encounter"
+#~ msgstr ""
+
+#~ msgid "Second encounter"
+#~ msgstr ""
+
```
```suggestion
```

pascalaubry (commented):
Just comments, could not test.

timothyarmes (approved):

PR #584, opened 2025-05-18
Fix alpha screens accessing unset player boards

Molrn:
Boards need to be built when setting the players in order for all the board values of the Player class to be set


pascalaubry (approved):
Fixes the cache issue on boards screns, thanks.
Note: the bug on rotators still remains.

PR #588, opened 2025-05-19
Optimized i18n updates on engine startup (developers only)

pascalaubry:
Also did some changes on export.py.


Molrn (commented):
This optimizes the time, but not the action. Would it be possible to only update the .mo file and not the .po ? the .po update does not change anything about the in-app translations (untranslated strings are put as empty), and when no string is modified it creates a git stash because the file's modified time was updated.

scripts/i18n/i18n_update.py
Molrn, line 1: This file does nothing

src/common/i18n/babel_checker.py
```diff
@@ -0,0 +1,163 @@
...
+                else:
+                    translator_strings.append(translator['name'] or '')
+            line += f'| {"<br/>".join(translator_strings)} |\n'
+            lines.append(line)
+        lines.append(
+            f'Last update: {datetime.strftime(datetime.fromtimestamp(time.time()), "%Y-%m-%d %H:%M")}\n'
+        )
+        with open(doc_file, 'w', encoding='utf-8') as f:
+            for line in lines_before_comment + lines + lines_after_comment:
+                f.write(line)
```
Molrn: This is adding another project modified file, but this time it gets modified all the time, not only if translations are not correct. See my main message as for why I think this is a bad idea.

Molrn (dismissed):
@pascalaubry I feel like we don't get each other on this PR, Imma try to clarify.  

It is considered a bad practice to modify project files when running a project. It complexifies testing and branch navigation. Automatically modifying project files should only be done with commands that explicitely do so (ex: ruff format, i18n_update, etc...).  

As I understand it, there are 2 problems that this process is trying to solve:
- make sure you get the translations of the `.po` files when you run the server
- avoid getting PRs with untranslated strings  

To fullfill these I suggest the following:
- When starting the server
  - compile the `.mo` files from the `.po` files if necessary (makes sure you have all the current translations when running the app)
  - check if `.po` files are correct, if they are not warn the dev with a message asking to run `i18n_update` (or equivalent)
- i18n_update: updates the `.po` files from the translations in the project
- prevent merge of PRs with invalid `.po` files (2 methods)
  - add a check to the pre-commit hook, like what we have for ruff
  - add it to the CI GitHub pipeline, such that PRs could not be merged if it fails

pascalaubry:
> @pascalaubry I feel like we don't get each other on this PR, Imma try to clarify.
> 
> It is considered a bad practice to modify project files when running a project. It complexifies testing and branch navigation. Automatically modifying project files should only be done with commands that explicitely do so (ex: ruff format, i18n_update, etc...).

OK, let's call that bad practice but I feel much more unconfortable with not up-to-date translations.

> As I understand it, there are 2 problems that this process is trying to solve:
> 
> * make sure you get the translations of the `.po` files when you run the server

Yes.

> * avoid getting PRs with untranslated strings

No really, since in a future when we should have several translators, there may be some unfilled translations if some translators are not available at the time we publish new realeases.
Precisely, the second goal is running each time with up-to-date translations without asking myself 'should I run i18n_update?'

> To fullfill these I suggest the following:
> 
> * When starting the server
>   
>   * compile the `.mo` files from the `.po` files if necessary (makes sure you have all the current translations when running the app)
>   * check if `.po` files are correct, if they are not warn the dev with a message asking to run `i18n_update` (or equivalent)
> * i18n_update: updates the `.po` files from the translations in the project
> * prevent merge of PRs with invalid `.po` files (2 methods)
>   
>   * add a check to the pre-commit hook, like what we have for ruff
>   * add it to the CI GitHub pipeline, such that PRs could not be merged if it fails

Doing it in a pre-commit may be a concern but it does not answer my need (having up-to-date translations).

I propose to get the actual behaviour only when a file ``src/common/i18n/.auto-update`` is found, and let you explicitely run ``i18n_update`` when not found (and add that to the dev tips and tricks).

scripts/i18n/i18n_update.py
pascalaubry: Give it a try and you will see that it does the job (when importing ``common.i18n`` the code in ``__init.py__`` is run).

Molrn:
> I propose to get the actual behaviour only when a file src/common/i18n/.auto-update is found, and let you explicitely run i18n_update when not found (and add that to the dev tips and tricks).  

Fine by me  

> it does not answer my need (having up-to-date translations)  

I don't get how `when running the compile the .mo files from the .po files if necessary` does not answer this need. The only addition of the current solution is that the untranslated strings are created as empty in the `.po` file, which does not have an impact on the server (empty string in the .po or not the string appears as the msgid).  

Pre-commit / CI pipeline were only for the objective `avoid getting PRs with untranslated strings`, it seems to not be that important of an objective so it's definitely not a necessity

pascalaubry:
> I don't get how `when running the compile the .mo files from the .po files if necessary` does not answer this need. The only addition of the current solution is that the untranslated strings are created as empty in the `.po` file, which does not have an impact on the server (empty string in the .po or not the string appears as the msgid).

We do much more than creating empty strings: we control if there are empty strings, if the tokens in the original strings are present in the translated strings, the presence of some special tags, ...

locale/en/LC_MESSAGES/messages.po
Molrn, line 1: all the strings from #605 were removed

scripts/i18n/i18n_update.py
Molrn, line 1: - Updating only the .mo files is useless in a command
- no command should be triggered only by an import from the project, let alone one that has nothing to do with what actually happens  
- this does not update the .mo files if there is no `.auto-update`

Molrn (dismissed):
If you think this is what I was talking about I get that you'd find it not relevant. The actions I described:
- when running the server, compile the .mo files from the .po files
- when running `i18n_update.py`, update the .po files from the translation strings in the project, with the logs checking the number of empty strings

What is implemented now (without `.auto-update` file):
- when running the server nothing
- when running i18n_update, update of .mo files (which does not work)

In this state the project is not usable without a `.auto-update` file (if you don't already have `.mo` files there is no way to create one (so the app does not start), and no way to update .po files). I you fully disagree and don't want to implement the not automatic `.po` file update, at least remove the `.auto-update` block and the `i18n_update` scripts which have no value currently. 

I could take care of this when I'm finished with prizes if you'd like

locale/en/LC_MESSAGES/messages.po
pascalaubry: Did not see that, I feel all the strings are present and correct.

scripts/i18n/i18n_update.py
pascalaubry: Now,
1. when running the server:
- if you have ``.auto-update``: all the i18n files (POT, PO and MO) are rebuilt, if needed only.
- if not, MO files are rebuilt from the PO files
2. when running i18n_update or export:
all the i18n files (POT, PO and MO) are rebuilt, if needed only (missing files are created).

scripts/export/export.py
```diff
@@ -181,22 +176,21 @@ def create_project(silent: bool = False):
     # just create an empty custom dir (dev custom files are embedded in the exe since 2.4.11)
     custom_dir: Path = PROJECT_DIR / 'custom'
     custom_dir.mkdir(exist_ok=True)
-    sharly_chess_copyright: str = SharlyChessConfig(silent).copyright
     target_file = tools_dir / 'chessevent.bat'
-    print_interactive_info(f'Creating batch file {target_file}...')
+    logger.info('Creating batch file [%s]]...', target_file)
```
Amaras, line 180:
```suggestion
    logger.info('Creating batch file [%s]...', target_file)
```

Amaras (dismissed):

pascalaubry (commented):
Applied changes.

Molrn (approved):
LGTM

PR #592, opened 2025-05-22
Prevent full Event loading during event selection

Molrn:


pascalaubry (approved):
Works fine, really nothing else to say.

PR #599, opened 2025-05-24
Fix error return in tournament creation

timothyarmes:


Molrn (changes requested):
Event creation from the UI works well, it though is still broken for example database creation (found out about it while testing this PR ^^'):

```
  File "C:\papi-web\src\plugins\ffe\ffe.py", line 598, in event_data_for_db_write
    'ffe_auto_upload': int(self.get_data(td, 'ffe_auto_upload')),
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: int() argument must be a string, a bytes-like object or a real number, not 'NoneType'
```

Molrn (changes requested):
Event creation still does not work when chessevent is disabled:  
```
  File "C:\papi-web\src\plugins\ffe\templates\ffe_event_info_rows.html", line 3, in top-level template code
    <td class="{% if not admin_event.plugin_data.chessevent.chessevent_user_id or not admin_event.plugin_data.chessevent.chessevent_event_id %}fst-italic{% endif %}">
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\papi-web\venv\Lib\site-packages\jinja2\environment.py", line 490, in getattr
    return getattr(obj, attribute)
           ^^^^^^^^^^^^^^^^^^^^^^^
jinja2.exceptions.UndefinedError: 'dict object' has no attribute 'chessevent'
```

Molrn (approved):
LGTM

PR #601, opened 2025-05-24
Fix fixed table update

Molrn:


Amaras:
Are we doing a 2.6.4 bugfix with this, or should we wait until 2.7?
In the first case, you should start a `hotfix/2.6.4` branch from `2.6.3` and point your PR to there.

pascalaubry (approved):

Molrn:
replaced by #602

PR #611, opened 2025-05-27
Add refresh message when checkins change on pairings tab

timothyarmes:


pascalaubry (approved):
Nice, was really missing.

PR #621, opened 2025-05-28
Shutdown

timothyarmes:


pascalaubry (approved):
Works nice with the EXE:
image
Still an exception in PyCharm but for developers only.
image

PR #632, opened 2025-05-29
Fix ODS Export + add fields

timothyarmes:


pascalaubry (approved):
I updated the release notes and the doc.

PR #642, opened 2025-06-01
Fixed the screen set creation

pascalaubry:


Amaras (dismissed):
Feels weird to duplicate code for no real reason

pascalaubry:
> Feels weird to duplicate code for no real reason

Which code?
``stored_screen = event_database.add_stored_screen(stored_screen)``?

Amaras:
> > Feels weird to duplicate code for no real reason
> 
> Which code? `stored_screen = event_database.add_stored_screen(stored_screen)`?

yeah, that.
Why don't you just copy the value unconditionally?
Either it's `None` and it's just a waste of value copy, or it's not `None`, and it's deeply necessary.

My version lacks comments, but I feel that it's not too bad.

Amaras (changes requested):
Doesn't feel better.
You're still conditionally copying it.

pascalaubry:
> Doesn't feel better. You're still conditionally copying it.

My concern is having the code understandable (obviously it was not).
OK to save/restore unconditionally but please add the needed comments to say that it is not needed for screens with no screen sets (I let you do it).

PR #647, opened 2025-06-01
Optimize i18n_update

pascalaubry:
Replaces #635


docs/technical-appendices/i18n.md
Molrn, line 1: really confused at why the table was not updated here, led me to think that removing the update of the table content was the purpose of the PR ^^'

Molrn (dismissed):
PR looks good and works well, just needs to run i18n_update after rebasing to apply the signature to dev

Molrn (approved):

PR #650, opened 2025-06-01
Hotfix 2.7.3

pascalaubry:


Amaras:
Are you sure the branch is right?
Also: **don't create a PR from `hotfix/*` branches**, it's a pain with conflicts (like we have now)

PR #651, opened 2025-06-01
Fixed the display of non-FIDE players when updating the players from FFE or FIDE databases

pascalaubry:
image
image


Amaras (approved):
Alright, why not?

PR #655, opened 2025-06-02
Added plugins.chessevent.TMP_DIR

pascalaubry:
Not really a bugfix but groups all the tmp files of the ChessEvent plugin to the same location (as already done for plugin FFE in #654).

### DIFF
--- a/src/plugins/chessevent/engine/action_selector.py
+++ b/src/plugins/chessevent/engine/action_selector.py
@@ -34,7 +34,8 @@ from database.access.papi.papi_database import (
 )
 from database.access.papi.papi_template import create_empty_papi_database
 from database.sqlite.event.event_database import EventDatabase
-from plugins.chessevent import PLUGIN_NAME, TMP_DIR
+from plugins import chessevent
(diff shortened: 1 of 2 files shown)


src/plugins/chessevent/engine/action_selector.py
```diff
@@ -165,9 +164,7 @@ def write_chessevent_info_to_database(
         # write data to a temporary file to limit the time no tournament file is available
         date: str = datetime.fromtimestamp(time.time()).strftime('%Y%m%d%H%M%S')
         tmp_file: Path = (
-            TMP_DIR
-            / 'chessevent'
-            / f'{tournament.file.stem}-{date}{tournament.file.suffix}'
+            TMP_DIR / f'{tournament.file.stem}-{date}{tournament.file.suffix}'
```
Amaras: I don't really like names that could potentially clash with core modules' members.
If you're calling them, please import the module and use the qualified name (in this case: `chessevent.TMP_DIR`)

Amaras (approved):

PR #658, opened 2025-06-03
Fix FFE auth test on the tournament editing form

pascalaubry:
Bug introduced in 2.7.3

### DIFF
--- a/src/plugins/ffe/ffe_session.py
+++ b/src/plugins/ffe/ffe_session.py
@@ -276,10 +276,9 @@ class FFESession(Session):
         logger.info('Testing FFE authentication for tournament [%d]...', ffe_id)
         if not self._ffe_init():
             return None
-        if not self._ffe_auth(ffe_id, ffe_password):
-            return False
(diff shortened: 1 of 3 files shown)


src/plugins/ffe/ffe_session.py
```diff
@@ -267,9 +269,7 @@ def _ffe_auth(self, ffe_id: int, ffe_password: str) -> bool:
         return True
 
     def test_auth(self, ffe_id: int, ffe_password: str) -> bool:
-        """Tries to authenticate on the FFE admin website for the tournament.
-        Returns True on success, False if the credentials are incorrect, or
-        None if they couldn't be tested"""
+        """Tries to authenticate on the FFE admin website for the tournament"""
 
         logger.info('Testing FFE authentication for tournament [%d]...', ffe_id)
         if not self._ffe_init():
```
timothyarmes, line 277: It should also return None here if _ffe_init fails

timothyarmes (dismissed):

timothyarmes (approved):

PR #688, opened 2025-06-09
Improved exception handling (removed the deep white screen)

pascalaubry:
image


src/web/templates/errors/restart.html
```diff
@@ -0,0 +1,5 @@
+<p>
+    <a href="#" onclick="location.reload(); return false;">{{ _('Reload the page') }}</a>
```
Molrn: This reloads the error page and not the one that triggered the error, that's not a behavior we should have

src/web/templates/errors/restart.html
```diff
@@ -0,0 +1,5 @@
...
+    -
+    <a href="#" onclick="location.assign('/'); return false;">{{ _('Restart the application') }}</a>
```
Molrn: The makes it look like the server is stopped and restarted. I think something like `Return to the home page` would bemore appropriate

Molrn (dismissed):

Molrn (dismissed):
@pascalaubry good for me, letting you have a look and merge if it suits you

Molrn (approved):
LGTM

PR #691, opened 2025-06-10
Prize calculations

timothyarmes:


tests/papi/empty.papi
Molrn, line 1: not used

example_events/yml/test-event.yml
Molrn, line 1: not used

src/data/prize/prize_group.py
```diff
@@ -105,3 +121,239 @@ def reorder_categories(self, sorted_category_ids: list[int] | None = None):
                     category.stored_prize_category.index = index
                     database.update_stored_prize_category_index(category.id, index)
             database.commit()
+
+    # ------------------------------------------------ ---------------------------------
```
Molrn:
```suggestion
    # ---------------------------------------------------------------------------------
```

tests/test_prizes.py
```diff
@@ -0,0 +1,667 @@
...
+        self.assert_has_no_prize(p3, prizes)
+
+        prizes = self.assign_prizes([top_women_category, elo_category], players)
+
+        # Elo attributed to first woman,
+        self.assert_has_prize(p1, first_woman, prizes)
+        self.assert_has_prize(p2, first_1600, prizes)
+        self.assert_has_no_prize(p3, prizes)
+
+    def test_promotion_waring(self):
```
Molrn:
```suggestion
    def test_promotion_warning(self):
```

Molrn (commented):
Code is working and well tested. the `assign_prizes` function is not optimized and its complexity could be reduced, but I think it's ok to merge it as is. I've tried, doing changes (even minimal ones) while keeping the tests working is not trivial, it could deserve it's own PR. 

I don't feel confident enough on the prize rules in general to decide if the tests are correct / enough to represent the rules, but if they are, IMO we should merge (it is also the base of 2 PRs I've got incoming, including #698). Leaving @Amaras approve this as you've mostly discussed this the 2 of you

Molrn (changes requested):
missing tests / features:
- Monetary prizes should have priority over non-monetary
- unassigned prizes should be returned with assiged_to == None
- shared prizes should not go under `prize_category.sharing_threshold` (you can define it as a constant while I implement it)

I went over most of your discussions with sammy, everything else seems implemented. With these 3 working, I'll approve it ;)

Molrn (approved):
LGTM

PR #701, opened 2025-06-12
Prize category sorting

timothyarmes:


src/data/prize/prize_group.py
```diff
@@ -53,6 +53,7 @@ def sorted_categories(self) -> list[PrizeCategory]:
             self.categories,
             key=lambda category: (
                 category.is_main,
+                -category.index,
                 category.total_prize_value,
                 category.name,
             ),
```
Molrn:
```suggestion
            key=lambda category: category.index,
```
without the `reverse=True`

Molrn (changes requested):

Molrn (approved):
LGTM

PR #703, opened 2025-06-12
Add option to deploy the server to the local network

Molrn:


pascalaubry:
I think that letting clients access the server should be managed by a more general authentication and authorization policy (cf https://github.com/Sharly-Chess/sharly-chess/blob/dev/docs/sandbox/authentication-and-roles.md).
I have a local branch on this, please wait for it before merging.

Molrn:
@pascalaubry I agree we should have roles to determine the accesses, but that does not mean the server should be deployed by default. The two are not related for me. Waiting to see your PR to discuss this more in detail ;)

Molrn:
replaced by #715

PR #705, opened 2025-06-13
Fix SQLite foreing keys pragma not enabled

Molrn:
[SQLite FK doc](https://sqlite.org/foreignkeys.html#:~:text=comparison%20is%20performed.-,2.%20Enabling%20Foreign%20Key%20Support,-In%20order%20to)


Molrn:
closed because of #707

PR #708, opened 2025-06-13
Fix bool or None form usage

Molrn:


pascalaubry (approved):
Other fields will need to use ``form_data_to_bool_or_none`` instead of ``form_data_to_bool_or_none`` but it is OK for the PR.

pascalaubry:
I need the PR for other fields on another PR, merging.

Issue #707, opened 2025-06-13
Restore FK constraints


Molrn:
The migration `m010_refactor_chessevent` broke the FK constraints, making the DB unusable with the `foreign_keys` pragma enabled. Some constraints still point to the table `tournament_copy`.  
Ex: `pragma foreign_key_list(screen_set)` --> 
![image](https://github.com/user-attachments/assets/45983025-ad42-4ace-bdae-938f1c526c12).  

I tunneled into finding a way to fix that, I could not get anything to work.  
It is something very important to have working IMO, it feels very unprofessional to not have it and could lead to a lot of issues in the future (or even currently some that we haven't noticed).  
I only see 2 options to get rid of it:  
- step 1 (both options): squash the migrations past `m010` to have a clean DB scheme for new created events
- step 2:
  - easy but painful: reset the upgrade, meaning that DB with a lower version could not be upgraded to version X
  - Hard / maybe not possble: during the migration process, have one migration that creates a new DB from scratch, then copies the data from one DB to the other.

Amaras:
This answer could do the trick, but I have not tested it: https://stackoverflow.com/a/76526551/5683504

In short:
```sql
ALTER TABLE some RENAME TO some_old;
CREATE TABLE IF NOT EXISTS some (.....)
INSERT INTO some SELECT * FROM some_old;
DROP TABLE IF EXISTS some_old;
ALTER TABLE some RENAME TO some_old;
ALTER TABLE some_old RENAME TO some;
```

EDIT: having tested it, it seems to do the trick on that table

PR #713, opened 2025-06-14
Only create indexes when they don't exist already

Amaras:
Following discussion #709, I've fixed the index creation on local player databases.

From the SQLite documentation : "If the optional IF NOT EXISTS clause is present and another index with the same name already exists, then [the CREATE INDEX command] becomes a no-op."


Molrn (approved):
LGTM

pascalaubry:
Closes #707

PR #715, opened 2025-06-14
Add option to deploy the server to the local network

Molrn:


Molrn:
discussed with @pascalaubry
- would be nice to deploy / undeploy on update instead of needing a restart
- @pascalaubry not fully convinced of the idea
- will require more thoughts and discussion on how to combine this with the roles

Bak to draft until then

pascalaubry:
Still waiting for roles to see how all this can match together.

Molrn:
> Still waiting for roles to see how all this can match together.  

from what you've showed, I'd say hiding the roles / devices menus and using no auth for the server feels appropriate. the help text would definitely need to ba adapted too, as it'd need to mention task sharing in addition to sceen display

PR #720, opened 2025-06-14
Prize fixes

Molrn:
- prize category reordering only considering the first prize group
- Unassigned prizes not showing on the Prize assignment doc


pascalaubry (approved):
LGTM

PR #725, opened 2025-06-15
Restrain the characters that can be used in event IDs

### LINKED ISSUE #718: [BUG] Event with spaces in id crashes after check-in
## Bug description
When I close the check-in, the application crashes because it tries to reach an invalid file name.

## How to reproduce
Steps to reproduce the bug:
1. Create an event with id "toto A"
2. Create a tournament and register two people in it
3. Open the check-in for the tournament
4. Close it, without checking in anyone
Image

7. See the error
Image

note that the file should be named "toto A.sce" instead.

That file still exist, but the .sce-wal and .sce-shm were removed

Image

## Behaviour expected
Either forbid spaces in ids, or do not crash if space is in the ID

## Environment
- OS: Windows 10, etc.
- Sharly Chess version: 2.7.7

pascalaubry:
Invalid filenames are updated when EventLoader loads the events.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -18,6 +18,7 @@ English version below
 ### Gestion des évènements
 
 - Correction d'un bug sur l'édition des évènements (2.7.6)
+- Restriction des caractères utilisables dans les identifiants des évènements (2.7.8)
 
 ### Gestion des tournois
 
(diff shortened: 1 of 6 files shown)


pascalaubry:
Closes #718

Amaras (approved):

PR #732, opened 2025-06-16
Prize currency

Molrn:


src/plugins/ffe/ffe.py
```diff
@@ -666,6 +666,10 @@ def get_validated_event_form_fields(
             }
         }
 
+    @hookimpl
+    def get_default_prize_currency(self) -> str:
```
timothyarmes, line 670: Maybe this should take the event's federation as a parameter.  The plugin would only return a value for the federations that it knows about.  That way if there are several plugins available (think of an international arbiter who moves about) then we take the first response that isn't None.

Molrn: it feels far fetched and useless. Currency is such a secondary thing, that is already is already fully customizable, with default options fetched with coutry libs, adding more ways to configure the default option is pointless. If there are currencies that are specific to some federations and unknown to the currency lib, they should be added overall in the `SharingUtils.get_country_currency` function, not handled by the plugin.  

Plus, this only makes sense for federations unknown by the `pycountry` module, so if we put here a `if federation == 'FRA'` it will never be executed.

timothyarmes: Ok. Do we need at plugin hook at all then ?

Molrn: I think it can be useful, say for example for an arbiter arbitering `FIDE` events in the US, then all of it's events being in euro is annoying. I see a usage here, but not by including the known federations (the only purpose of this is for unknown federations)

timothyarmes (dismissed):
Just missing the currency in the event overview page, other than that 👍

timothyarmes (approved):

PR #741, opened 2025-06-18
Move the FFE database, recover player database files when upgrading

Molrn:
replaces #722


src/common/engine.py
```diff
@@ -325,6 +328,20 @@ def _recover_previous_version(
                     shutil.copy(src_file, tournament.file)
                     logger.debug('%s > %s', str(src_file), str(tournament.file))
                     tournaments_number += 1
+        logger.info('Recovering misc files...')
+        files_to_recover: list[Path] = [
+            TMP_DIR
+            / f'{FideDatabase.static_id()}].{SharlyChessConfig.federation_database_ext}',
+        ] + list(
+            itertools.chain.from_iterable(plugin_manager.hook.get_files_to_recover())
+        )
```
Molrn:
```suggestion
        files_to_recover: list[Path] = [
            database.file for file in
            LocalSourceDatabaseManager.objects()
        ]   
```
There already is a way to list source databases, no need to use another hook

Molrn (commented):
@pascalaubry I think this is an unsafe feature as these databases do not have a migration structure. If we do any change to the database structure, it will not be in the copied database, which will most definitely result in a 500. 

IMO we should not merge this. We can, however, redownload the ones that are found in the previous version

pascalaubry:
> @pascalaubry I think this is an unsafe feature as these databases do not have a migration structure. If we do any change to the database structure, it will not be in the copied database, which will most definitely result in a 500.

Good point, we must be sure that downloading old databases will not break new instances.

That said, it is a legitimate feature do get back the local databases without having to download them again, especially for people with slow connections. So we should recover these databases whenever possible, i.e. when the database structures did not change since the release from which we can recover the files. Since we create these databases, we have full control on the databases and their structures and we are able to say wether a file can be recovered or not, depending on the release we are recovering files from. 

In our case (2.8.0), we know that the last change was done on the database structure on #713 for 2.7.8. So we should recover the database from release 2.7.8+. This needs some changes to set the minimum release to take into accound for each file.

Note: not using the entity manager was intented to make file recovering more flexible, but the entity manager may be a nicer way.  Let's talk about it.

pascalaubry (approved):
Just added comments on the last changes of the databases' structure.

PR #747, opened 2025-06-19
Relative performance print document (closes #724)

### LINKED ISSUE #724: Add print sort option for relative performance
Sorted by winner elo - loser elo,  lowest on top

timothyarmes:
Separate print doc

### DIFF
--- a/locale/fr/LC_MESSAGES/messages.po
+++ b/locale/fr/LC_MESSAGES/messages.po
@@ -3678,7 +3678,7 @@ msgid "Relative performances for round"
 msgstr "Performances relatives pour la ronde"
 
 msgid "Relative performances for round #{round}"
-msgstr "Performances relatives – ronde #{round}"
+msgstr "Performances relatives – Ronde #{round}"
 
 msgid "Reload the page"
 msgstr "Recharger la page"
--- a/locale/fr/LC_MESSAGES/messages.po
+++ b/locale/fr/LC_MESSAGES/messages.po
@@ -3170,6 +3170,9 @@ msgstr "Performance ELO parfaite"
 msgid "Performance"
 msgstr "Performance"
 
+msgid "Performance *** PERF FOR TABULAR DISPLAY"
+msgstr "Perf."
+
 msgid "Personal"
 msgstr "Personnel"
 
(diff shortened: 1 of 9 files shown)


locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3651,6 +3654,12 @@ msgstr "Actualiser"
 msgid "Refresh this page."
 msgstr "Actualiser la page."
 
+msgid "Relative performances for round"
+msgstr "Performances relatives pour la ronde"
+
+msgid "Relative performances for round #{round}"
+msgstr "Performances relatives – ronde #{round}"
```
```suggestion
msgstr "Performances relatives – Ronde #{round}"
```

src/data/print_documents/documents.py
```diff
@@ -1,7 +1,7 @@
 from abc import ABC, abstractmethod
 from functools import cached_property, partial
 import itertools
-from typing import Any, override
+from typing import Any, Tuple, override
```
Amaras: what's the point of `typing.Tuple` compared to `tuple`?

pascalaubry: Better use built-in ``tuple`` no?
```suggestion
from typing import Any, override
```

src/data/print_documents/documents.py
```diff
@@ -249,6 +249,89 @@ def is_crosstable(self) -> bool:
...
+            self._get_option(RoundPrintOption).value
+            or self.tournament.max_ranking_round
+        )
+
+    @property
+    def template_name(self) -> str:
+        return '/admin/print/round_performance.html'
+
+    @property
+    def ordered_players(self) -> list[Tuple[Player, Player, Result, float]]:
```
pascalaubry:
```suggestion
    def ordered_players(self) -> list[tuple[Player, Player, Result, float]]:
```

src/data/print_documents/documents.py
```diff
@@ -249,6 +249,89 @@ def is_crosstable(self) -> bool:
...
+        assert self.tournament is not None
+        ranking_round = self.ranking_round
+        if not ranking_round:
+            return []
+        results: list[Tuple[Player, Player, Result, float]] = []
```
pascalaubry:
```suggestion
        results: list[tuple[Player, Player, Result, float]] = []
```

Amaras (changes requested):

pascalaubry (changes requested):
Just a problem for round #0.
I think we should not be able to print until one round is finished.
(and maybe preset the last finished round?)
image
image
Except this it is a nice view.
Last thing: to make it more readable, could we add a *diff* column?

pascalaubry (dismissed):

PR #749, opened 2025-06-19
Set the user language at system-level when first starting the application

pascalaubry:


src/common/sharly_chess_config.py
```diff
@@ -44,32 +45,40 @@ def reload(cls):
...
+                windll.GetUserDefaultUILanguage()
+            ]
+            logger.info('User locale: %s', system_user_locale)
+            locale: str = system_user_locale[:2]
+            if locale in locales:
+                return locale
+            logger.warning('Unknown locale: %s', locale)
+            return None
+        else:
+            raise OSError(f'{sys.platform=}')
```
Molrn:
```suggestion
        return None
```
If we've got that kind of OS filter, it definitely should not be here, and it should be very explicit. As we are doing whatever we can to make the app open to more OSs, I'm really opposed to having that kind of error raised. it also excludes the other windows distrib, on which SC might actually work ('cygwin' / 'msys')

Molrn (dismissed):

Molrn (approved):
LGTM

PR #752, opened 2025-06-20
Set the default federation on the first startup

pascalaubry:
Set ``FRA`` if the language is ``fr``.
@timothyarmes Feel free to set something for ``en`` but I am not sure that we have to choose between ``ENG`` and ``USA``.

### DIFF
--- a/src/common/sharly_chess_config.py
+++ b/src/common/sharly_chess_config.py
@@ -57,25 +57,26 @@ class SharlyChessConfig(metaclass=Singleton):
                 windll.GetUserDefaultUILanguage()
             ]
             logger.info('User locale: %s', system_user_locale)
-            locale: str = system_user_locale[:2]
+            user_locale: str = system_user_locale[:2]
             if locale in locales:
-                return locale
-            logger.warning('Unknown locale: %s', locale)
+                return user_locale
+            logger.warning('Unknown locale: %s', user_locale)
         return None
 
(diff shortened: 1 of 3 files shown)


timothyarmes:
> Set `FRA` if the language is `fr`. @timothyarmes Feel free to set something for `en` but I am not sure that we have to choose between `ENG` and `USA`.

The date format is inverted, so ideally you'd use the right locale.

Molrn:
I think it would feel right for it to be `FID` for locale `en`. English is our international format, it is not specific to one federation.   
We could also go much further with this, and use the `pycountry` module (already in the project for prize currency) to fetch the federation matching the country of the locale, and return 'FID' if none is found.  
this should do the trick:
```py
    @staticmethod
    def get_locale_federation(locale: str) -> str:
        country_code = locale.split('_')[-1].upper()
        country = pycountry.countries.get(alpha_2=country_code)
        if country and country.alpha_3 in SharlyChessConfig.federations:
            return country.alpha_3
        return SharlyChessConfig.default_federation
```

Molrn (dismissed):
image  
 "Par défaut" values should be removed here

pascalaubry:
We talked about it yesterday with @timothyarmes : the final solution is not this one but we progress... ;-)

Molrn (approved):
LGTM

PR #763, opened 2025-06-23
Improved the QR code

pascalaubry:
I feel that the QR codes will have many other use cases in the future so I:
- factorize the code and made the QR code available with a request to a new controller (that may be used for future other QR codes)
- added the logo for branding and use the colors of the application for it
- used JPG in place of PNG to get quite small images dispite the logo
- made the logo bigger to get printable images
sharly_chess_qr (16)


timothyarmes:
There are a few thinks that I don't like:

* The grey background.  It looks dirty. White's not only looks cleaner, it takes less ink to print
* The logo. The fact that the circle touches the rest of the QR code looks ugly, and the use of colours is unnecessary, especially on a black and white printer.  I think you should use just the S (no circle), in black, and make sure there a bit of space around it (for example, make the existing circle white).
* Writing into the app.  I _really_ don't like that you're writing the logo into the app's static files, we have a tmp folder, that's what you should be using
* The size in the collapse.  The QR code is now too big when you open click "See QR code".  I see no reason for the border, especially if you make the background white again, so that'll help a bit

The logo's a nice addition when you print, but it makes the QR code bigger to add enough error correction.  We may wish at some point to display the QR code in the bottom of a Screen, in which case the bigger size will take up precious real estate.  I think it should be optional.

pascalaubry:
I agree with everything but the static files, details below.

> There are a few thinks that I don't like:
> 
> * The grey background.  It looks dirty. White's not only looks cleaner, it takes less ink to print

Black on white now.

> * The logo. The fact that the circle touches the rest of the QR code looks ugly, and the use of colours is unnecessary, especially on a black and white printer.  I think you should use just the S (no circle), in black, and make sure there a bit of space around it (for example, make the existing circle white).



> * Writing into the app.  I _really_ don't like that you're writing the logo into the app's static files, we have a tmp folder, that's what you should be using

Nothing written to the static files, I only use tmp files.
The only static file is the logo used to build the QR code (could not find a better place but feel free to find one).

> * The size in the collapse.  The QR code is now too big when you open click "See QR code".  I see no reason for the border, especially if you make the background white again, so that'll help a bit
> 
> The logo's a nice addition when you print, but it makes the QR code bigger to add enough error correction. We may wish at some point to display the QR code in the bottom of a Screen, in which case the bigger size will take up precious real estate. I think it should be optional.

The logo is now optional in the URL, and the printable version is available via a new link.

image

And the QR code with the logo is also B&W.
 
image

timothyarmes (approved):

PR #764, opened 2025-06-23
Change relative performance formula

timothyarmes:


src/data/print_documents/documents.py
```diff
@@ -291,15 +291,15 @@ def ordered_players(self) -> list[tuple[Player, Player, Result, float]]:
...
-                max_score = Result.GAIN.points(self.tournament.point_values)
-                fractional_score = round(
-                    pairing.result.points(self.tournament.point_values) / max_score, 2
-                )
-                bonus = StaticUtils.performance_bonus(fractional_score)
-                tpr_for_round = StaticUtils.round_ranking(opponent.rating + bonus)
-                results.append(
-                    (player, opponent, pairing.result, tpr_for_round - player.rating)
-                )
+                perf = opponent.rating - player.rating
```
Molrn: unused. if it is the default value, should explicitely be in the `case _:` part of the match statement

src/data/print_documents/documents.py
```diff
@@ -291,15 +291,15 @@ def ordered_players(self) -> list[tuple[Player, Player, Result, float]]:
...
+                match pairing.result:
+                    case Result.GAIN:
+                        perf = opponent.rating + 400
+                    case Result.LOSS:
+                        perf = opponent.rating - 400
+                    case Result.DRAW:
+                        perf = opponent.rating
```
Molrn: UNRATED_{result} are not taken into account, I feel like they sould be

Molrn (changes requested):

src/data/print_documents/documents.py
timothyarmes: This has changed, but note that only played games are listed anyway.

Molrn (approved):
LGTM

PR #773, opened 2025-06-25
Beep when result differs on keyboard entry

timothyarmes:


Molrn (changes requested):
works, but I find the sound rather agressive... I like the first one of this page: https://pixabay.com/sound-effects/search/error/

Molrn (approved):
LGTM

PR #779, opened 2025-06-26
Create multiple prize from a `values` text field

Molrn:


src/web/templates/admin/prizes/prize_form_modal.html
```diff
@@ -50,13 +51,23 @@ <h4 class="mb-3">
...
-                        help_text='',
-                        step=0.01,
-                        data=data, errors=errors
-                    ) }}
+                    {% if action == 'create' %}
+                        {{ admin_macros.text_input(
+                            name='values',
+                            label=_('%(string)s:', string=_('Values')),
+                            placeholder=_('E.g.: %(string)s', string='1000 500,5 333.33'),
+                            help_text=_('List of values of the prizes to create. Values can be floats and have to be positive.'),
```
Amaras: You have to say that they have to be space-separated, otherwise it's gonna be confusing

Molrn: `Space-separated list of numbers. One prize is created per value in the list.`

timothyarmes:
Works perfectly, but even though this isn't linked to the PR, can you please fix it here:

image

The value is NON-monetary, it shouldn't have a € sign. Perhaps call it "comparative value"?

Molrn:
image  
Not a fan of "comparative value", feels too verbous. this looks fine to me

timothyarmes (approved):

PR #782, opened 2025-06-27
Fix foreign key pointers

Amaras:
Up until now, foreign keys pointing to the `tournament` table actually point to `tournament_copy`.

This PR adds a DB migration to make sure all the foreign keys point to the correct table.


Molrn (approved):
I was doing the same PR ^^ this can go to 2.8, it's the one enabling the fk pragma which is dangerous. I'm going through it rn

PR #784, opened 2025-06-27
Enable foreign key pragma

### LINKED ISSUE #707: Restore FK constraints

Molrn:
closes #707 
I went around and everything appears to work well. Milestone TBD

### DIFF
--- a/src/data/tournament.py
+++ b/src/data/tournament.py
@@ -639,14 +639,13 @@ class Tournament:
                     | ScreenType.PLAYERS
                     | ScreenType.RANKING
                 ):
-                    for screen_set in screen.screen_sets_sorted_by_order:
-                        if screen_set.tournament.id == self.id:
-                            dependent_screens.append(screen)
-                case ScreenType.RESULTS:
-                    if (
-                        not screen.results_tournament_ids
-                        or self.id in screen.results_tournament_ids
+                    if all(
+                        screen_set.tournament.id == self.id
+                        for screen_set in screen.screen_sets_by_id.values()
                     ):
+                        dependent_screens.append(screen)
+                case ScreenType.RESULTS:
+                    if screen.results_tournament_ids == [self.id]:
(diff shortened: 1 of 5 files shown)


src/data/tournament.py
```diff
@@ -631,6 +631,32 @@ def dependent_families(self) -> list[Family]:
     @property
     def dependent_screens(self) -> list[Screen]:
```
Amaras, line 632: I'd prefer if those (`Tournament.dependant_screens` and `Tournament.related_screens`) were generators instead of returning lists, but it's alright

Amaras (commented):
Haven't tested, the changes, but the code looks good

src/data/tournament.py
Molrn: I wasn't careful about the difference until then, I tend to only use `list` so there must be a bunch of lists that should instead be generators... It might be worth doing a PR going over all of those in order to be more memory efficient.  
That being said, I don't think it should be in that PR

Amaras: Yeah, not that PR

Amaras (approved):

PR #785, opened 2025-06-28
Uniformize player localized name

Molrn:


src/data/player.py
```diff
@@ -126,7 +126,9 @@ def tournament(self, tournament: Optional['Tournament']):  # type: ignore
     @property
     def full_name(self) -> str:
         if self.first_name:
-            return f'{self.last_name} {self.first_name}'
+            return _('{first_name} {last_name}').format(
```
Amaras, line 129: Alright, but name translation is a pain.
Are you really going to use non-native name order with native name?

Molrn: the rest of it had already been localized in #598, I think this one was forgotten. I don't have much of an opinion, this was more about uniformizing it. Having `first_name` first in english semt important, in french it's more natural to have it the other way around, so having the order localized ok. I don't understand what you mean by native / non-native, can you elaborate please ?

Amaras: Compare "Paul Erdős" (in English with name translated), with "Erdős Pál" (in Hungarian).
Sorry, he's my go-to example on this subject.

French tends to actually use `first_name last_name` order, even with foreign names, so you'll see "Pál Erdős" on his Wikipedia page.

The name order is not that easy, as English tends to respect name order, but may translate names (and change the name order as a result, because why not)

Molrn: not much we can do to distinguish these usages though. Most important is that it is coherent throughout the entire app, which is what's intended by this PR.

timothyarmes (approved):

PR #786, opened 2025-06-30
Add roles

pascalaubry:
Big PR that implements https://github.com/Sharly-Chess/sharly-chess/blob/feature/roles/docs/sandbox/authentication-and-roles.md
A few screenshots below, I probably need to explain you a few things before you start reviewing...
image
image
image
image
image
image


src/data/auth/roles.py
```diff
@@ -0,0 +1,277 @@
...
+                return _('Application')
+            case RoleScope.EVENT:
+                return _('Event')
+            case RoleScope.TOURNAMENT:
+                return _('Tournament')
+            case _:
+                raise ValueError(f'role={self}')
+
+
+class Role(IntEnum):
```
Molrn: Using big enums like this to represent complex structures is something we've been trying to remove from the project. It would instead be much better to use `class Role(IdentifiableEntity)` and an entity manager

src/data/auth/mode.py
```diff
@@ -0,0 +1,140 @@
+from enum import IntEnum
+from typing import Self
+
+from common.i18n import _
+from data.auth.roles import Role
+
+
+class Mode(IntEnum):
```
Molrn: if at least modes could be in an individual PR it would be good. seems doable to isolate this code
PS: IntEnum -> IdentifiableEntity

src/database/sqlite/event/migrations/m031_add_roles.py
```diff
@@ -0,0 +1,81 @@
+from data.auth.mode import Mode
+from database.sqlite.event.event_store import (
+    anonymous_stored_account,
+    localhost_stored_computer,
+    unknown_stored_computer,
+)
```
Molrn: No project import in a migration for stability reasons. If you need to use constants, you redeclare them as they are in the project at the moment of the migration

src/database/sqlite/event/event_store.py
```diff
@@ -220,6 +223,74 @@ class StoredDisplayController:
...
+)
+
+unknown_stored_computer: StoredComputer = StoredComputer(
+    id=ANY_COMPUTER_ID,
+    edit_properties=False,
+    edit_permissions=True,
+    active=True,
+    permissions={},
+    ip=None,
+)
```
Molrn: these should be defined outside of the store (ex: `auth.__init__`)

src/plugins/chessevent/templates/chessevent_event_card_block.html
```diff
@@ -1,4 +1,4 @@
-{% if admin_events_show_details %}
+{% if client.can_view_detailed_event_cards and admin_events_show_details %}
```
Molrn, line 1:
```suggestion
{% if admin_events_show_details %}
```
event_admin_show_details should always be false if not `client.can_view_detailed_event_cards`. This avoids propagation in the templates and potenitial errors. Same for every `client.can_view_detailed_{tab}_cards`

Molrn (changes requested):

Molrn:
# UI / integration feedback  

- suggestion: card containers for devices / accounts
- Tournament roles selection should be a multi-select (see prize criteria > Federation)
image
- The `Accès` menu should not appear if the mode does not allow it. In general, for modes, hiding the options should be prefered to disabling them
image
- this should IMO be a select in an individual form, similar to prizes. When a prize category is created, it has no criteria with a warning on its card. You can then add criteria individually, with a select in a form modal showing a dynamic form (in this case: a tournament select + the exlanatory text for the permission)  
- There should not be elements displayed by default and instead a documentation text (like screens and prizes). The doc ending by `Create default devices to get started` with a button `Create default devices`
image
- this should be something less techy and generic. The language should only be displayed on the modal. 
image suggestion:  image

src/data/auth/roles.py
pascalaubry: I leave this one open since I will need you for it.

src/data/auth/mode.py
pascalaubry: I leave this one open since I will need you for it.

src/database/sqlite/event/migrations/m031_add_roles.py
pascalaubry: I understand but I could not just add constants so I added a `_post_migrate()` to `MigrationDatabase`.
Tell me if you have a better idea.

src/database/sqlite/event/event_store.py
pascalaubry: Moved this code to a class method.

src/plugins/chessevent/templates/chessevent_event_card_block.html
pascalaubry: In  fact no: permissions can change while admin_xxx_show_details is and remains true, so both tests are needed.

pascalaubry:
> * suggestion: card containers for devices / accounts

Excellent idea!

pascalaubry:
> * Tournament roles selection should be a multi-select (see prize criteria > Federation)
>   image

Why not.

pascalaubry:
> * The `Accès` menu should not appear if the mode does not allow it. In general, for modes, hiding the options should be prefered to disabling them
>   image

Done.

pascalaubry:
> * this should IMO be a select in an individual form, similar to prizes. When a prize category is created, it has no criteria with a warning on its card. You can then add criteria individually, with a select in a form modal showing a dynamic form (in this case: a tournament select + the exlanatory text for the permission)

I don"t understand this one, to be discussed.

pascalaubry:
> * There should not be elements displayed by default and instead a documentation text (like screens and prizes). The doc ending by `Create default devices to get started` with a button `Create default devices`
>   image

They are not elements displayed by default, they are the real permissions set in the mode (now they are shown only for `ExecMode.CUSTOM`)..

pascalaubry:
> * this should be something less techy and generic. The language should only be displayed on the modal.
>   image suggestion:  image

The profile is show to all users so I tried to find the good compromise between too techchy and useles. Not sure for the language, but I think all the other information is needed.

pascalaubry:
Very nice review, very complementary to Tim's.
Thank you very much.

src/database/sqlite/migration_database.py
```diff
@@ -126,6 +127,12 @@ def create(self):
         with self.create_instance(self.file, True) as database:
             for manager in database.migration_managers:
                 manager.migrate()
+            database._post_migrate()
+
+    def _post_migrate(self):
+        """Allow databases to do thing after the last migration,
+        see EventDatabase._post_migrate for an example."""
+        pass
```
Molrn: That's a terrible idea, it will be executed at the end of every new migration. Currently, it would add your new devices over and over. Even with a check of only append if the ids exist, it is inefficient and a bad maintenance idea.  
If you need to add default objects, it has to be done in the migration, like the one row in the info table which is added in m001. Data insertion is also a valid migration operation, it's not just about the data structure.

pascalaubry: You are absolutely right (forgot that it will be run at **every** migration).
I will restore what I did previously until we find a better way (let's talk about it).

src/web/templates/common/profile/profile_modal.html
```diff
@@ -0,0 +1,221 @@
...
+    </div>
+</div>
+
+<script>
+    // We need to reload the page when the modal is hidden, because profile and language may have changed.
+    $(document).one('hide.bs.modal', function(e) {
+        e.preventDefault();
+        document.body.dispatchEvent(new CustomEvent('request_refresh'));
+    });
+</script>
```
Molrn, line 221: requires the base page to have a `refresh_listener`, but the modal can be opened everywhere, even on pages without it. It therefore just prevents the default behavior and does not close the modal

Molrn (dismissed):

timothyarmes (approved):
I"ve been through it.  It's basically impossible to review, there are so many changes.  In particular to the templates.
I just want to get it merged so that we can move forwards with refactoring and writing tests

Molrn (approved):
Re-read, test needs fixing before merging but otherwise we can merge

PR #795, opened 2025-07-01
Test infrastructure

timothyarmes:


src/common/engine.py
```diff
@@ -571,6 +572,8 @@ def _check_version(cls) -> tuple[Version | None, str | None]:
         """Compares the current version with the most recent version on the Sharly Chess GitHub repository
         If the current release is stable, more recent pre-releases are ignored; otherwise the most recent release is chosen.
         Returns the most recent version available and the corresponding down URL if any, None otherwise."""
+        if os.getenv('TEST_ENV') == 'true':
```
pascalaubry: As `os.getenv('TEST_ENV')` is used in sevevral files, can we add a `TEST_ENV` next to `DEVEL_ENV`?

pascalaubry (dismissed):

src/common/engine.py
timothyarmes: Quite right, I wanted to do that and forgot

pascalaubry (dismissed):
Most of the tests failed (cf enclosed test.txt) but the architecture itself works (no impact on the program itself).
I still believe that we should all use the same venv and create it only when it does not exist (can be done in another PR).

pascalaubry:
[test.txt](https://github.com/user-attachments/files/21019797/test.txt)

pascalaubry (approved):

PR #796, opened 2025-07-01
Fix default FFE ID being None

Amaras:
When creating a player from the form, we needed to create a default "RefFFE" value.

However, since all `Player` instances have a `ffe_id` field defaulting to `None`, this value was simply not set in the DB.

As a consequence, this created a problem when modifying players which had a NULL "RefFFE" field in Papi, since there is a UNIQUE index on that field.

Since the `dateutil` dependency was only used in the modified file, and I removed that use, I also removed the depencency from the `pyproject.toml`

- **Change the default RefFFE**
- **Remove unused dateutil dependency**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,6 @@ dependencies = [
   "colorama ~= 0.4.6",
   "colorlog ~= 6.9.0",
   "cryptography ~= 44.0.0",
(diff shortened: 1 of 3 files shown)


pascalaubry (approved):
Nice.

PR #797, opened 2025-07-01
Fix default RefFFE not being set correctly

Amaras:
This caused problems when creating players from scratch.

See #796 for the same problem's fix in 2.8


pascalaubry (approved):
Nice too ;-)

PR #799, opened 2025-07-02
Use entities for roles + Controller refactor + permissions multi-select form

Molrn:
Requires a rollback / migrate of migration m031


Molrn:
@timothyarmes still a detail to handle on the permission form JS, putting this as ready so it can move forward

timothyarmes (approved):

PR #805, opened 2025-07-03
Move the creation of exec mode objects from the controllers to the database layer

pascalaubry:


Molrn (changes requested):
That is not what we mean by `gathering the logic`. We mean that functions should execute what they appear to execute, not more, not less, and that the execution should not rely on a previous operation.  

Here, you call the method `add_stored_account` which in the end adds 2 objects (=> not the expected behavior). Before, you had it in such a way that the interpretation of the operation `add_stored_account` depended on the previous operation which was adding the default values.  

Concretely, that means:  
- no notion of the default objects in the `EventDatabase` class
- a way to identify the default objects (if even needed) not based on the ID, but based on specific values (in your case, `username is None` and `IP == '127.0.0.1'`)
- an abstract method `create_default_objects` in `AuthEntity`

Molrn:
adressed by #799, I think we can close

PR #806, opened 2025-07-04
Fix archive tab styling

timothyarmes:


Molrn (changes requested):
image
the info link leads nowhere, if you don't plan to put a doc page for it (not needed IMO) it probably should not be there

Molrn (approved):
LGTM

PR #809, opened 2025-07-04
Improve security through password handling

Amaras:
Prior to this PR, the security of the proposed authentication method was severely lacking.

This PR proposes a POC which solves the following issues in the previous model:
1. Passwords were stored in plaintext (absolutely not acceptable in any serious application, which we aim to be).
2. The candidate and stored passwords were compared using the `==` operator, which bails at the first failing comparison, allowing attackers to gain information about the position of the first difference.
3. Invalid username and invalid password errors were fully differentiated on the user side, which allows attackers to know which of the long/password pair is wrong, simplifying their brute-force search.

While the problem with 1 seems obvious to me, allow me to explain a possible problematic situation:
- Suppose we have a malicious insider who helps us with support (M).
- The arbiter (A) of a tournament in which M wants to play needs help with his configuration.
- A sends M his config, including the roles with plaintext passwords, and M helps A.
- M then plays in the tournament, and, without any one realizing, changes values to favour them with a higher-level account, then reconnects to the expected account.
- Because no one realizes, M wins a prize they should not have, and A is in trouble after the tournament, because the prize was already given to M.
- M refuses to give the prize back, leading to disciplinary or legal action against M (for dishonesty), against A (because it's their responsibility to check).

In that situation, I think we would be responsible both for allowing a malicious actor to be in such a priviledged position in our organization, and for making it easy for that malicious actor to attack an event.
It also deals a huge blow to our reputation as a serious tournament manager, which may prevent us from being endorsed by FIDE for some time, or even get our endorsement suspended if we already are endorsed.

To solve 1 properly, we need a secure hashing algorithm with secure random salting.
This is achieved through argon2-cffi, which provides Python bindings to Argon2, the winner of the [Password Hashing Competition](https://www.password-hashing.net).
The implementation is the Argon2 reference implementation, which was proven as secure as possible, although cryptanalysis was not as thourough as during NIST competitons.
Parameters are thus far left to the library author, and seem to be secure.
Password hash verification seems to take around 100ms on my machine.
To check on your machine, install `argon2-cffi`, then run `python -m argon2`

Things to improve for a better version:
- Fine-tune Argon2 parameters to get it working similarly on all our supported platforms (even potato laptops).
- Use Litestar's authentication features (rolling your own auth is not really advised in general).
- Other possible improvements I did not think about yet.


timothyarmes (approved):

Molrn (approved):
LGTM

PR #817, opened 2025-07-07
Add GH Action infrastructure to autorun tests

Amaras:
This PR adds two workflows for two different purposes.

The first (test.yml) runs the collection of tests that are not slow, so tests that are can be run quickly are always checked.

The second (full_tests.yml) runs all the tests  (including the slow ones) on each push to `dev`, as well as when manually triggered by a maintainer.

Both workflows should be integrated as a prerequisite to merging on `dev`, especially the full test one.

Note: both have to install all tests dependencies and browsers each time. It might not be worth the installation time.

Both files were fully generated by Copilot with minor adjustments, but they look pretty good, so I hope they are not faulty.


Amaras:
Drafted because I need to see if I can install the ODBC drivers

Amaras:
Undrafted, because somehow it works

timothyarmes (approved):

PR #821, opened 2025-07-08
Fix cleared result from input screen

Molrn:
Careful on the merge back, this will fail on dev because of the stored player changes


pascalaubry (approved):

pascalaubry (commented):
Maybe add a message to the release notes?

PR #826, opened 2025-07-08
Test screens

timothyarmes:


src/common/engine.py
```diff
@@ -113,7 +113,7 @@ def __init__(self):
                 file.rename(event_database.file)
                 # now load the new database
                 EventLoader.get(request=None).load_event(event_uniq_id)
-        if not EventLoader.get(request=None).event_uniq_ids:
+        if not EventLoader.get(request=None).event_uniq_ids and not TEST_ENV:
```
Molrn:
```suggestion
        if not EventLoader().event_uniq_ids and not TEST_ENV:
```

src/common/sharly_chess_config.py
Molrn, line 1: I think that's the kind of override that would be better with a `unittest.mock.patch`. It would also allow you to put less than 1 without failing the type constraint. If you disagree / it's a pain to use a patch here feel free to leave it as such ;)

src/common/engine.py
```diff
@@ -98,7 +98,7 @@ def __init__(self):
                 ]:
                     break
                 raise ValueError(f'choice=[{choice}]')
-        if not EventLoader.get(request=None).event_uniq_ids:
+        if not EventLoader.get(request=None).event_uniq_ids and not TEST_ENV:
```
Molrn:
```suggestion
        if not EventLoader().event_uniq_ids and not TEST_ENV:
```

tests/papi/test-screensLs.html
Molrn, line 1: to remove

tests/papi/papi-berger.papi
Molrn, line 1: shoudn't be modified. Same for other papi files

Molrn (changes requested):

Molrn (approved):
LGTM

PR #831, opened 2025-07-11
Do not update i18n files for tests

pascalaubry:


src/common/__init__.py
```diff
@@ -18,7 +18,7 @@
 # True when the program is running in a development environment, False if running as an EXE file.
 DEVEL_ENV: bool = not getattr(sys, 'frozen', False)
 TEST_ENV: bool = os.getenv('TEST_ENV') == 'true'
-
+print(f'{os.environ=}')
```
timothyarmes: Remove?

PR #835, opened 2025-07-11
Make role permissions form dynamic

Molrn:
- display the help text in a tooltip
- disable the inherited roles with a `inherited` label
- only display the tournament selector if a tournament scope role is selected


timothyarmes (approved):
Nice.

PR #837, opened 2025-07-11
Added a first guard

pascalaubry:
The goal of this PR is to show how it could be possible to move the access control from the controllers to guards.


Molrn (approved):

pascalaubry:
Thanks for the review, I keep it open until we discuss the opportunity of reading things from the request (event, client, screen) rather than computing them again in the web context.

Molrn:
IMO it would be good for the client as it would avoid duplicating a few computations, however I see no value for screen or event, Adding this kind of value to requests objects is not a clean practice, I think it would be better to limit it to when it actually brings value

Amaras (dismissed):
The `_get_*_from_*_uniq_id` name make no sense to me, you are passing only a request, not a `uniq_id`.

Amaras (approved):

PR #839, opened 2025-07-11
Order roles

pascalaubry:
Before:
image
After:
image

### DIFF
--- a/src/data/auth/client.py
+++ b/src/data/auth/client.py
@@ -234,7 +234,7 @@ class Client:
         for auth_entity in (self.active_account, self.active_device):
             for role in auth_entity.roles:
                 manageable_roles |= role.manageable_roles()
-        return list(manageable_roles)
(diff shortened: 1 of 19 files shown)


Molrn (dismissed):
the manager is here for that, if you want to reorder the entities you reorder the classes in this list:
```py
class RoleManager(EntityManager[Role]):
    @staticmethod
    def entity_types() -> list[type[Role]]:
        return [
            AdministrationRole,
            OrganizationRole,
            ScreenManagementRole,
            ChiefArbitrationRole,
            DeputyChiefArbitrationRole,
            SectorArbitrationRole,
            PairingRole,
            CheckInRole,
            ResultsEntryRole,
            SpectatorRole,
        ]
```
As this is about reordering a set into a list, you can do:  
`[role for role in RoleManager.objects() if role in manageable_roles]`

Molrn (approved):

PR #840, opened 2025-07-11
Allow devices/accounts to have no permissions

pascalaubry:
Prior to this PR, granted permissions could not be all revoked.

### DIFF
--- a/locale/fr/LC_MESSAGES/messages.po
+++ b/locale/fr/LC_MESSAGES/messages.po
@@ -471,9 +471,6 @@ msgstr "Au moins une ligue est attendue."
 msgid "At least one rating type is expected."
 msgstr "Au moins un type de classement est attendu."
 
-msgid "At least one role is expected."
-msgstr "Au moins un rôle doit être donné."
-
 msgid "At least one value is expected."
 msgstr "Au moins une valeur est attendue."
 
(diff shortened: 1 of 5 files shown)


locale/fr/LC_MESSAGES/messages.po
```diff
@@ -841,15 +838,15 @@ msgstr "Choisir au moins une ligue."
...
-msgstr "Veuillez choisir au moins un rôle."
-
 msgid "Choose one, many or no tournaments."
 msgstr "Choisissez aucun, un ou plusieurs tournois."
 
 msgid "Choose the columns displayed."
 msgstr "Choisir les colonnes affichées."
 
+msgid "Choose the roles."
+msgstr Choisissez les rôles.
```
```suggestion
msgstr "Choisissez les rôles."
```

Molrn (dismissed):
That's IMO not the correct approach, it is important to keep an error in the form if no role is defined as it mostly will happen as an error from the user. I suggest instead to add a no-op role from which Spectator inherits.

pascalaubry:
> That's IMO not the correct approach, it is important to keep an error in the form if no role is defined as it mostly will happen as an error from the user. I suggest instead to add a no-op role from which Spectator inherits.

I think everybody understands that no roles means no permissions (at all), I changed the placeholder to make it explicit.

Molrn (approved):

PR #841, opened 2025-07-12
Keep only two execution modes (standard/custom)

pascalaubry:


src/data/auth/exec_mode.py
Molrn, line 1: It is now a boolean, it should be treated as such. Still having it in a enum is legacy of it being a select.   
All of the used properties of the enum should be defined with `if event.custom_exec_mode: ... else: ...`

src/database/sqlite/event/event_store.py
```diff
@@ -249,6 +249,8 @@ class StoredAccount(StoredAccess):
 
 DEFAULT_HIDE_BACKGROUND_IMAGE: bool = False
 
+DEFAULT_CUSTOM_EXEC_MODE: bool = False
+
```
Molrn: This kind of constant has nothing to do in the store, but belong it the `SharlyChessConfig`. While you're at it, could you please also refoactor `DEFAULT_HIDE_BACKGROUND_IMAGE` ? StoredEvent is no longer used without all of its values defined, it is no longer useful to Have defaults (I think it was mainly for testing)

src/database/sqlite/event/migrations/m031_add_roles.py
```diff
@@ -1,3 +1,4 @@
+from database.sqlite.event.event_store import DEFAULT_CUSTOM_EXEC_MODE
```
Molrn:
```suggestion
from database.sqlite.event.event_store import DEFAULT_CUSTOM_EXEC_MODE
```

No import from the project in migrations

Molrn (dismissed):
image  
There is no execution in the config at app level, this should not be there.

src/database/sqlite/event/event_store.py
pascalaubry: I removed the constants from stored_event.py on my last commit.
If you want me to remove the reference to SharlyChessConfig from m005 and m031, please confirm that you want to define the default values twice, in the migration and in SharlyChessConfig (bad idea IMO).

src/data/auth/exec_mode.py
pascalaubry: > It is now a boolean, it should be treated as such. Still having it in a enum is legacy of it being a select. All of the used properties of the enum should be defined with `if event.custom_exec_mode: ... else: ...`

OK but what about `predefined_devices` and `predefined_accounts`?

src/database/sqlite/event/event_store.py
Molrn: yes, self containing the code of each migration to its file only is a stability measure, even if it means doing a 200 / 300 lines migration.  

In addition, they do not represent the same thing:
- In the migration: value given to previous events when they are upgraded
- In SharlyChessConfig: default value of the event creation form.  
Those are not necessarly the same

src/data/auth/exec_mode.py
Molrn: in Event:
```py
@property
def predefined_devices(self):
    if self.custom_exec_mode:
        ...
    else:
        ...
```

Molrn (approved):
LGTM

PR #844, opened 2025-07-13
Fix the abstraction level of the Client class

Amaras:
Currently, the `Client` class uses tournament ids as a part of its main interface to check for permissions.

This is completely the wrong level of abstraction, as it should be an implementation detail.

I have started to change the templates to use the higher abstraction methods.

FIXME: I have skipped a test that was not passing, but it should be reactivated before merging on `feature/roles`.

### DIFF
--- a/src/web/templates/admin/pairings/pairing_modal.html
+++ b/src/web/templates/admin/pairings/pairing_modal.html
@@ -22,9 +22,9 @@
                         {% if
                             board.exempt
                             or
-                            (admin_tournament.id not in client.tournament_ids_allowing_enter_results)
+                            client.can_enter_results(admin_tournament.id)
                             or
-                            (board.result and (admin_tournament.id not in client.tournament_ids_allowing_update_results))
+                            (board.result and not client.can_update_results(admin_tournament.id))
                         %}
                             disabled
                         {% else %}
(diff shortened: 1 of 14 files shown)


src/web/templates/admin/pairings/pairing_modal.html
```diff
@@ -22,9 +22,9 @@
                         {% if
                             board.exempt
                             or
-                            (admin_tournament.id not in client.tournament_ids_allowing_enter_results)
+                            client.can_enter_results(admin_tournament.id)
```
pascalaubry:
```suggestion
                            not client.can_enter_results(admin_tournament.id)
```

src/web/templates/admin/pairings/pairing_modal.html
```diff
@@ -60,9 +60,9 @@
...
```
pascalaubry:
```suggestion
                            not client.can_enter_results(admin_tournament.id)
```

src/web/templates/user/modals.html
```diff
@@ -38,7 +38,7 @@
                             >{{ _('DRAW') }}<br/>½ - ½<br/><br/><br/><br/><br/></button>
                             <button
                                 class="result-button black text-white bg-black col-3 ms-auto me-auto btn fs-4 fw-bold p-3"
-                                {% if board.result and (wp.tournament.id not in client.tournament_ids_allowing_update_results) %}
+                                {% if board.result and client.can_update_results(wp.tournament.id) %}
```
pascalaubry:
```suggestion
                                {% if board.result and not client.can_update_results(wp.tournament.id) %}
```

pascalaubry (changes requested):
Really can't understand why I used dicts then lists instead of methods...

pascalaubry (approved):

PR #847, opened 2025-07-19
Various improvements I see from Avoine 2025

Amaras:
- Fixed an uncaught exception
- Added the number of unpaired players to the list (mainly to easily check if there is a need to add a player dedicated to parity regularization)

Draft because I think I'll add other things.
It will also probably be split up into several PRs


timothyarmes (approved):
LGTM

timothyarmes:
The changes have been added as a PR to dev instead

PR #848, opened 2025-07-19
Output nothing to PGN when players have no FIDE title

pascalaubry:


Amaras (dismissed):
Note: according to the specs, the behaviour is correct, but DGT LiveChess does not parse correctly with `"-"`
Also, I'd rather not have `"?"` as the first name when it is not set

src/data/board.py
```diff
@@ -81,11 +81,17 @@ def _player_to_pgn(cls, player: Player | None, is_white: bool) -> str:
         if player is None:
             return f'[{field_prefix} ""]'
         rating = player.rating if player.rating_type == PlayerRatingType.FIDE else '-'
-        name = f'{player.last_name}, {player.first_name or "?"}'
+        name = f'{player.last_name}' + (
```
Molrn:
```suggestion
        name = player.last_name + (
```

Molrn (dismissed):

Molrn (approved):
LGTM

PR #851, opened 2025-07-22
Fix Client.imputable_results_for_tournament()

pascalaubry:


Molrn:
why merge to [bugfix/user-result-modal-clear](https://github.com/Sharly-Chess/sharly-chess/tree/bugfix/user-result-modal-clear) instead of `feature/roles` ?

PR #852, opened 2025-07-22
Add guards on every user end-point

pascalaubry:
Access control is no more in controllers but in guards.
Common code between controllers and guards is factorized in RequestUtils.


Molrn:
please set them all to merge to feature/roles, it's important to not create a chain of PRs. you can directly commit the small fix PRs to feature/roles, no need to validate everything (ex: #850, #851) at this point

pascalaubry:
> please set them all to merge to feature/roles, it's important to not create a chain of PRs. you can directly commit the small fix PRs to feature/roles, no need to validate everything (ex: #850, #851) at this point

Seen this afternoon: #850 and #851 were needed by #852, which was automatically rebased on ``feature/roles`` when #850 and #851 have been merged.

src/web/controllers/user/screen_user_controller.py
```diff
@@ -104,10 +105,14 @@ def _user_screen_refresh_needed(
                 return True
         return False
 
+    screen_guards = EventUserController.event_guards + [
```
timothyarmes, line 108: Would this not be nicer generalised as a super.guards?

src/web/controllers/user/screen_user_controller.py
```diff
@@ -146,12 +149,18 @@ async def htmx_user_screen_head(
     ) -> None:
         pass
 
+    rotator_guards = [
```
timothyarmes: Here you've forgotten to use the `EventUserController.event_guards`

src/web/controllers/user/screen_user_controller.py
```diff
@@ -188,11 +198,17 @@ async def htmx_user_rotator_head(
     ) -> None:
         pass
 
+    display_controller_guards = [
```
timothyarmes: Here you've forgotten to use the `EventUserController.event_guards`

tests/e2e/screens/test_single_screens.py
```diff
@@ -87,6 +87,21 @@ def test_check_in_screen(
 
         row.click()
         modal = lan_page.locator('.modal-dialog')
+        # The modal should not be visible since check-in is not open yet
+        expect(modal).not_to_be_visible()
```
timothyarmes: This doesn't test that the modal won't open unfortunately.  It tests that it isn't open in the instant after you click, but it takes time to open the modal, it might open after this test passes.

In test_spectator_role I had to do this:

```
        # We should NOT see the modal
        row = rows.filter(has_text='ALYX')
        row.click()
        modal = auth_page.locator('.modal-dialog')
        auth_page.wait_for_timeout(200)
        expect(modal).not_to_be_visible()
```

It's not nice, but I don't have a better solution for the moment

timothyarmes (dismissed):

tests/e2e/screens/test_single_screens.py
timothyarmes: Update: `board_row_player_cell.html` only includes the `hx-get` if the row is editable.  It would be much better to test that the `hx-get` isn't there rather than wait 200ms and hope that's a reasonable time

src/web/controllers/user/screen_user_controller.py
pascalaubry: Yes probably, but I am not sure of the class hierarchy we will have with the arbiter pages.

pascalaubry: I think we could also maybe set the class variable ``guards``, which applies to all the end-points of the controller.

PR #853, opened 2025-07-22
Restrain the visibility of passed/coming events to administrators

pascalaubry:
Spectators are now allowed to select only the public current events.


src/database/sqlite/event/event_store.py
```diff
@@ -275,6 +276,16 @@ class BaseStoredEvent:
...
+    def passed(self, now: float | None = None) -> bool:
+        return self.stop < (now or time.time())
+
+    def coming(self, now: float | None = None) -> bool:
+        return self.start > (now or time.time())
+
+    def current(self, now: float | None = None) -> bool:
+        now = now or time.time()
+        return not self.passed(now) and not self.coming(now)
+
```
Molrn: no methods in the store, that's why we've got a wrapper around the data object

Molrn (dismissed):

Molrn (dismissed):
LGTM

PR #858, opened 2025-07-23
Allow Performance tie-break for RR tournaments

pascalaubry:
Used by ED (cf mail).
image
image

### DIFF
--- a/src/data/tie_breaks/tie_breaks.py
+++ b/src/data/tie_breaks/tie_breaks.py
@@ -1077,9 +1077,7 @@ class AverageRatingOpponentsTieBreak(TieBreak):
 
 
 class PerformanceTieBreak(TieBreak, ABC):
-    @property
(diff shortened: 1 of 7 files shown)


src/data/tie_breaks/tie_breaks.py
```diff
@@ -1077,9 +1077,7 @@ def compute_player_value(
 
 
 class PerformanceTieBreak(TieBreak, ABC):
-    @property
-    def forbidden_pairing_systems(self) -> list[PairingSystem]:
-        return [RoundRobinPairingSystem()]
+    pass
```
Molrn:
```suggestion
    @property
    def forbidden_pairing_systems(self) -> list[PairingSystem]:
        # TODO (Molrn) Forbid for Round-Robin after big-move
        return []
```

Molrn (dismissed):

Molrn (approved):
LGTM

PR #861, opened 2025-07-25
Merge back 2.8.3

pascalaubry:


src/web/templates/common/base.html
```diff
@@ -173,6 +184,7 @@
 
                     // TODO we shouldn't do this for every modal
                     updateRecordByes();
+                    updateTieBreaksHelp();
```
Molrn, line 187: @pascalaubry mb, that's a miss from the review. Functions related to one modal should not be defined in base.html and executed for every modal.  
instead in `tournament_modal.html`:
```js
<script>
    $('#pairing-system').change(function() { [...] });
</script>
```
doing a PR to dev to apply this to `updateTieBreaksHelp` and `updateRecordByes`

Molrn (approved):
LGTM

PR #863, opened 2025-07-25
Remove modal specific functions from base.html

Molrn:


pascalaubry (approved):
Fine for me.
I don't merge if we want to change the content of the help texts (needs to be discussed).

timothyarmes:
I think we can merge now - that whole section will change anyway to allow for all the tie-break, more than 3 or them etc.

PR #867, opened 2025-07-27
Renamed Sonnenborn to Sonneborn

pascalaubry:


timothyarmes (approved):

pascalaubry:
Note: I left this PR draft at the moment because while testing this code I found a serious bug concerning tie-breaks (probably the same mentioned by Eric, I try to reproduce in a clean environment).

pascalaubry:
Here is how to reproduce the bug.
From the ``dev`` branch:
1. Start the server with a clean install
2. Save the default configuration
3. Create an event with the default settings
4. Create a tournament (**Create file**, 7 rounds, **Round-robin**, **Set the recommended tie-breaks**, create with no screen)
Now restart the server from branch ``bugfix/rename-sonneborn``:
1. Go to the previously created event
2. Edit the previously created tournament

Sonneborn-Berger disappeared from the tie-beaks.

Amaras:
I have had a bug like this: all my tie-breaks (Buchholz Cut-2 and Performance) disappeared at some point, and I had to set them up again (didn't disappear again since).
I have no clue why it happened

timothyarmes:
> Sonneborn-Berger disappeared from the tie-beaks.

That's because `static_papi_id` is the text that has to be stored in the Papi file's `DepX` field.  You have to keep the incorrect spelling.

When you set it on dev it'll be stored in Papi with the correct (badly spelt) name.  When you move to the new branch the mapping from the Papi name to the tiebreak no longer matches.

src/data/tie_breaks/tie_breaks.py
```diff
@@ -740,23 +740,23 @@ class SonnebornBergerTieBreak(TieBreak):
...
 
     @staticmethod
     def static_id() -> str:
-        return 'SONNENBORN_BERGER'
+        return 'SONNEBORN_BERGER'
 
     @staticmethod
     def static_papi_id() -> str:
-        return 'Sonnenborn-Berger'
+        return 'Sonneborn-Berger'
```
timothyarmes: You must keep the bad spelling here, it's what Papi expects to find in the database

timothyarmes (dismissed):

pascalaubry:
Note @Amaras: I also faced all the tie-breaks disappearing once, this issue must be tracked very closely when it happens again.

timothyarmes (approved):

timothyarmes (dismissed):
You need to rename the migration to m33

Amaras (approved):

PR #874, opened 2025-08-01
Avoine improvement

timothyarmes:


src/data/print_documents/documents.py
```diff
@@ -404,6 +415,33 @@ def static_name() -> str:
...
+    @property
+    def is_pairings_list(self) -> bool:
+        return True
+
+    @property
+    def template_context(self) -> dict[str, Any]:
+        context = {}
+        context.update(BoardPrintDocument.template_context.fget(self))
+        context.update(PlayerPrintDocument.template_context.fget(self))
+        return context
```
Molrn:
```suggestion
        return (
            BoardPrintDocument.template_context.fget(self) | 
            PlayerPrintDocument.template_context.fget(self)
        )
```
merge dict operator `|` was introduced to improve `.update()`'s lack of readability, its usage is prefered

src/data/print_documents/pairing_styles.py
Molrn, line 1: If these 2 are the only 2 styles we'll encounter, if feels a bit excessive to use this kind of code complexity to handle it. Having `PairingStylePrintOption` return a boolean and handling the only distinctive value (the template) in `PairingPrintDocument.template_context` with a simple if seems more than enough

Molrn (changes requested):
for the print document, I think a switch might be more appropriate than a select. It would simplify both the UI and the code.

Molrn (approved):

PR #881, opened 2025-08-04
ChessEvent support

timothyarmes:


src/database/sqlite/event/event_database.py
```diff
@@ -102,6 +113,10 @@ def upgrade(self):
     def event_database_path(uniq_id: str) -> Path:
         return EVENTS_DIR / f'{uniq_id}.{SharlyChessConfig.event_database_ext}'
 
+    @classmethod
+    def database_modified_timestamp(cls, uniq_id: str) -> float:
+        return cls.event_database_path(uniq_id).lstat().st_mtime
```
Molrn, line 118: as the usages are similar as `SQLiteDatabase.file_modified_at`, I think we should only keep one of the methods. If you could replace the usages of `SQLiteDatabase.file_modified_at` by `EventDatabase.database_modified_timestamp` it would be cleaner

src/plugins/chessevent/engine/action_selector.py
```diff
@@ -162,44 +140,35 @@ def write_chessevent_info_to_database(
...
+                'UPDATE `tournament` SET `chessevent_last_download_md5` = ?, '
+                '`last_update` = ? WHERE `id` = ?',
+                (
+                    chessevent_download_md5,
+                    time.time(),
+                    tournament.id,
+                ),
+            )
+
+            event_database.set_last_update()
```
Molrn: function was removed by #872

src/database/sqlite/event/event_database.py
```diff
@@ -59,6 +59,17 @@ def __init__(self, uniq_id: str, write: bool = False):
...
         super().__init__(self.event_database_path(self.uniq_id), write)
 
+    def __exit__(self, exc_type, exc_value, traceback):
+        from data.loader import EventLoader
+
+        super().__exit__(exc_type, exc_value, traceback)
+
+        # Inform EventLoader that this event has been updated internally
+        if self.write:
+            EventLoader.set_last_known_update(
```
Molrn, line 69: Really not a fan of adding an operation at each DB write, especially just for ChessEvent support... Is there a problem with what I suggested this morning ? 
```py
class EventLoader:
    def _load_event(self, uniq_id: str, reload: bool) -> Event:
        if event.check_update():
              return self.reload_event(uniq_id)
 ```
 It seems much simpler, and I don't see why it would not work, as tournament.last_update is updated by ChessEvent

File: src/plugins/chessevent/engine/action_selector.py [hunk 4]
```diff
@@ -23,133 +24,110 @@
...
+                    value=player.blitz_rating,
+                    type=player.blitz_rating_type,
+                ).stored_value,
+            },
+            fide_id=player.fide_id if player.fide_id else None,
+            federation=player.federation,
+            club=player.ffe_club,
+            fixed=player.board,
+            check_in=check_in_started and player.check_in,
+            plugin_data=ffe_plugin_data.to_stored_value(),
```
Molrn:
```suggestion
            plugin_data={'ffe': ffe_plugin_data.to_stored_value()},
```

src/plugins/chessevent/engine/action_selector.py
```diff
@@ -252,13 +217,13 @@ def run(self, event_uniq_id: str) -> bool:
...
-        create_answer: str = _('C *** THE LETTER TO ANSWER CREATE')
-        upload_answer: str = _('U *** THE LETTER TO ANSWER UPLOAD')
+        create_answer: str = _('R *** THE LETTER TO ANSWER REPLACE')
         quit_answer: str = _('Q *** THE LETTER TO ANSWER QUIT')
         default_answer: str = create_answer
         actions1: dict[str, str] = {
-            create_answer: _('Create the Papi files'),
-            upload_answer: _('Create the Papi files and send them to the FFE website'),
+            create_answer: _(
+                'Replace all players your tournaments with those from ChessEvent-'
```
Molrn:
```suggestion
                'Replace all the players in your tournaments with those from ChessEvent'
```

Molrn (changes requested):

Molrn (approved):
LGTM

PR #882, opened 2025-08-04
Use PapiConverter to convert player database

timothyarmes:


src/plugins/ffe/papi_converter.py
```diff
@@ -10,3 +12,22 @@ class PapiConverter:
     @property
     def executable_path(self) -> Path:
         return PapiConverterInstaller().executable_path
+
+    def convert_player_database(self, source_file: Path, target_file: Path) -> bool:
```
Molrn, line 16: type error (does not return bool)

Molrn (changes requested):

Molrn (approved):
LGTM

PR #888, opened 2025-08-05
Add titles to prize assignment

Amaras:


src/web/templates/admin/print/prizes.css
```diff
@@ -5,7 +5,7 @@
 
 .prize-table {
     display: grid;
-    grid-template-columns: max-content max-content auto max-content max-content max-content max-content auto max-content max-content;
+    grid-template-columns: max-content max-content auto auto max-content max-content max-content max-content auto max-content max-content;
```
pascalaubry, line 8: Twice auto?

timothyarmes (approved):

src/web/templates/admin/print/prizes.css
Amaras: No need to stretch the column I added too much.

PR #897, opened 2025-08-06
Remove results and illegal_move tables

timothyarmes:


src/database/sqlite/event/event_database.py
```diff
@@ -47,6 +44,7 @@
 
 if TYPE_CHECKING:
     from data.loader import EventBackup
+    from data.pairing import Pairing
```
Molrn: The Database class is meant to only handle stored Objects

src/database/sqlite/event/event_database.py
```diff
@@ -1784,6 +1782,7 @@ def _row_to_stored_pairing(cls, row: dict[str, Any]) -> StoredPairing:
             round_=row['round'],
             result=row['result'],
             board_id=row['board_id'],
+            illegal_moves=row.get('illegal_moves', 0),
```
Molrn:
```suggestion
            illegal_moves=row['illegal_moves'],
```
We need thiss to fail if there's an issue with the migration, it will fail at another layer which will be harder to realize otherwise

src/database/sqlite/event/event_database.py
```diff
@@ -1854,14 +1853,18 @@ def _row_to_stored_board(cls, row: dict[str, Any]) -> StoredBoard:
             white_player_id=row['white_player_id'],
             black_player_id=row['black_player_id'],
             index=row['index'],
+            last_result_update=row.get('last_result_update'),
```
Molrn:
```suggestion
            last_result_update=row['last_result_update'],
```

src/database/sqlite/event/event_database.py
```diff
@@ -1854,14 +1853,18 @@ def _row_to_stored_board(cls, row: dict[str, Any]) -> StoredBoard:
...
     def load_tournament_stored_boards_by_round(
         self, tournament_id: int
     ) -> dict[int, list[StoredBoard]]:
         self.execute(
             (
-                'SELECT `board`.*, `pairing`.`round` '
+                'SELECT `board`.`id`, `board`.`white_player_id`, `board`.`black_player_id`, '
+                '`board`.`index`, '
+                'datetime(`board`.`last_result_update`, "unixepoch") as last_result_update, '
+                '`pairing`.`round` '
```
Molrn:
```suggestion
                'SELECT `board`.*, `pairing`.`round` '
```
It's important to maintain the `.*` select to avoid maintenance issues. I'm all for putting datetimes in the DB, but if that means having all the `SELECT *` be replaced by the list of column it's not worth it (would be a nightmare to maintain for events or tournaments)

src/data/tournament.py
Molrn, line 1: Tournament.add_result and remove_result now make no sense, there should only be one `update_result`

src/database/sqlite/event/event_database.py
Molrn, line 1: It's important to keep as little logic as possible in SQL queries as it is harder to maintain.  
Similar to the other usages, most of these functions should be included in the wrapper, and the illegal_moves and last_result_update properties should be added as @property to the Board and Pairing classes

- get_stored_illegal_moves
  - move the logic to tournament.get_illegal_moves (using tournament.get_round_boards)
- add_stored_illegal_move / delete_stored_illegal_moves
  - Pairing.increment/decrement_illegal_moves
  - uses EventDatabase.update_stored_pairing
- update_board_result_from_pairing
  - Board.set_last_result_update(new_result: Result, event_database: EventDatabase)
  - uses EventDatabase.update_stored_board
- get_stored_results
  - Move the logic to Screen._result

Molrn (changes requested):
the DataResult class concretly has all the properties of a Board, its usages have to be replaced and the class deleted

src/data/screen.py
```diff
@@ -655,17 +656,26 @@ def results_tournament_names(self) -> str:
...
+        boards: list[Board] = []
+        oldest = time.time() - self.results_max_age * 60
+        for tournament in self.event.tournaments_by_id.values():
+            if not self.results_tournament_ids or tournament.id in self.results_tournament_ids:
+                for round_ in range( 1, tournament.current_round + 1):
+                    for player in tournament.players_by_id.values():
+                        pairing = player.pairings[round_]
+                        if pairing.board and pairing.board.white_player.id == player.id:
+                            if pairing.board.last_result_update and pairing.board.last_result_update >= oldest:
+                                boards.append(pairing.board)
```
Molrn:
```suggestion
            if self.results_tournament_ids and tournament.id in self.results_tournament_ids:
                continue    
            for board in tournament.boards_by_id.values():
                if board.last_result_update and board.last_result_update >= oldest:
                    boards.append(pairing.board)
```

src/database/sqlite/event/event_database.py
```diff
@@ -1906,198 +1904,29 @@ def update_stored_board(self, stored_board: StoredBoard):
...
-    @staticmethod
-    def _row_to_stored_result(row: dict[str, Any]) -> StoredResult:
-        return StoredResult(
-            id=row['id'],
-            tournament_id=row['tournament_id'],
-            board_id=row['board_id'],
-            result=row['result'],
-            date=row['date'],
-        )
+    def update_board_last_result_update(self, board_id: int, clear: bool = False) -> float | None:
```
Molrn, line 1907: Move to # Board section + delete # Result section

src/web/templates/user/screen/boards/set.html
```diff
@@ -79,10 +79,10 @@ <h2 class="set-title text-center">
                                                 [{{ wp.points_str }}]
                                             </td>
                                         {% endif %}
-                                        {% with player=wp %}{% with opponent=bp %}
+                                        {% with player=wp %}{% with opponent=bp %}{% with pairing=board.white_pairing %}
```
Molrn:
```suggestion
                                        {% with player=wp, opponent=bp, pairing=board.white_pairing %}
```

src/web/templates/user/screen/boards/set.html
```diff
@@ -79,10 +79,10 @@ <h2 class="set-title text-center">
...
                                             {% include 'board_row_illegal_moves_cell.html' %}
                                             {% include 'board_row_player_cell.html' %}
-                                        {% endwith %}{% endwith %}
+                                        {% endwith %}{% endwith %}}{% endwith %}
```
Molrn:
```suggestion
                                        {% endwith %}
```

src/web/templates/user/screen/boards/set.html
```diff
@@ -96,10 +96,10 @@ <h2 class="set-title text-center">
                                         >
                                             {% if board.result_str %}{{ board.result_str }}{% else %}{{ _('#%(board_number)d', board_number=board.number) }}{% endif %}
                                         </td>
-                                        {% with player=bp %}{% with opponent=wp %}
+                                        {% with player=bp %}{% with opponent=wp %}{% with pairing=board.black_pairing %}
```
Molrn:
```suggestion
                                        {% with player=bp, opponent=wp, pairing=board.black_pairing %}
```

src/web/templates/user/screen/boards/set.html
```diff
@@ -96,10 +96,10 @@ <h2 class="set-title text-center">
...
                                             {% include 'board_row_illegal_moves_cell.html' %}
                                             {% include 'board_row_player_cell.html' %}
-                                        {% endwith %}{% endwith %}
+                                        {% endwith %}{% endwith %}}{% endwith %}
```
Molrn:
```suggestion
                                        {% endwith %}
```

src/web/templates/user/screen/results.html
```diff
@@ -18,13 +18,13 @@ <h2 class="set-title text-center">{{ screen.name }}</h2>
                 <tbody>
                 {% for result in results %}
                     <tr class="result-row">
-                        <td scope="row" class="time bg-transparent text-center">{{ result.timestamp_str }}</td>
-                        {% with tournament=user_event.tournaments_by_id[result.tournament_id] %}
+                        <td scope="row" class="time bg-transparent text-center">{{ result.last_result_update_str }}</td>
+                        {% with tournament=user_event.tournaments_by_id[result.tournament.id] %}
```
Molrn:
```suggestion
                        {% with tournament=result.tournament %}
```

src/web/templates/user/screen/results.html
```diff
@@ -18,13 +18,13 @@ <h2 class="set-title text-center">{{ screen.name }}</h2>
                 <tbody>
                 {% for result in results %}
```
Molrn:
```suggestion
                {% for board in result_boards %}
```
Would also need a rename

src/web/templates/user/screen/results.html
```diff
@@ -18,13 +18,13 @@ <h2 class="set-title text-center">{{ screen.name }}</h2>
...
                     <tr class="result-row">
-                        <td scope="row" class="time bg-transparent text-center">{{ result.timestamp_str }}</td>
-                        {% with tournament=user_event.tournaments_by_id[result.tournament_id] %}
+                        <td scope="row" class="time bg-transparent text-center">{{ result.last_result_update_str }}</td>
+                        {% with tournament=user_event.tournaments_by_id[result.tournament.id] %}
                         {% with players_by_id=tournament.players_by_id %}
-                        {% with white_player=players_by_id[result.white_player_id] %}
-                        {% with black_player=players_by_id[result.black_player_id] %}
+                        {% with white_player=players_by_id[result.white_player.id] %}
+                        {% with black_player=players_by_id[result.black_player.id] %}
```
Molrn:
```suggestion
                        {% with white_player=result.white_player %}
                        {% with black_player=result.black_player %}
```

src/web/templates/user/screen/results.html
```diff
@@ -18,13 +18,13 @@ <h2 class="set-title text-center">{{ screen.name }}</h2>
...
                         {% if print_tournament %}<td scope="row" class="tournament bg-transparent">{{ tournament.uniq_id }}</td>{% endif %}
-                        <td scope="row" class="board-number bg-transparent text-center">{{ result.round }}.{{ result.board_id }}</td>
+                        <td scope="row" class="board-number bg-transparent text-center">{{ result.round }}.{{ result.id }}</td>
```
Molrn:
```suggestion
                        <td scope="row" class="board-number bg-transparent text-center">{{ result.round }}.{{ result.index }}</td>
```
`Board.id` is legacy, as it does not refer to the Database ID as it would be expected

src/data/board.py
```diff
@@ -120,6 +128,12 @@ def permute_colors(self):
                 database.update_stored_pairing(self.black_pairing.stored_pairing)
             database.commit()
 
+    def update_board_result_from_pairing(self, pairing: 'Pairing', database: EventDatabase):
+        """Updates board timestamp. Clears board timestamp if result is NO_RESULT."""
+
+        if pairing.board:
+            pairing.board.stored_board.last_result_update = database.update_board_last_result_update(pairing.board.id, clear=pairing.result == Result.NO_RESULT)
```
Molrn:
```suggestion
    def set_last_result_update(self, new_result: Result, database: EventDatabase):
        """Updates board timestamp. Clears board timestamp if result is NO_RESULT."""
        self.stored_board.last_result_update = database.update_board_last_result_update(
            self.identifier, clear=new_result == Result.NO_RESULT
        )
```
If that's a `Board` method, it should take effect on this board (i.e. self), not the one in the pairing.
It just needs the result as param, so let it just be that

src/data/board.py
```diff
@@ -91,6 +91,14 @@ def no_result(self) -> bool:
     def result_str(self) -> str:
         return str(self.result)
 
+    @property
+    def last_result_update(self) -> float | None:
+        return self.stored_board.last_result_update
+
+    @property
+    def last_result_update_str(self) -> str:
+        return format_timestamp(self.stored_board.last_result_update) if self.stored_board.last_result_update else ''
```
Molrn:
```suggestion
        return format_timestamp(self.last_result_update) if self.last_result_update else ''
```
calls to the stored objects are only necessary for setting them (self.stored_board.last_result_update = None), otherwise that's why the property is there

src/data/pairing.py
```diff
@@ -59,6 +66,23 @@ def update(self, event_database: EventDatabase):
             event_database.add_stored_pairing(self.stored_pairing)
             self.exists = True
 
+    def add_illegal_move(self, event_database: EventDatabase):
+        if self.stored_pairing.illegal_moves < self.player.tournament.record_illegal_moves:
```
Molrn:
```suggestion
        if self.illegal_moves < self.player.tournament.record_illegal_moves:
```

src/data/pairing.py
```diff
@@ -59,6 +66,23 @@ def update(self, event_database: EventDatabase):
...
+            self.stored_pairing.illegal_moves += 1
+            self.update(event_database)
+            logger.info('An illegal move has been recorded for player [%s].', self.player.id)
+            return True
+
+    def delete_illegal_move(self, event_database: EventDatabase):
+        if self.stored_pairing.illegal_moves > 0:
```
Molrn:
```suggestion
        if self.illegal_moves > 0:
```

Molrn (changes requested):
image

couple visual comments modifications: 
- `}}}}` before the table
- Once illegal move has been modified, player always stays highlighted in red
- the table head `Blancs` / `Noirs` not being aligned with the player name feels weird

Molrn (approved):
LGTM

PR #923, opened 2025-08-10
Use sce for the examples format

timothyarmes:


Molrn:
discussed this morning: requires #924 and m035 migrated .sce files

PR #927, opened 2025-08-11
Big move

Molrn:


timothyarmes (dismissed):
I'm happy.

Molrn (commented):
Good for me too ;)

src/data/input_output/dict_reader.py
```diff
@@ -0,0 +1,130 @@
...
+    """Method converting a dictionary into a dataclass object.
+    Can recursively convert the sub dictionaries into other dataclasses.
+
+    Raises a DictReaderException if:.
+        - A required (i.e. without default value) field is missing
+        - An unknown field is encountered
+        - a value is of incorrect type"""
+
+    if not is_dataclass(data_class):
+        raise ValueError(f'{data_class} is not a dataclass')
```
Amaras: Sounds like it should be a `TypeError`, rather than `ValueError`

src/data/event.py
```diff
@@ -614,6 +514,99 @@ def clear_player_cache(self):
...
+                counter[club] += tournament.club_counts[club]
+        return counter
+
+    @cached_property
+    def check_in_counts(self) -> Counter[bool | None]:
+        counter: Counter[bool | None] = Counter[bool | None]()
+        for tournament in self.tournaments_by_id.values():
+            for check_in in tournament.players_by_check_in_status:
+                counter[check_in] += tournament.check_in_counts[check_in]
+        return counter
```
Amaras, line 608: Those should be SQL queries

src/database/sqlite/event/event_database.py
```diff
@@ -1492,35 +811,38 @@ def update_stored_tournament(
     def delete_stored_tournament(self, tournament_id: int):
         self.execute('DELETE FROM `tournament` WHERE `id` = ?;', (tournament_id,))
 
-    def _set_tournament_last_illegal_move_update(self, tournament_id: int):
+    def _set_tournament_timestamp_field(self, field_: str, tournament_id: int) -> float:
+        # TODO (Molrn) replace all these usages with the appropriate SQL triggers
+        timestamp = time.time()
         self.execute(
-            'UPDATE `tournament` SET `last_illegal_move_update` = ? WHERE `id` = ?',
+            f'UPDATE `tournament` SET `{field_}` = ? WHERE `id` = ?',
```
Amaras, line 820: This is increasing the SQL injection attack surface.
I don't know the whole call tree, so this might be overly pessimistic

src/plugins/ffe/ffe_sql_server.py
```diff
@@ -197,14 +205,14 @@ async def search_player(
         params: list[Any] = []
         for token in tokens:
             token_expressions: list[str] = [
-                f'(UPPER({field[0]}) LIKE ?)' for field in str_fields
+                f'(UPPER({field[0]}) LIKE %s)' for field in str_fields
```
Amaras, line 208: That's not the standard SQL prepared statement syntax.
What gives?
It feels like it opens the door to SQL injection, and I can't allow that.

Amaras (changes requested):

src/plugins/ffe/ffe_sql_server.py
timothyarmes: It's the python-tds syntax: https://python-tds.readthedocs.io/en/latest/

Amaras (approved):
Alright for now, there definitely needs to be a trigger added

PR #930, opened 2025-08-11
Remove i18n log messages

pascalaubry:


Molrn:
Why ? this message is important to have in french, like it's not some random log thing, it's an important message for users to know something went wrong and that they can do something manually to try it again

pascalaubry:
> Why ? this message is important to have in french, like it's not some random log thing, it's an important message for users to know something went wrong and that they can do something manually to try it again

For support: we agreed at the very beginning of the project that everything should be i18ned on the UI but the logs must be in English for us not to have to read all the languages.
If you really want these messages translated, I think you should use methods `print_interactive_xxx()` (but as everything should be run from the UI at the end I am not sure it is a good idea).

Molrn (dismissed):
LGTM

PR #934, opened 2025-08-12
Git workflow with mac build

timothyarmes:


.github/workflows/export.yml
```diff
@@ -45,18 +70,119 @@ jobs:
...
+          rm -rf dist/
+          rm -rf build/
+          rm -f *.spec
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: sharly-chess-${{ matrix.platform }}-${{ steps.version.outputs.version }}
+          path: |
+            export/sharly-chess-${{ steps.version.outputs.version }}.zip
+            export/papi-web-${{ steps.version.outputs.version }}.zip
```
pascalaubry: The creation of papi-web-<version>.zip from `export.py` has been removed in 2.8 (removing this line had been forgotten at that time).
```suggestion
```

pascalaubry (changes requested):
Can not test so trusting you for the Mac part.

PR #943, opened 2025-08-13
Solve check-in and pairing issues

Amaras:


tests/e2e/screens/test_single_screens.py
```diff
@@ -87,9 +87,22 @@ def test_check_in_screen(
         row = rows.filter(has_text='ALYX')
         expect(row.locator('i.bi-square')).to_be_visible()
 
+        row.click()
+        modal = lan_page.locator('.modal-dialog')
+        expect(modal).to_be_hidden()
```
timothyarmes: This test is wrong.  It'll always be closed at the instant after you click - it takes time to open.  You should instead use the solution used in the role branch - example:

```py
expect(result_cell).not_to_have_attribute(
                'hx-get', re.compile(r'.*result-modal.*')
            )
```

tests/e2e/screens/test_single_screens.py
```diff
@@ -87,9 +87,22 @@ def test_check_in_screen(
...
+
+        api_request_context.patch(
+            f'/admin/tournament-open-check-in/{EVENT_ID}/{unpaired_tournament.id}'
+        )
+
+        # NOTE(Amaras): because the check-in was updated, we either need to refresh the page, or wait
+        # until it refreshes itself
+        lan_page.reload(wait_until='commit')
```
timothyarmes: Don't need do to this.  The tests are configured for refreshing pages to refresh very quickly, so waiting isn't costly and it tests that the refresh really is working.

timothyarmes (changes requested):

timothyarmes (approved):

PR #954, opened 2025-08-17
Remove event messages

Molrn:


timothyarmes (dismissed):

src/data/screen_set.py
```diff
@@ -87,13 +80,6 @@ def __init__(
                     self.last = max(1, self.stored_screen_set.last)
                     if self.first:
                         self.last = max(self.first, self.last)
-                if self.first and self.last and self.first > self.last:
-                    self.event.add_warning(
-                        _('Invalid range [{first}-{last}].').format(
-                            first=self.first, last=self.last
-                        ),
-                        screen_set=self,
-                    )
```
pascalaubry, line 96: IMO This should be logged.

pascalaubry (dismissed):
Please also remove class EventMessage.

src/data/screen_set.py
Molrn: it is already validated by the form, you can't input a first higher than a last

pascalaubry (approved):

PR #960, opened 2025-08-18
Fix papi import with string ffe_id

Molrn:


src/plugins/ffe/papi_converter.py
```diff
@@ -418,10 +422,22 @@ def raise_unknown_value(field_: str, value: Any):
...
-            }
+        ffe_id = None
+        if variables.homologation:
+            if not variables.homologation.isdigit():
+                # Papi form allows for a non-integer value,
+                # so the value is ignored instead of raising
+                logger.warning(
+                    f'Homologation number [{variables.homologation}] '
+                    'is not an integer, its value is ignored.'
+                )
```
pascalaubry:
```suggestion
                logger.warning(
                    'Homologation number [%s] '
                    'is not an integer, its value is ignored.',
                    variables.homologation
                )
```

pascalaubry (changes requested):
Just a detail.

pascalaubry (approved):

PR #991, opened 2025-08-22
Generate i18n.md only with i18n_update --generate-doc

pascalaubry:
bugfix: Update the MO files when PO files only have changed
improvement: Generate i18n.md only with i18n_update --generate-doc


timothyarmes (approved):
Seems like a lot of code changes to remove a header :). I'm just trusting you

PR #992, opened 2025-08-22
Improved version recovering

pascalaubry:
We now try to recover previous versions by looking for `../sharly-chess-{version}-windows` (in addition to `../papi-web-{version}` and `../sharly-chess-{version}`) since unzipping `sharly-chess-{version}-windows.zip` (the Windows bundle since 3.0.0b1) will by default create folder `sharly-chess-{version}-windows` (note: when users upgrade from the CLI, `sharly-chess-{version}-windows.zip` is unzipped to folder `sharly-chess-{version}`).


scripts/export/export.py
```diff
@@ -54,7 +54,7 @@
 basename: str = f'sharly-chess-{SHARLY_CHESS_VERSION}'
 EXPORT_DIR: Path = BASE_DIR / 'export'
 PROJECT_DIR: Path = DIST_DIR / basename
-ZIP_FILE: Path = EXPORT_DIR / f'{basename}.zip'
```
timothyarmes: I really don't like this.  export.py is meant to be as generic is possible. If I run this script on my Mac I now get a Mac version in a zip called xxx-windows.  Why do you have to change the name here?
If there's a technical reason, then at least only do it on the windows version.

We'll have linux one day too, it lmay be able to use the zip file.

timothyarmes (dismissed):

src/common/engine.py
```diff
@@ -123,14 +123,29 @@ def __init__(self):
...
+                    version_dir.name,
+                ):
+                    # 2.7.1 <= version
+                    dir_name = matches.group(0)
+                    version: Version = Version(matches.group(1))
+                elif matches := re.match(
+                    r'^sharly-chess-(\d+.\d+.\d+(?:a\d+|b\d+|rc\d+)?)-windows$',
+                    version_dir.name,
+                ):
+                    # 3.0.0 <= version
```
timothyarmes: Looks like you can just to this, no?

```py
dir_name: str
version: Version
pattern = r'^(?:papi-web|sharly-chess)-(\d+\.\d+\.\d+(?:a\d+|b\d+|rc\d+)?)(?:-windows)?$'
if matches := re.match(pattern, version_dir.name):
    dir_name = matches.group(0)
    version: Version = Version(matches.group(1))
else:
    logger.debug('Not a release: [%s].', version_dir)
    continue
```

timothyarmes (dismissed):

timothyarmes (approved):

PR #994, opened 2025-08-22
Allow to reimport into the same tournament

Molrn:


pascalaubry:
Works fine except the 2 missing strings.

src/data/input_output/tournament_importers.py
```diff
@@ -48,9 +47,9 @@ def load_tournament(
         """Load a tournament into an event.
         If tournament is provided, update this tournament, otherwise create a new one.
         Raises if the tournament already has players."""
-        if tournament and tournament.players:
+        if tournament and tournament.started:
             raise SharlyChessException(
-                _('Impossible to import into a tournament already having players.')
+                'Impossible to import into a started tournament.'
```
timothyarmes: Import only possible before the tournament starts.

src/database/sqlite/event/event_database.py
```diff
@@ -1050,6 +1040,27 @@ def delete_stored_tournament_player(self, tournament_id: int, player_id: int):
             (tournament_id, player_id),
         )
 
+    def delete_stored_tournament_players(self, tournament_id: int):
```
timothyarmes: I don't know if this is possible in this case since we're not deleting the tournament, but it feels like the operations here should be part of an AFTER DELETE ON trigger.  I have no suggestions for doing that without code duplication though :(

timothyarmes (dismissed):

src/database/sqlite/event/migrations/m038_create_tournament_player_on_delete_triggers.py
```diff
@@ -0,0 +1,51 @@
...
+    def forward(self):
+        self.database.execute(
+            """
+            CREATE TRIGGER IF NOT EXISTS
+                `delete_player_on_tournament_player_delete`
+            AFTER DELETE ON `tournament_player`
+            BEGIN
+                DELETE FROM `player`
+                WHERE `id` = `OLD`.`player_id` AND (
+                    SELECT COUNT(*) FROM `tournament_player`
```
timothyarmes, line 17: A comment to explain what's going on here would be nice.

timothyarmes (dismissed):

timothyarmes (approved):

PR #995, opened 2025-08-23
Manual tiebreaks  (closes #740)

### LINKED ISSUE #740: Add a manual tie-break option
For tournaments where some part of the tournament is able to advance to OTB tie-breaks (e.g. the possibility is there for qualifiers to the French Youth Championships), there is a need to say "this player is ahead of those other ones" depending on the results of the tie-breaks.

This is currently not available at all.

There needs to be a menu that allows arbiters to enter the values, maybe also a way to specify if the tie-break is ascending (position in the tie-break bracket) or descending (number of tie-break matches won).

timothyarmes:


Molrn (dismissed):
The interface feels a bit buggy:
- you can drag and drop a player into another group. It then snaps right back to it's original position without explanation
- the white of the background feels out of place, it should have the base color of the page
- it is unclear what the boundaries of the sort group are

## Suggestion
a rectangle box around each group, which appears when an element of the group is dragged

Molrn (approved):
LGTM

PR #997, opened 2025-08-23
Pairings search (closes #823)

### LINKED ISSUE #823: Pairings search bar
Add a search bar to the pairings tab. it would at least allow to search boards by player names, I was thinking integer values could search for table numbers

timothyarmes:


pascalaubry (approved):
Thought you requested the server for that but it is a nice way to do it on the client (limit: `tim arm` not found).

PR #1009, opened 2025-08-25
Correctly show real points

Amaras:
Currently, reals points are only shown and printed if the tournament's
current round is a round where the real points should be printed.
This is incorrect in the case where you want to check the pairings of
e.g. round 2 during the last round of the tournament.

This commit makes sure the real points are computed at the correct time.


timothyarmes (approved):
LGTM

PR #1015, opened 2025-08-27
Allow plugins to alter the reference year for category calculations (closes #1014)

### LINKED ISSUE #1014: Categories needs federation plugin hooks
Currently the categories displayed are based on the age before the 1st September.

This is clearly France specific, and suggests that they should be calculated using a hook.

timothyarmes:


Molrn (approved):
LGTM

PR #1017, opened 2025-08-27
Sanitize and hide the uniq_id fields

Molrn:

PR #1024, opened 2025-08-28
Fix player movement

### LINKED ISSUE #1020: [BUG] Exception when moving a player from one tournament to another one
## Bug description
When changing the tournament of a player, the player is grayed and nothing happens. When reloading the player tab, the player has been deleted.

The following exception is raised:
```
Uncaught exception (connection_type=http, path=/admin/player-move/domloup-fide-40/192/1):
Traceback (most recent call last):
  File "C:\OneDrive\echecs\sharly-chess\dev\.venv-dev\Lib\site-packages\litestar\middleware\_internal\exceptions\middleware.py", line 158, in __call__
    await self.app(scope, receive, capture_response_started)
  File "C:\OneDrive\echecs\sharly-chess\dev\.venv-dev\Lib\site-packages\litestar\routes\http.py", line 81, in handle
    response = await self._get_response_for_request(
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\OneDrive\echecs\sharly-chess\dev\.venv-dev\Lib\site-packages\litestar\routes\http.py", line 133, in _get_response_for_request
    return await self._call_handler_function(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\OneDrive\echecs\sharly-chess\dev\.venv-dev\Lib\site-packages\litestar\routes\http.py", line 181, in _call_handler_function
    else await route_handler.fn(**parsed_kwargs)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\OneDrive\echecs\sharly-chess\dev\src\web\controllers\admin\player_admin_controller.py", line 1232, in htmx_admin_player_move
    dst_tournament.add_player_to_tournament(admin_player.stored_player)
  File "C:\OneDrive\echecs\sharly-chess\dev\src\data\tournament.py", line 1112, in add_player_to_tournament
    database.add_stored_tournament_player(stored_tournament_player)
  File "C:\OneDrive\echecs\sharly-chess\dev\src\database\sqlite\event\event_database.py", line 1007, in add_stored_tournament_player
    self.execute(
  File "C:\OneDrive\echecs\sharly-chess\dev\src\database\sqlite\sqlite_database.py", line 82, in execute
    self.cursor.execute(query, params)
sqlite3.IntegrityError: FOREIGN KEY constraint failed
```

## How to reproduce
Steps to reproduce the bug:
1. Go to the Players tab
2. Click on change the tournament of a player
3. Reload the page

## Environment
- OS: Windows 11
- Sharly Chess version: dev branch
- event to reproduce on Discord

Amaras:
Before this commit, tournament players were deleted from the source
tournament before being added to the destination tournament.
However, when tournament players are deleted, the player is also
deleted.
This means that the player is not moved, unless they are in at least two
tournaments, which is a rare use case.

This commit fixes that by adding an event-level function to move players
between tournament

### DIFF
--- a/src/web/controllers/admin/player_admin_controller.py
+++ b/src/web/controllers/admin/player_admin_controller.py
@@ -1226,10 +1226,12 @@ class PlayerAdminController(BaseEventAdminController):
         admin_player = web_context.get_admin_player()
         dst_tournament = web_context.get_admin_tournament()
         src_tournament = admin_player.tournament
+        admin_event = web_context.get_admin_event()
         try:
             self._validate_player_tournament_move(admin_player, dst_tournament)
-            src_tournament.delete_player_from_tournament(admin_player.id)
-            dst_tournament.add_player_to_tournament(admin_player.stored_player)
+            admin_event.move_player_to_tournament(
+                admin_player, src_tournament, dst_tournament
+            )
             if not self.filtered_players(request, event_uniq_id, [admin_player]):
                 self.delete_from_search_results(request, admin_player.id)
             Message.success(
--- a/src/data/event.py
+++ b/src/data/event.py
@@ -495,6 +495,25 @@ class Event:
                 counter[check_in] += tournament.check_in_counts[check_in]
         return counter
 
+    def move_player_to_tournament(
+        self,
+        player: Player,
+        source_tournament: Tournament,
+        destination_tournament: Tournament,
+    ):
+        """Moves the given player from tournament *source_tournament* to *destination_tournament*."""
+        stored_tournament_player = player.stored_tournament_player
+        with EventDatabase(self.uniq_id, write=True) as database:
+            database.move_tournament_player(
+                stored_tournament_player, destination_tournament.id
+            )
+            del source_tournament.players_by_id[player.id]
+            player.stored_tournament_player = (
+                database.load_player_stored_tournament_player(player.id)
+            )
+            destination_tournament.players_by_id[player.id] = player
+            database.commit()
+
     def get_unused_tournament_uniq_id(
         self,
         base_uniq_id: str | None = None,
--- a/src/database/sqlite/event/event_database.py
+++ b/src/database/sqlite/event/event_database.py
@@ -1026,6 +1026,27 @@ class EventDatabase(MigrationDatabase):
             ),
         )
 
+    def move_tournament_player(
+        self,
+        stored_tournament_player: StoredTournamentPlayer,
+        destination_tournament_id: int,
+    ):
+        params = {
+            'new_tournament_id': destination_tournament_id,
+            'old_tournament_id': stored_tournament_player.tournament_id,
+            'player_id': stored_tournament_player.player_id,
+        }
+        self.execute(
+            'UPDATE `tournament_player` '
+            'SET `tournament_id` = :new_tournament_id '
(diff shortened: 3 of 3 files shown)


pascalaubry (approved):
Works fine!

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -1226,10 +1226,12 @@ async def htmx_admin_player_move(
...
         dst_tournament = web_context.get_admin_tournament()
         src_tournament = admin_player.tournament
+        admin_event = web_context.get_admin_event()
         try:
             self._validate_player_tournament_move(admin_player, dst_tournament)
-            src_tournament.delete_player_from_tournament(admin_player.id)
-            dst_tournament.add_player_to_tournament(admin_player.stored_player)
+            admin_event.move_player_to_tournament(
+                admin_player, src_tournament, dst_tournament
+            )
```
Molrn, line 1234:
```suggestion
            dst_tournament.add_player_to_tournament(admin_player.stored_player)
            src_tournament.delete_player_from_tournament(admin_player.id)
```
This is all that is needed, we need a fresh tournament_player record when changing a player's tournament, because the values other than `tournament_id` will differ. For now, it means the player keeps its pairing number, but from what is planned tournament_player will contain all the player values specific to the tournament.

Molrn (changes requested):

Molrn:
Already applied to #1017

src/web/controllers/admin/player_admin_controller.py
Amaras: There is a subtle problem with your approach, in that it can duplicate players if something goes wrong, while it doesn't in mine.
I don't see why moving players between tournaments needs to keep being a tournament-level operation instead of an event-level operation, since all the values are inside the event database

Molrn: how could the player be duplicated ? all it does is delete the tournament_player that exists and create a new one, I don't how it would be more dangerous than yours. Anything that handles `player` is at event level, but this handles `tournament_player`, which is something at tournament level

timothyarmes: you could argue that the controller shouldn't be responsible for knowing _how_ to move players, that should be an atomic operation in EventDatabase.

Molrn:
```py
            self._validate_player_tournament_move(admin_player, dst_tournament)
            dst_tournament.add_player_to_tournament(admin_player.stored_player)
            src_tournament.delete_player_from_tournament(admin_player.id)
```

If you wanna put these lines into an `Event.move_player` be my guest. not sure it's worth arguing about as moving player will most likely disappear with multi-tournament, the important part was to fix the bug which is done

Amaras: > how could the player be duplicated ? all it does is delete the tournament_player that exists and create a new one, I don't how it would be more dangerous than yours. Anything that handles `player` is at event level, but this handles `tournament_player`, which is something at tournament level

Add a player to the destination tournament, crash before deleting from the source tournament, now the player is assigned to two tournaments

PR #1045, opened 2025-08-29
Add a migration test

pascalaubry:


Molrn (dismissed):
I think we'd better not overload the test, recovering one event per version you want to test already seems enough IMO.

Molrn (approved):
LGTM

PR #1051, opened 2025-08-30
Clean cached Event unique IDs if file does not exist

### LINKED ISSUE #1050: [BUG] Server restart needed on event deletion/renaming
## Bug description
Users are not supposed to delete or rename the event databases but when they do an exception is raised and the server has to be restarted to reach the event on renaming.

Image

Molrn:
Closes #1050

### DIFF
--- a/src/data/loader.py
+++ b/src/data/loader.py
@@ -67,11 +67,11 @@ class EventLoader:
 
     @classmethod
     def _clean_not_existing_event_database_files(cls, event_uniq_ids: list[str]):
-        to_remove = [
+        to_remove = (
             uniq_id
             for uniq_id in event_uniq_ids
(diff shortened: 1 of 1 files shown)


src/data/loader.py
```diff
@@ -63,6 +65,16 @@ def load_event_ids(cls, uniq_id: str | None = None):
                 logger.error(e)
                 cls._invalid_uniq_ids.append(event_id)
 
+    @classmethod
+    def _clean_not_existing_event_database_files(cls, event_uniq_ids: list[str]):
+        to_remove = [
```
Amaras: If you're iterating once, use a generator expression rather than a list comprehension

Amaras (approved):

PR #1052, opened 2025-08-30
Use a single transaction to move player to another tournament

Molrn:


Amaras (approved):
I would have preferred an update query and a pairing delete query, rather than three queries for the same thing, but this version is good

PR #1062, opened 2025-08-31
Use sensible default for the print view (closes #1056)

### LINKED ISSUE #1056: preset print modal variable depending on the user context
Most of the times, it is easy to predict was the users want to print depending on the tab and its context. Setting the document type, the selected tournament and the selected round when opening the print modal would save a few clicks and really improve the UX.

- From the Pairings tab, document_type=pairings if the tournament is not finished otherwise ranking.
- From the Players tab, document_type=player-checkin-list if Check-in is open otherwise player-list
- From the Prizes tab, document_type=prize-assignment if the tournament is finished otherwise prize-list
- ...

timothyarmes:


src/web/controllers/admin/event_print_controller.py
```diff
@@ -51,7 +53,8 @@ def _print_modal_context(
                 tournament_id = event.tournaments_sorted_by_uniq_id[0].id
             data = {
                 'tournament_id': WebContext.value_to_form_data(tournament_id),
-                'document': PlayerListPrintDocument.static_id(),
+                'document': document_id or PlayerListPrintDocument.static_id(),
+                'round': WebContext.value_to_form_data(_round) or '',
```
pascalaubry:
```suggestion
                'round': WebContext.value_to_form_data(_round),
```

src/web/controllers/admin/pairings_admin_controller.py
```diff
@@ -216,12 +220,23 @@ def __init__(
...
             allowed_actions = permission_handler.allowed_actions(
                 self.round_status, self.safety_mode
             )
             existing_actions = permission_handler.existing_actions(self.round_status)
+
+            if (
+                self.display_rankings
+                or self.admin_round < self.admin_tournament.current_round
+                or self.admin_tournament.is_round_finished(self.admin_round)
+            ):
```
pascalaubry, line 237:
```suggestion
            if self.display_rankings:
```

src/web/controllers/admin/player_admin_controller.py
```diff
@@ -87,6 +91,20 @@ def __init__(
...
                 return
 
+        print_tournament_id = self.default_tournament_for_print_modal(
+            tournament_id=None
+        )
+        if print_tournament_id is None:
+            tournaments = list(self.admin_event.tournaments_by_id.values())
+        else:
+            tournaments = [self.admin_event.tournaments_by_id[print_tournament_id]]
+        check_in_open = all(tournament.check_in_open for tournament in tournaments)
```
pascalaubry, line 101: I would like to get `PlayerCheckinListPrintDocument` selected if **any** tournament has check-in open.
(no suggestion, we must discuss it first)

src/web/controllers/admin/pairings_admin_controller.py
```diff
@@ -216,12 +220,23 @@ def __init__(
     def template_context(self) -> dict[str, Any]:
         allowed_actions = []
         existing_actions = []
+
+        default_print_document = PairingPrintDocument.static_id()
```
pascalaubry, line 224: Shouldn't we print the results by default when the current round is finished?

Molrn (dismissed):
Round and document work well, the tournament does not on my end (on both prize and pairings tab)

pascalaubry (commented):
We probably need to talk about the default values (#1056 miss some details).

src/web/controllers/admin/pairings_admin_controller.py
timothyarmes: It does. Try it

timothyarmes:
> Round and document work well, the tournament does not on my end (on both prize and pairings tab)

I don't have any issues, can you investigate?

src/web/controllers/admin/pairings_admin_controller.py
pascalaubry: It displays the ranking, not results. I think that the results should be displayed except maybe if the last round is finished (in this case I understand that we display the ranking).
Anyway, if you prefer the ranking let's go with it.

src/web/controllers/admin/player_admin_controller.py
pascalaubry: Are you OK with that?

pascalaubry (approved):

PR #1066, opened 2025-08-31
Sort screens by name (closes #1058)

### LINKED ISSUE #1058: Order the screen cards by name
Screen cards are actually sorted by unique ID, they should be sorted by name.

Image

timothyarmes:


Molrn (dismissed):
For coherence screen families also need to be sorted by name

pascalaubry (changes requested):
I think that all the methods `xxx_sorted by uniq_id()` should be removed at the same time.

Molrn (approved):
LGTM

timothyarmes:
> I think that all the methods `xxx_sorted by uniq_id()` should be removed at the same time.

We will remove all the uuid references after roles had been merged

PR #1076, opened 2025-09-01
Fix warning on prizes assignment view

pascalaubry:


src/data/prize/prize_sharing.py
```diff
@@ -157,7 +157,7 @@ def calculate_prizes(
                         value=(own + total / len(players_in_tie)) / 2,
                         is_main=True,
                         warning=_(
-                            'Other players in this score group are not included since the share would be less than the threshold.'
+                            'Other players in this score group are not included in the prizes (price below the minimum threshold).'
```
timothyarmes: Other players in this score group are not awarded prizes, as their share would be less than the minimum threshold.

src/data/prize/prize_sharing.py
```diff
@@ -95,7 +95,7 @@ def calculate_prizes(
                         value=share,
                         is_main=True,
                         warning=_(
-                            'Other players in this score group are not included since the share would be less than the threshold.'
+                            'Other players in this score group are not included in the prizes (price below the minimum threshold).'
```
timothyarmes: Other players in this score group are not awarded prizes, as their share would be less than the minimum threshold.

timothyarmes (dismissed):

Molrn (approved):
LGTM

PR #1081, opened 2025-09-02
Simplify i18n update

pascalaubry:
Now rely on file fingerprints instead of file modification dates (broken by git pulls).
Also fixed the absence of the MO files when testing on GitHub (https://github.com/Sharly-Chess/sharly-chess/actions/runs/17402909617/job/49400005231).


pascalaubry:
Was written in ``feature/roles``.

PR #1091, opened 2025-09-04
Fix the language used everywhere

pascalaubry:


src/common/sharly_chess_config.py
```diff
@@ -43,14 +43,9 @@
 class SharlyChessConfig(metaclass=Singleton):
     """The configuration for the application, read from the database."""
 
-    _stored_config: ClassVar[StoredConfig | None] = None
-
```
Molrn, line 47: It's important that this value is cached, `SharlyChessConfig()` opens and reads the database and it is called everywhere. This could add 5 to 10 database opening / closing per request, I'm really opposed to that.

Molrn (dismissed):

Molrn (approved):

PR #1119, opened 2025-09-08
Fix estimated rankings on Papi export

Molrn:


Amaras (approved):
Works well, I see no difference between our papi-TPR and papi's Performance export

timothyarmes (approved):
Oof, it's back to normal, identical with and without overrides.

PR #1124, opened 2025-09-09
Restore log colours

timothyarmes:


pascalaubry (dismissed):
1. Colors are back, thanks.
2. Updating the config on the web UI updates the way logs are displayed in the console, nice.
3. Printing the time is enough (date is in the file logs); I changed `logging_config['formatters']['console_formatter']['date_fmt']` from `'%Y-%m-%d %H:%M:%S'` to `'%H:%M:%S'` (format is still `'%Y-%m-%d %H:%M:%S'` for the file formatter).
3. The date should be printed to the console only when set this way in the application config; I changed `div.innerHTML = '<span class="ts">[' + ts + ']</span> ' + html;` to `div.innerHTML = html;` for that.

The two last points are on PR #1125 (on branch `bugfix/colors-2`, based on your `bugfix/colors`).

4. Something is wrong on the PyCharm console and I did not find how to solve it:
- we lost the colours, I don't understand why because the console handler has not been changed (or I missed something);
- all the logs are duplicated with a strange ``None`` at the end of the second line.

pascalaubry (approved):

PR #1131, opened 2025-09-11
Add database view infrastructure

Amaras:
This PR creates the infrastructure to introduce database views and creates 3 views I find useful.

This is mainly a convenience to compute norms, but could also be used to fetch data in a more natural way for pairings, or the players tab.

The names are up for debate, as is the structure.
If the underlying structure changes, the views will need to be updated.

This will mainly be a concern when we want to handle putting players in several tournaments of the event.


Amaras:
TODO: use the new database structure to change the views

PR #1133, opened 2025-09-11
Cleaner SQL

timothyarmes:


src/database/sqlite/event/views/v002_all_boards.sql
```diff
@@ -0,0 +1,64 @@
...
+    ON pa.board_id = b.id
+   AND pa.player_id = b.white_player_id
+LEFT JOIN pairing o
+    ON o.tournament_id = pa.tournament_id
+   AND o.round = pa.round
+   AND o.player_id = b.black_player_id
+LEFT JOIN player wp ON wp.id = b.white_player_id
+LEFT JOIN player bp ON bp.id = b.black_player_id
+LEFT JOIN tournament t ON t.id = pa.tournament_id
+JOIN info i ON i.ROWID = 1
```
Amaras: You don't need `info` since you've put the need in `player_effective_ratings`
```suggestion
```

src/database/sqlite/event/views/v003_all_byes.sql
```diff
@@ -0,0 +1,31 @@
...
+    pr.effective_rating_type AS rating_type,
+    pr.is_overridden AS overridden,
+    p.plugin_data,
+
+    -- RESULT
+    pa.result
+FROM pairing pa
+LEFT JOIN player p ON p.id = pa.player_id
+LEFT JOIN tournament t ON t.id = pa.tournament_id
+JOIN info i ON i.ROWID = 1
```
Amaras: Same as above
```suggestion
```

src/database/sqlite/event/views/v004_all_players.sql
Amaras, line 1: I guess you didn't develop what is now `v001` when you made this view.
Use that instead and remove the named subquery `rating_override`

Amaras (changes requested):
Nice, could you add back at least part of my relatively detailed comments?

I fell like explaining the structure is a good idea for future us and other devs.

Amaras (approved):

PR #1143, opened 2025-09-14
Fixed i18n PO files rebuild

pascalaubry:
MO files were not always rebuilt when PO files were manually updated with no change in the sources files.


Molrn (changes requested):
Still the same issue, .mo files not updated on startup:
- Run the app
- update the .po
- restart the app
- Translations appear as before updating the .po

I'm certain the updated translations appeared until a few weeks ago, not sure what broke that

Molrn (approved):
LGTM

PR #1144, opened 2025-09-14
Fix select_input() with one option

pascalaubry:
When not in a multiple select, the input type is not filled if only one option (and the select is disabled).

### DIFF
--- a/src/web/templates/admin/common/macros.j2
+++ b/src/web/templates/admin/common/macros.j2
@@ -265,7 +265,14 @@ with_indicator: display a loading indicator, optional
                 "
             >
                 {% if input_type == 'select' %}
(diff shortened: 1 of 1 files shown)


Molrn:
In which select does this happen ? It should be selected without your fix, the problem is probably that data['name'] is incorrectly parametered

pascalaubry:
> In which select does this happen ? It should be selected without your fix, the problem is probably that data['name'] is incorrectly parametered

On the profile modal.

Molrn:
I don't have a select on the profile modal ^^'

pascalaubry:
> I don't have a select on the profile modal ^^'

You need a premium version for that ;-)

Molrn:
image  
works well on the players modal, please setup the data correctly, this fix adds unnecessary complexity (and does not work when optgroups are defined)

PR #1146, opened 2025-09-14
Remove device-based access control

pascalaubry:
Removed the devices
Simplified the profile modal
Force the clients to reconnect when the user account is disabled or the password is changed


src/web/templates/common/profile/profile_modal.html
```diff
@@ -47,57 +47,62 @@ <h2 class="modal-title">
...
+                            hx-post="{{ url_for('profile-login', event_uniq_id=admin_event.uniq_id) }}"
+                            hx-target="body"
+                            hx-indicator="#please-wait"
+                        >
+                            <div class="row">
+                                <div class="col-12 col-md-6 mb-3">
+                                    {{ admin_macros.select_input(
+                                        name='account_id',
+                                        label=_('Account:'),
+                                        help_text=_('Choose the account.'),
```
Molrn:
```suggestion
                                        help_text='',
```

src/web/templates/common/profile/profile_modal.html
```diff
@@ -47,57 +47,62 @@ <h2 class="modal-title">
...
+                                        select_options=active_user_account_options,
+                                        data=data, errors=errors
+                                    ) }}
+                                </div>
+                                <div class="col-12 col-md-6 mb-3">
+                                    {{ admin_macros.password_input(
+                                        name='password',
+                                        label=_('Password:'),
+                                        help_text=_('Enter the password of the account.'),
```
Molrn:
```suggestion
                                        help_text='',
```

src/database/sqlite/event/migrations/m045_add_roles.py
```diff
@@ -1,40 +1,43 @@
...
+            ],
+            2: [
+                CheckInAccessLevel.static_id(),
+                ResultsEntryAccessLevel.static_id(),
+            ],
+        }.items():
+            self.database.execute(
+                'INSERT INTO `account`(`id`, `access_levels`) VALUES (?, ?)',
+                (account_id, SQLiteDatabase.dump_to_json_database_field(access_levels)),
+            )
```
Molrn: we need to have no value in the table by default, such that we can detect when to display the documentation vs when to display the default objects
```suggestion
```

src/database/sqlite/event/migrations/m045_add_roles.py
Molrn, line 1: no import from the core into migrations. When values are needed, you copy them (ex: `AdministrationAccessLevel.static_id()` --> `ADMINISTRATION`)
This way the migration is self contained and static (important to have all the DBs in the same state)

src/web/controllers/admin/access_admin_controller.py
```diff
@@ -320,24 +275,44 @@ def _validate_account_form_data(
...
-                action != FormAction.UPDATE
-                or (account and account.username != username)
-            ):
-                errors[field] = _('Account [{username}] already exists.').format(
-                    username=username
+            last_name: str = (
+                WebContext.form_data_to_str(data, field := 'last_name') or ''
+            )
+            if not last_name:
+                errors[field] = _('Please enter the last name.')
```
Molrn:
```suggestion
                errors[field] = _('This field is required.')
```
It's the kind of message that sounds cleaner and are more instinctive to users when they are generic and not customized (ex: the user sees the message and automatically knows without reading what the problem is, without even reading).

It is also easier to maintain / to consider a new translation to not have all of these customized messages for no reason

src/web/controllers/admin/access_admin_controller.py
```diff
@@ -320,24 +275,44 @@ def _validate_account_form_data(
...
+                    check_full_name_not_used
+                    and full_name in event.user_accounts_sorted_by_name
+                ):
+                    errors[field] = _(
+                        'Account [{account_name}] already exists.'
+                    ).format(account_name=full_name)
+            if action == FormAction.CREATE:
+                password = WebContext.form_data_to_str(data, field := 'password')
+                if not password:
+                    errors[field] = _('Please enter the password.')
```
Molrn:
```suggestion
                    errors[field] = _('This field is required.')
```
up

src/web/templates/common/profile/profile_button.html
Molrn, line 1: image
Without device, the IP should not be on the Profile button.  
It should be simplified to
`<i class="bi-person"></i> {{ Anonymous | Administrator | Jean-Michel }}`  
Also if you could remove the language from this button it would be nice

src/web/templates/admin/accounts/account_modal.html
```diff
@@ -35,10 +35,19 @@ <h4>
...
+                            help_text=_('The first name.'),
+                            data=data, errors=errors
+                        ) }}
+                    </div>
+                    <div class="row mb-3">
+                        {{ admin_macros.text_input(
+                            name='first_name',
+                            label=_('First name:'),
+                            placeholder=_('E.g.: John'),
+                            help_text=_('The last name.'),
```
Molrn:
```suggestion
                            name='last_name',
                            label=_('Last name:'),
                            placeholder=_('E.g.: DOE'),
                            help_text='',
                            data=data, errors=errors
                        ) }}
                    </div>
                    <div class="row mb-3">
                        {{ admin_macros.text_input(
                            name='first_name',
                            label=_('First name:'),
                            placeholder=_('E.g.: John'),
                            help_text='',
```
Funny swap ^^  Besides that, the helptexts is useless. If you look at all the recent forms (prizes, rotators, display controllers) none of them have helptexts, or only when it is necessary. When the form is in one column it also does not make the layout feel weird to not have a help text so please try to only add some if necessary

src/web/templates/admin/accounts/account_modal.html
```diff
@@ -11,7 +11,7 @@ <h2 class="modal-title">
                     {% if admin_account.anonymous %}
                         {{ _('Editing of unauthenticated access') }}
                     {% else %}
-                        {{ _('Editing of account [%(username)s]', username=admin_account.username) }}
+                        {{ _('Editing of account [%(account_name)s]', account_name=admin_account.full_name) }}
```
Molrn: The titles were modified on every modal before the roles merge. Please also edit the other titles (Create account, Edit anonymous access)
```suggestion
                        {{ _('Edit account') }}
```

src/data/access_levels/entities.py
Molrn, line 1: move to `src/data/account.py`

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -173,7 +173,7 @@ msgid "(family)"
 msgstr "(famille)"
 
 msgid "(inherited)"
-msgstr "(hérité)"
+msgstr "hérité"
```
```suggestion
msgstr "(hérité)"
```
or the parenthesis should not be there in english

Molrn (dismissed):

src/web/templates/common/profile/profile_button.html
pascalaubry: OK for the device, but I want to keep the language.

src/database/sqlite/event/migrations/m045_add_roles.py
pascalaubry: I do not agree.
There is no need to make the code more complex by restoring all the `*_predefined_default_*()` I removed, it is easy to know if the access levels are the default ones and print de default documentation in this case only.

Molrn: Ok, to make this work with values already in the DB we'd need two buttons: `Create account` and `Edit anonymous access`. Not what I had in mind originally (one button with a label like `Get started` or `Create default access` which inserts the objects) but works, let's go with that ;)

pascalaubry: We now start with an empty database and add real stored accounts just before adding an account or updating an existing one.
Call `Event.predefined_accounts()` if you want to display a specific message before starting to edit the accounts.

Molrn (commented):
All good for me, just a #1154, #1155 and the failing test to adress

Molrn (approved):
LGTM

PR #1151, opened 2025-09-14
Fix export on Mac, use correct icon

timothyarmes:


pascalaubry (approved):
#1152 to prepare Linux support

pascalaubry (approved):

PR #1155, opened 2025-09-14
Remove app level profile modal

Molrn:


pascalaubry (changes requested):
OK with everything, please can you also clean ProfileController?

PR #1156, opened 2025-09-15
Home tab improvements + remove settings tab

Molrn:


src/web/controllers/admin/index_admin_controller.py
```diff
@@ -172,7 +172,7 @@ def _admin_render(
             'home': {
                 'title': _('Home'),
                 'template': 'index/home_tab.html',
-                'icon_class': 'bi-qr-code',
+                'icon_class': 'bi-house-fill',
```
pascalaubry: I remember Tim did not want to have the house icon to goto the home page because having the icon on a page means you already are on the home page.
```suggestion
                'icon_class': 'bi-qr-code',
```

src/web/templates/admin/index/home_tab.html
```diff
@@ -70,7 +74,9 @@ <h4 class="text-center mt-3">{{ _('Connect other devices!') }}</h4>
                             {{ macros.tooltip_attributes('info', _('Connect your devices (laptops, tablets, smartphones) to the Sharly Chess server by scanning this QR code (click below to get a printable version).'), 'right') }}
                         />
                         <br/>
-                        {{ sharly_chess_config.app_url(ip) }}
+                        <a href="{{ sharly_chess_config.app_url(ip) }}">
```
pascalaubry:
```suggestion
                        <a href="{{ sharly_chess_config.app_url(ip) }}" target="_blank">
```

timothyarmes:
Just one thing that bugs me....

The home page has a house icon and the title "home", but it's not a button
The Events page also has a house icon and the title "home", but it is a button
On the tournament pages we have a back button to get back to the event

I'd prefer:

* To use a back button, it's clearer that it's an action button
* To not have the "Home" title on the home page at all.

Molrn:
I don't think I get what bugs you, it seems pretty standard to have a home button that displays a welcome page. 
> To use a back button, it's clearer that it's an action button   

image   

not fan honestly. On the event inteface it shows that you move back to a different part of the app, here it stays on the same interface, it's just to display the home page.    

> To not have the "Home" title on the home page at all.    

I'm confused, how would you access the home page like this ?

locale/en/LC_MESSAGES/messages.po
```diff
@@ -6418,3 +6333,15 @@ msgstr ""
...
+
+#~ msgid "Error message"
+#~ msgstr ""
+
+#~ msgid "Information message"
+#~ msgstr ""
+
+#~ msgid "Warning message"
+#~ msgstr ""
+
```
```suggestion
```

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -6418,3 +6333,15 @@ msgstr "½ - 0 (NC)"
...
+
+#~ msgid "Error message"
+#~ msgstr "Message d'erreur"
+
+#~ msgid "Information message"
+#~ msgstr "Message d'information"
+
+#~ msgid "Warning message"
+#~ msgstr "Message d'avertissement"
+
```
```suggestion
```

pascalaubry (dismissed):
Note: review started very early this morning and finished only now.

PR #1160, opened 2025-09-15
Fix multi-select placeholder

Molrn:
before:   
image

After:  
image


pascalaubry (approved):

timothyarmes (dismissed):

timothyarmes (dismissed):
Select2 styling should be in admin.css with eh rest of the overrides

pascalaubry (approved):

PR #1163, opened 2025-09-15
Update access levels doc generation script

pascalaubry:


src/data/tie_breaks/tie_breaks.py
```diff
@@ -1288,14 +1288,14 @@ class DirectEncounterTieBreak(TieBreak):
...
-    @staticmethod
-    def static_name() -> str:
-        return _('Direct encounter')
-
     @staticmethod
     def static_id() -> str:
         return 'DIRECT_ENCOUNTER'
 
+    @classmethod
+    def static_name(cls, locale: str | None = None) -> str:
```
Molrn:
```suggestion
    def static_name(cls) -> str:
```

src/data/tie_breaks/tie_breaks.py
```diff
@@ -1451,14 +1451,14 @@ def _compute_player_min_max_points(
...
-    @staticmethod
-    def static_name() -> str:
-        return _('Manual')
-
     @staticmethod
     def static_id() -> str:
         return 'MANUAL'
 
+    @classmethod
+    def static_name(cls, locale: str | None = None) -> str:
```
Molrn:
```suggestion
    def static_name(cls) -> str:
```

src/data/print_documents/documents.py
```diff
@@ -584,7 +584,7 @@ def available_options() -> list[type[PrintOption]]:
 
     @property
     def title(self) -> str:
-        return self.name
+        return self.name()
```
Molrn:
```suggestion
        return self.name
```

src/data/print_documents/documents.py
```diff
@@ -689,7 +689,7 @@ def available_options() -> list[type[PrintOption]]:
...
```
Molrn, line 692:
```suggestion
        return self.name
```

Molrn (dismissed):
Note: in the future, please try to avoid as much as possible to use long `match...case` statements to map values, this is much harder to maintain and much less efficient than having all the values of an element gathered in the same place (ex: IdentifiableEntity, or just OOP in general). 

Enums are not meant to have such properties, if you end up needing to add properties mapping over all the values of an enum it's that the choice of en enum was not appropriate to begin with

Molrn (approved):
LGTM

PR #1164, opened 2025-09-15
Translate Staff to Staff

pascalaubry:
And do not rebuild the MO file twice when running i18n_update and the PO file has been changed


Molrn (approved):
LGTM

PR #1166, opened 2025-09-15
Allow any character for account names

pascalaubry:
No need to consider names as IDs anymore.


Molrn (dismissed):
At this point I don't think it's even necessary to validate it. Which characters are not included by this regex that would cause a problem / should definitely not be there ?

pascalaubry:
> At this point I don't think it's even necessary to validate it. Which characters are not included by this regex that would cause a problem / should definitely not be there ?

You are right, I do not see any. Change applied.

Molrn (approved):
LGTM

PR #1183, opened 2025-09-18
Make create event prominent

timothyarmes:


pascalaubry (commented):
I must say that I prefer the old "Create an event" button (I like it top right on the home page like on the other events pages).

But if I am the only one i can do with the new layout.

Before:
image

After:
image

The other pages:
image

timothyarmes:
I'm normally the one to be militant about UX consistency! :)

In this case however:

* Home pages very often have a different layout, so it's not shocking.
* Without the design change the first thing a user sees is how to connect your devices, as if that's somehow the core feature of the app.  I don't agree - I'm not even sure that I want to see that on the home page at all.
* The most important feature for a new user is to create an event.  I think it makes sense to have it was a sort of "getting started" button

pascalaubry (approved):
OK, convinced.

PR #1190, opened 2025-09-20
Fix Export

timothyarmes:


Amaras (approved):
trusting you on the mac build, we'll see soon enough if the mac export works

PR #1199, opened 2025-09-22
Raise an HttpException instead of returning redirect_error

Molrn:
After this, we won't rely on returning web_context.error to stop the propagation of an error, which definitely was not everywhere.


timothyarmes (approved):
LTGM, hard to test though.

PR #1200, opened 2025-09-22
QR Code print option

timothyarmes:


Molrn (dismissed):
> Scannez pour connecter votre appareil au serveur Sharly Chess.  

I think it lacks a message about the requirement to connect to the network before the url is accessible. It might lead to users just scanning the QR code and unexpectedly getting nowhere.  

Suggestion: 
- if it is possible to fetch the name of the network (ex: SFR_XXXX) it would be nice (for wifi)  
- For not wifi networks (or if fetching the name is impossible), just a "Connect to the Sharly Chess network via {VPN|ethernet}, then scan to access to SC"

Molrn (approved):
LGTM

PR #1214, opened 2025-09-25
Replace sse with web sockets

timothyarmes:


src/web/templates/common/js/sse.js
Molrn, line 1: is there still a usage for this file ? It appears not to be imported

src/web/static/lib/htmx/sse-2.2.3/sse.js
Molrn, line 1: to remove

Molrn (changes requested):
`Connection open` / `Connection closed` are a bit log poluting, it would be nice to remove them. otherwise works fine, even with many tabs

Molrn (approved):
LGTM

PR #1225, opened 2025-09-27
Always display lastname first on the Players tab to ease check-in (closes #1224)

### LINKED ISSUE #1224: Change the display of the players' name on the Players tab?
Not easy to check-in players in English with this display, I think we should not use Player.full_name on the Players tab.
@timothyarmes up to you.

Image

Image

pascalaubry:


Amaras (dismissed):
This can't be happening without changing the tests.

timothyarmes (approved):

PR #1230, opened 2025-09-28
Improve player search

Molrn:


pascalaubry (approved):
Works fine (tested it with local databases present/absent and connected/disconnected).

timothyarmes (approved):

PR #1247, opened 2025-09-29
make the home page more compact (closes #1237)

### LINKED ISSUE #1237: Improve the layout of the home page for lanscape network clients
The message at the bottom of the home page is not visible on most devices. The display is probably well adapted to portait tablets but locale selector and logo should be reorganized for landscape devices.

Image

timothyarmes:


Molrn (approved):
LGTM

PR #1259, opened 2025-09-29
Insert new boards instead of appending them

Molrn:
closes #1241

PR #1261, opened 2025-09-30
Use tournament full_name for papi export

timothyarmes:


Molrn:
@timothyarmes not sure you meant for it to be based off of #1259

PR #1270, opened 2025-09-30
Take check-in status into account

Amaras:


pascalaubry (changes requested):
Add "Display the check-in status on check-in lists" to the release notes?

pascalaubry (approved):

PR #1282, opened 2025-10-01
Sort last name matches first

timothyarmes:


Molrn (dismissed):
Works well, LGTM. We gotta do something about these failing tests tho

Molrn (approved):
indeed, good catch ^^

PR #1283, opened 2025-10-01
Directly update local database options

Molrn:
This avoids having to save the database when exiting the modal, which is counter-intuitive as the modal is both informative and interactable. Applies the same styling as the the history modal


timothyarmes (approved):
Cool

PR #1284, opened 2025-10-02
Clarify event auth actions

Molrn:
Event-level action: singular
All events action: `MANAGE_EVENTS`
Should be the same for actions of the category `TOURNAMENTS` (so far they all are applied to all tournaments)


pascalaubry (dismissed):
Please run script generate_access_levels_doc and update the docs.

pascalaubry (approved):

PR #1287, opened 2025-10-03
Separate E/F/N ratings

timothyarmes:


src/plugins/ffe/templates/ffe_tournament_form_fields.html
```diff
@@ -16,3 +16,8 @@ <h4>
         ) }}
     </div>
 </div>
+
+<script>
+    $('#player-rating-type').val('3');
+    $('#player-rating-type').prop('disabled', true);
+</script>
```
Molrn, line 24: Missing an explanation as of why it is disabled (either in the tooltip or help_text)

src/plugins/ffe/templates/ffe_event_form_fields.html
```diff
@@ -24,3 +24,8 @@ <h4>
         ) }}
     </div>
 </div>
+
+<script>
+    $('#player-rating-type').val('3');
+    $('#player-rating-type').prop('disabled', true);
```
Molrn, line 30: Missing an explanation as of why it is disabled (either in the tooltip or help_text)

Molrn (changes requested):
All good, one detail. Might wanna have a look at my review commit too

Molrn (approved):
LGTM

PR #1299, opened 2025-10-06
Ignore unknown fields on ChessEvent Import

Molrn:


src/data/input_output/dict_reader.py
```diff
@@ -3,10 +3,14 @@
 from types import UnionType, GenericAlias
 from typing import Any, get_origin, get_args, cast
 
+from common import get_logger
```
pascalaubry:
```suggestion
from common.logger import get_logger
```

pascalaubry (changes requested):

src/plugins/chessevent/tournament_importer/data.py
```diff
@@ -31,6 +31,10 @@ class ChessEventPlayer:
     ffe_club_id: int = 0
     ffe_club: str = ''
     category: PlayerCategory = PlayerCategory.NONE
+    initial_fee: str | int | float | None = None
+    discount: str | int | float | None = None
+    paid_before: str | int | float | None = None
+    paid_site: str | int | float | None = None
```
pascalaubry, line 37: I would prefer these fields to be added in a next release as mandatory fields but why not.

pascalaubry (approved):

PR #1317, opened 2025-10-10
Move upload button back to the side bar

timothyarmes:


src/web/templates/admin/event_layout_nav_item.html
```diff
@@ -67,6 +67,13 @@
                                 >
                                     {% if nav_tab.icon_class %}
                                         <i class="{{ sub_nav_tab.icon_class }} pe-1"></i>
+                                    {% endif %}{{nav_tab.icon_class}}
```
Molrn: I don't get this, it's the nav_tab.icon_class at a sub_nav_tab level, and {{nav_tab.icon_class}} is placed in the template directly (probably a mistake). I don't get why it nver shows in the template, but it most likely needs to be removed

timothyarmes: It was debug...

Molrn (approved):

PR #1327, opened 2025-10-13
Use mac to run tests

timothyarmes:


pascalaubry (approved):
I am OK with this one but we should let @Amaras some time to investigate on sessions FileStore before merging.

timothyarmes:
> I am OK with this one but we should let @Amaras some time to investigate on sessions FileStore before merging.

That's a different issue - the sessions need fixing on Windows, but this is about reliable and faster tests on GitHub

pascalaubry:
OK so let's merge.

PR #1330, opened 2025-10-13
Move sessions to SQLiteStore

Amaras:
FileStore regularly raised `PermissionError` on Windows platforms when used in the context of sessions.

This SQLite store with connection pooling runs about as fast as the FileStore, from my cursory exploration, and never raised any PermissionError.

Maybe 3.1 deserves to have this change as well.

Note that I don't load anything from the previous session store, so all sessions settings are lost.
I think this change is worth the annoyance.


pascalaubry (approved):
LGTM (tried to challenge it with a dozen screens without any problem).

PR #1332, opened 2025-10-14
Federation plugins

timothyarmes:


src/plugins/chessevent/templates/chessevent_event_form_fields.html
```diff
@@ -1,11 +1,11 @@
-<h4>
+<h4 id="chessevent-settings-header">
     {{ _('ChessEvent settings') }}
     <span style="font-size: 90%">{{ macros.doc_icon('chessevent', 'ms-2') }}</span>
 </h4>
 <h6 class="text-body-secondary pt-1 pb-1">
     {{ _('Default credentials used to import tournaments from the ChessEvent platform.') }}
 </h6>
```
Molrn: not hidden. Maybe put it all into one single `chessevent-settings` container ?

src/plugins/utils.py
```diff
@@ -157,11 +154,20 @@ def is_state_editable(self) -> bool:
         (enabled / disabled) is editable"""
         return True
 
+    @property
+    def federation(self) -> str | None:
```
Molrn, line 158:
```suggestion
    def federations(self) -> list[str] | None:
```
It's ok if the current UI only takes into account the current possibilities, but the code needs to take into account potential evolutions. We discussed this extensively a few months ago, this needs to be a list. I had amended my position a lot until this one, I won't move on this

tests/test_config.py
```diff
@@ -232,6 +238,7 @@ def create_tournament(
                 headers={'Content-Type': 'application/x-www-form-urlencoded'},
                 data=form_data,
             )
+            print(res)
```
Molrn:
```suggestion
```

RELEASE_NOTES.md
```diff
@@ -3,6 +3,7 @@ English version below
 
 ## Fonctionnement
 
+- Plug-ins spécifiques à une fédération (v3.2.0)
```
Molrn:
```suggestion
- Plug-ins spécifiques à une fédération (3.2.0)
```

src/plugins/chessevent/templates/chessevent_event_form_fields.html
```diff
@@ -40,3 +40,15 @@ <h6 class="text-body-secondary pt-1 pb-1">
...
+<script>
+    function updateChessEventVisibility() {
+        const federation = $('#federation').val();
+        const show = federation === 'FRA';
+        $('#chessevent-settings, #chessevent-settings-header').toggle(show);
+    }
+
+    updateChessEventVisibility();
+    $('#federation').on('change', updateChessEventVisibility);
+</script>
```
Molrn, line 54: Should use the Plugin.federations property. Currently the logic for this is duplicated 3 times per plugin, it should not be duplicated at all.

This way the toggle could be handled dynamycally, with each plugin having its own container with a generic id.

Molrn (dismissed):
Did a first pass, I'll need to look at it with fresh eyes tomorrow.  

For such an important change, it is not obvious enough what happens in the event form. 
There needs to be a clear indicator of which plugins are enablable for the chosen federation, and which ones are disabled when you change the federation.  

I suggest adding a `Plugins` section similar to the one in the application settings modal, where the switch would be disabled with a tooltip if the federation does not match. It would also allow to enable / disable other plugins specifically for this event. It requires a `plugin` table in the event database, but I think it's for the best

src/plugins/utils.py
timothyarmes: I agree, but this s just used for the setting display modal as it is today.

The enabling logic is in is_enabled_for_event, eg:

```
    @override
    def is_enabled_for_event(self, event: Optional['Event']) -> bool:
        return event is not None and event.federation == 'FRA'
```

It's not just a list, but anything we want to be super flexible

src/plugins/chessevent/templates/chessevent_event_form_fields.html
timothyarmes: Plugin.federations just for the current organisation of the plugins in the parameter settings.  We'll see how that evolves, I don't have visibility.

src/plugins/utils.py
Molrn: this is what I'm saying, the enabling / disabling logic should not be different from the one organizing the UI, to ensure what the user understands is what actually happens. Multi-federations could even simply be integrated in the current UI:  
image   

Federations already bring a great deal of complexity on their own for plugin enabling / disabling, I don't think we need to yet take more complexity into account. However it needs to be as flexible as possible on federations.

src/plugins/ffe/templates/ffe_event_form_fields.html
```diff
@@ -29,4 +29,13 @@ <h4>
     $('#player-rating-type').val('3');
     $('#player-rating-type').prop('disabled', true);
     $('#player-rating-type-help').text('{{ _('The FFE always uses the FIDE rating when available.') }}');
```
Molrn: now need to change on federation update

Molrn (approved):
LGTM

PR #1338, opened 2025-10-18
Tie-break interface

Molrn:

PR #1340, opened 2025-10-19
Added missing tie-break i18n strings

pascalaubry:


Molrn:
directly added to #1338

PR #1342, opened 2025-10-19
Fixed the display of fixed board numbers on boards/input screens

pascalaubry:
Before:
image

After:
image


Amaras (commented):
I'm not sure if I'm okay with changing which is brackets.
Otherwise, it's nice :)

PR #1359, opened 2025-10-21
Use FIDE ones for now

timothyarmes:


src/data/pairings/systems.py
```diff
@@ -189,12 +189,15 @@ def recommended_tie_breaks(self) -> list['TieBreak']:
...
         return [
-            tie_breaks.DirectEncounterTieBreak(),
-            tie_breaks.WinsTieBreak(),
             tie_breaks.StandardBuchholzTieBreak(
                 [CutterWithMedianTieBreakOption(Cut1TieBreakCutter().id)]
             ),
+            tie_breaks.StandardBuchholzTieBreak(),
+            tie_breaks.SonnebornBergerTieBreak(),
             tie_breaks.ProgressiveScoresTieBreak(),
+            tie_breaks.DirectEncounterTieBreak(),
```
pascalaubry: IMO there is no interest and no use setting DE as the 5th tie-break, better remove it in this case.

Molrn (approved):
LGTM

PR #1371, opened 2025-10-22
Calculate if norm requirement 1.5.6a is met

timothyarmes:


Amaras (commented):
Nice!
Just need to check the failing tests

Amaras (approved):

PR #1376, opened 2025-10-22
Fix online player search error messages

Molrn:
Currently any error lead to a `No internet connection` error.


timothyarmes (approved):
LGTM

PR #1387, opened 2025-10-27
Player and Board Cards

pascalaubry:
The goal of this PR is to implement Player and Board Cards.

Cards are based on Jinja templates.

To allow users to create their own cards, very simple and frozen data is passed to the templates not to permit any change of the data on the server.

- ✅ Make the HTML output self-contained (to be sent to anybody not connected to the Sharly Chess server)
- ✅ Support text items
- ✅ Support image items
- ✅ Support fonts
- ✅ Allow users to select a template from a list of pre-defined and custom templates
- ✅ Add mirroring (duplication on the back side)
- ✅ Add the possibility to set different items on the back side
- ✅ Simplify the default board card (remove the Sharly Chess logo)
- ✅ Add a print option for the crop marks
- ✅ Recover the templates when upgrading
- ✅ Restore the dotted line on top
- ✅ Fix option display
- ✅ Finalize default boards
- ✅ Apply real mirroring on pairing cards (rotating for the others)**
- ✅ Card preview with type card toolips
- ✅ Compact the place card options on the print modal
- ✅ Memorize the last print view used (and if possible the last print card type
- ✅ Minimize the size of HTML files
- ✅ Fix the border around images

Player cards:
- ✅ Create a default Player Card
- ✅ Add a hook to allow plugins to augment PrintCardPlayer
- ✅ Document the data passed to the templates
- ✅ Allow users to choose some players only

Board/Pairing cards:
- ✅ Create a simple Board Card with no player information (all the possible standard numbers + the fixed boards)
- ✅ Create a simple Pairing Card with player information
- ✅ Allow users to choose some board numbers only
- ✅ Allow users to choose a round and pass corresponding pairings to the templates
- ✅ Document the data passed to the templates


timothyarmes (approved):
Top

timothyarmes (approved):

PR #1395, opened 2025-10-29
Fix typo in norm_report.html

pascalaubry:


Amaras:
While I agree with you, this typo is also in the document, so I don't know if we should be more correct than FIDE

Amaras (approved):

PR #1419, opened 2025-11-01
Hide the 'Return to tournament' and 'Withdraw from tournament' buttons when not needed

pascalaubry:


Amaras (approved):
I have slight concerns for HPB and FPB at the last round, but this is good for this release

pascalaubry:
> I have slight concerns for HPB and FPB at the last round, but this is good for this release

Yes, we have to discuss before disabling bye management once a tournament is finished.

PR #1425, opened 2025-11-03
Staff roles

timothyarmes:


src/database/sqlite/event/migrations/m056_account_roles.py
```diff
@@ -0,0 +1,48 @@
...
+class Migration(BaseMigration):
+    def forward(self):
+        self.database.execute('ALTER TABLE `account` ADD `fide_id` INTEGER')
+
+        # Create account_role table
+        self.database.execute(
+            'CREATE TABLE `account_role` ('
+            '   `account_id` INTEGER NOT NULL,'
+            '   `tournament_id` INTEGER,'  # NULL for organisers
+            '   `role` TEXT NOT NULL CHECK (`role` IN ("chief_arbiter", "deputy_arbiter", "organiser")),'
```
Molrn:
```suggestion
            '   `role` TEXT NOT NULL,'
```
We've been avoiding putting that kind of check in the DB structure, it is really a blockage to evolutions and does not make our data any safer.

src/database/sqlite/event/migrations/m056_account_roles.py
```diff
@@ -0,0 +1,48 @@
...
+            '   FOREIGN KEY (`account_id`) REFERENCES `account`(`id`) ON DELETE CASCADE,'
+            '   FOREIGN KEY (`tournament_id`) REFERENCES `tournament`(`id`) ON DELETE CASCADE,'
+            '   CHECK ('
+            '       (`role` = "organiser" AND `tournament_id` IS NULL) OR '
+            '       (`role` IN ("chief_arbiter", "deputy_arbiter") AND `tournament_id` IS NOT NULL)'
+            '   )'
```
Molrn: against this for the same reasons, not a fan of having to recreate the table from scratch and restore all the FK relations if we'd like to add a new role. We know other FIDE roles exist, it might be in an evolution of TRF, let's not limit ourselves now

src/database/sqlite/event/event_database.py
```diff
@@ -2491,6 +2499,77 @@ def update_stored_account(self, stored_account: StoredAccount) -> StoredAccount:
...
+        for row in self.fetchall():
+            role = row['role']
+            tournament_id = row['tournament_id']
+            tournament_ids_by_role[role].append(tournament_id)
+        return [
+            StoredRole(account_id, role, tournament_ids or None)
+            for role, tournament_ids in tournament_ids_by_role.items()
+        ]
+
+    def update_stored_role(self, stored_role: StoredRole):
```
Molrn: That's too much logic for the EventDatabase, we have different layers especially to avoid relying this much on SQLite features, and to make maintenance easier

src/database/sqlite/event/event_store.py
```diff
@@ -323,16 +324,56 @@ class StoredPermission:
     tournament_ids: list[int] | None = None
 
 
+class RoleKind(StrEnum):
+    CHIEF_ARBITER = 'chief_arbiter'
+    DEPUTY_ARBITER = 'deputy_arbiter'
+    ORGANISER = 'organiser'
```
Molrn: that goes in utils/enums, not in the store. Or even better, it would bee nicer as an Identifiable entity (those match: case to map values are legacy)

src/data/account.py
```diff
@@ -17,6 +22,34 @@
     from data.event import Event
 
 
+@dataclass
+class Role:
+    stored_role: StoredRole
+
+    @property
+    def role_kind(self) -> RoleKind:
+        return RoleKind(self.stored_role.role)
```
Molrn:
```suggestion
        return RoleType(self.stored_role.role)
```
We have `{entity}Type` everywhere, feels out of place to have `Kind` here

src/web/controllers/admin/account_admin_controller.py
```diff
@@ -211,6 +223,8 @@ def _read_account_form_data(
         account = web_context.admin_account
         first_name = WebContext.form_data_to_str(data, 'first_name') or ''
         last_name = WebContext.form_data_to_str(data, field := 'last_name') or ''
+        fide_id = WebContext.form_data_to_int(data, 'fide_id')
+        print(data)
```
Molrn:
```suggestion
```

src/web/templates/admin/accounts/account_modal.html
```diff
@@ -103,3 +131,16 @@ <h4 class="mb-3">
         </div>
     </div>
 </div>
+<script>
+    function selectPlayer(last_name, first_name, fide_id) {
+        console.log(last_name, first_name, fide_id);
```
Molrn:
```suggestion
```

src/web/templates/admin/accounts/account_search_results.html
Molrn, line 1: Only diff I see with the other one is the hx-vals passed to the `<li>` element, couldn't it be parametered instead ?

Molrn (changes requested):

Molrn (approved):
LGTM

Molrn (approved):
LGTM again

Molrn (approved):

PR #1429, opened 2025-11-05
Berger related fixes

Molrn:
- Round option not in results doc
- Berger grid document raising a 500
- close check-in message not shown
- Remove check-in bye options for RR
- add Delete close check-in option 
- remove record modal `Withdraw` option
- Display REST_GAME last instead of first
- hide pab_value for Berger tournaments


timothyarmes (approved):
LGTM

PR #1431, opened 2025-11-06
Fix the pairings unsafe safety mode showing unexpectedly

Molrn:
To reproduce before the fix:
- Enable an unsafe mode on the current round of the first tournament
- go to another tab (ex: Tournaments)
- come back (mode appears safe, but session value is unsafe, just not read)
- do any operation (session value is now read, becoming unsafe)

Now: coming back to the Pairings tab resets the safety mode

also:
image


timothyarmes (approved):
LGTM

PR #1524, opened 2025-11-28
Detect antivirus and add Windows Defender exclusion

pascalaubry:


pascalaubry:
Split into PRs #1600 and #1601

PR #1537, opened 2025-12-02
Add FFE licence type to check-in document

Molrn:


timothyarmes:
Would it be useful to show the licence number if they have once too. eg V52441 (A)

Molrn:
I disagree, it doesn't fit the use-case, and it ends up taking a lot of space for nothing, this already barely fits in an A4, unnecessary space should not be added lightly

timothyarmes (approved):

PR #1544, opened 2025-12-05
Use a representative year of birth for Papi important with age category only

timothyarmes:


Molrn (approved):
I need to nuke the PlayerCategory enum, so Imma refactor most of this ^^ but let's merge it for now

PR #1554, opened 2025-12-06
Fix `NOT NULL` constraint violation

Molrn:


pascalaubry (approved):
Works for me with 3.3.6 databases, also successfully recovered much older databases.

PR #1555, opened 2025-12-06
Refactored board_table.py

pascalaubry:


src/data/columns/board_table.py
```diff
@@ -35,31 +35,79 @@ def shared_classes(self) -> str:
...
+class WhitePointsColumn(PointsColumn):
+    @property
+    def is_black(self) -> bool:
+        return False
+
+
+class BlackPointsColumn(PointsColumn):
+    @property
+    def is_black(self) -> bool:
+        return False
```
Molrn:
```suggestion
        return True
```

docs/technical-appendices/i18n.md
Molrn: Weren't we supposed to not edit this file with i18n_update ? Is it a bug, or if not how did you get it updated ?

src/data/columns/board_table.py
Molrn, line 1: A lot of the column logic is now wrong, especially points. Also:  
Image

src/data/columns/player_datasheet.py
Molrn, line 1: datasheet are not displayed in the UI, the tags are extracted to the csv files

Molrn (dismissed):
This is gonna be a nightmare to maintain, it'll have to be added to half the strings. The fact that you did not test the most important feature doesn't bring me confidence that the rest  will be displayed correctly, and it's too wide to be tested.  

Also having it correctly displayed everywhere will most likely mean adding a `|safe` jinja tag on every free text field inputted by the user, which is precisely what it should not be used on. If one user for some reason puts a `<` or a `>` in a name, the layout would be broken.

I'm in favor of just adding a global translate="no". If you don't wanna do that, please only protect the important parts: table rows and player names.

Molrn:
replaced by #1557

pascalaubry:
Reopening for the refactoring of board_table.py

Molrn (changes requested):
This is unnecessary refactoring, which breaks stuff in the process (the points / vpoints logic is all wrong). There is a reason why there was an `is_black` property only in the illegal moves, because it needed to be accessed in the templates. The other classes are self sufficient and don't need to be gathered by type, it adds unnecessary complexity.  

I would really like not to get this merged, the point of the columns class is one column one class, it's much easier to aprehend, you have the logic of the column in the class itself, header + cell content + classes

timothyarmes:
I agree, it’s more complex.

PR #1563, opened 2025-12-10
Raise a clear exception on missing SignTool

pascalaubry:


scripts/export/windows/win_project_builder.py
```diff
@@ -30,6 +30,11 @@ def __init__(self):
         self.signtool_dir: Path = Path(
             f'C:/Program Files (x86)/Windows Kits/10/bin/{signtool_version}/x64'
         )
+        self.signtool_exe: Path = self.signtool_dir / 'signtool.exe'
+        if not self.signtool_exe.is_file():
```
Molrn: The export can currently also be used without signing, I don't think we should remove this usage by raising here. all the sign process only is triggered if the param `--windows-signtool-cert-fingerprint` is used on the script

Molrn (dismissed):

Molrn (approved):
LGTM

src/web/templates/user/screen/timer.js
Amaras, line 1: why is this timer change bundled with the signtool fix?

Molrn: ah, missed that. it comes from #1565, I think it might have been removed from @timothyarmes 's force push this morning. Going over the PRs to ensure there aren't others

PR #1566, opened 2025-12-11
Remove tournaments default modal

Molrn:


timothyarmes:
When I create a new rapide tournament out of France, the override isn't on by default

timothyarmes (approved):
LGTM

PR #1570, opened 2025-12-13
Linux build

timothyarmes:


pascalaubry (approved):
Can't see anything wrong, must be tested now by real Linux users.

PR #1578, opened 2025-12-14
Try to improve build

timothyarmes:


pascalaubry (commented):
Couldn't make it work on Fedora :-(
But still works on Mint.

PR #1579, opened 2025-12-14
Improve update experience

timothyarmes:


locale/fr/LC_MESSAGES/messages.po
```diff
@@ -1569,6 +1569,9 @@ msgstr "Ne rien faire (les joueur·euses seront apparié·es)"
 msgid "Do you want to install example event databases"
 msgstr "Voulez-vous installer des bases d'évènement d'exemple"
 
+msgid "Do you want to quit this older version"
+msgstr "Voulez-vous quitter cette ancienne version ?"
```
Molrn: `?` issue here. I wanted to remove add one in english, but it seems one gets added automatically for some reason, letting you have a look

Image

pyproject.toml
```diff
@@ -6,7 +6,7 @@ build-backend = "setuptools.build_meta"
 name = "sharly-chess"
 description = "An application to help chess arbiters and organizers manage their tournaments"
 requires-python = ">=3.13.0"
-version = "3.4.0b2"
+version = "3.4.0b1"
```
Molrn:
```suggestion
version = "3.4.0b2"
```

Molrn (commented):
- There is still a lack of user feedback if the install fails, the error is logged but IMO there'd need to be another alert considering there is one on success
- Maybe instead of asking to quit we could ask to start the downloaded version ?

pascalaubry:
I have a problem with this one, please do not merge.
image
image
I have checked, all the files are present.

src/common/engine.py
```diff
@@ -509,153 +510,169 @@ def _install_new_version(version: Version, download_url: str) -> bool:
...
+                                if machine in ('aarch64', 'arm64'):
+                                    downloaded_file = (
+                                        tmp_dir
+                                        / f'sharly-chess-{version}-linux-arm64.zip'
+                                    )
+                                elif machine in ('x86_64', 'amd64'):
+                                    downloaded_file = (
+                                        tmp_dir
+                                        / f'sharly-chess-{version}-linux-x86_64.zip'
+                                    )
```
pascalaubry:
```suggestion
                            if machine in ('aarch64', 'arm64'):
                                downloaded_file = (
                                    tmp_dir
                                    / f'sharly-chess-{version}-linux-arm64.zip'
                                )
                            elif machine in ('x86_64', 'amd64'):
                                downloaded_file = (
                                    tmp_dir
                                    / f'sharly-chess-{version}-linux-x86_64.zip'
                                )
```

src/common/engine.py
```diff
@@ -509,153 +510,169 @@ def _install_new_version(version: Version, download_url: str) -> bool:
...
-                )
-                response: Response = get(download_url, allow_redirects=True, timeout=10)
-                response.raise_for_status()
-                if not response:
-                    logger.error('No response from GitHub.')
-                    return False
-                if response.status_code != 200:
-                    logger.error(
-                        'Downloading failed with code [%d].', response.status_code
+        if not new_version_dir.exists():
```
pascalaubry: I think that we better keep this error.
```suggestion
        if new_version_dir.exists():
            logger.error(
                'Version [%s] is already installed in directory [%s], please manually delete this folder before installing.',
                version,
                new_version_dir.absolute(),
            )
            return False
        else:
```

pascalaubry (changes requested):
There is still something wrong (not in this PR) but I will fix fix it in another PR.

pascalaubry (changes requested):
After having tested, I really think that on upgrade success there is no good reason to continue with the old release, so the only option should be exiting the program.

locale/en/LC_MESSAGES/messages.po
```diff
@@ -3889,6 +3892,9 @@ msgstr ""
 msgid "Refresh this page."
 msgstr ""
 
+msgid "Release {version} has been installed in \"{folder}\"."
```
```suggestion
msgid "Release {version} has been installed in [{folder}]."
```

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -3889,6 +3892,9 @@ msgstr "Actualiser"
 msgid "Refresh this page."
 msgstr "Actualiser la page."
 
+msgid "Release {version} has been installed in \"{folder}\"."
+msgstr "La nouvelle version {version} a été installée dans « {folder} »."
```
```suggestion
msgid "Release {version} has been installed in [{folder}]."
msgstr "La nouvelle version {version} a été installée dans [{folder}]."
```

src/common/engine.py
```diff
@@ -509,153 +510,169 @@ def _install_new_version(version: Version, download_url: str) -> bool:
...
+                return False
+            except Exception as ex:
+                logger.error('Unexpected error during installation: [%s]', ex)
+                return False
+
+        if input_interactive_yn(
+            _('Release {version} has been installed in "{folder}".').format(
+                version=version,
+                folder=new_version_dir.absolute(),
+            )
```
pascalaubry:
```suggestion
            _('Release {version} has been installed in [{folder}].').format(
                version=version,
                folder=new_version_dir.absolute(),
            )
```

pascalaubry (dismissed):
Last details.

src/common/engine.py
timothyarmes: It's not an error, there's no point downloading it again, instead we fall through

pascalaubry: Not an error but at least a warning (I don't think that people not downloading the new release because of an already existing folder will understand it if they are not warned).

pascalaubry (approved):

PR #1580, opened 2025-12-14
Fix `Modified since last upload` message on never uploaded tournaments

Molrn:
On new tournaments, you had the message `Modified since last upload`, even when the tournament had never been uploaded. Now: `Tournament not yet uploaded.`


timothyarmes (approved):
LGTM

PR #1594, opened 2025-12-17
Fix event ordering

Molrn:


src/data/loader.py
```diff
@@ -192,7 +192,7 @@ def get_events_metadata(
             key=lambda event: (
                 get_date_timestamp(event.stop_date) * sort_order,
                 get_date_timestamp(event.start_date) * sort_order,
-                normalized_key('name'),
```
timothyarmes, line 195: Why aren't you normalising it now?  We need to ensure that names with accents are sorted appropriately.

Molrn: Missed your comment ^^' we don't do that in any other sort, it's either everywhere or nowhere. For a fix I say nowhere, we could discuss everywhere, honestly I'm not sure it's that important

timothyarmes (approved):

PR #1595, opened 2025-12-17
Fix Mac build

timothyarmes:


src/utils/scripts.py
```diff
@@ -15,54 +15,56 @@ def default_workdir() -> Path:
...
-        # Case: macOS .app onedir
-        # .../My.app/Contents/MacOS/exe
-        if (
-            exe.parent.name == 'MacOS'
-            and exe.parent.parent.name == 'Contents'
-            and exe.parent.parent.parent.suffix == '.app'
-        ):
-            return exe.parent.parent.parent.parent  # -> parent of the .app bundle
         # Case: onefile or frozen onedir (not .app)
         return exe.parent
```
Molrn: `exe` is undefined for windows, I'm guessing prod windows would fail. I don't get how mypy passed on GH, it fails locally

Molrn (changes requested):

Molrn (approved):

PR #1606, opened 2025-12-21
Fixed the display of pairings on the FFE website for accelerated tournaments

pascalaubry:
image

### DIFF
--- a/src/plugins/ffe/papi_converter.py
+++ b/src/plugins/ffe/papi_converter.py
@@ -758,17 +758,15 @@ class PapiConverter:
             if len(group_max_numbers) > 0 and group_max_numbers[0] > 0:
                 rating_threshold_1 = max(
                     tournament.tournament_players_by_pairing_number[
-                        group_max_numbers[0] - 1
-                    ].rating
-                    - 1,
+                        group_max_numbers[0]
+                    ].rating,
                     0,
                 )
                 if len(group_max_numbers) > 1 and group_max_numbers[0] > 0:
                     rating_threshold_2 = max(
                         tournament.tournament_players_by_pairing_number[
-                            group_max_numbers[1] - 1
-                        ].rating
-                        - 1,
+                            group_max_numbers[1]
+                        ].rating,
                         0,
                     )
         return rating_threshold_1, rating_threshold_2
--- a/src/plugins/ffe/papi_converter.py
+++ b/src/plugins/ffe/papi_converter.py
@@ -755,18 +755,14 @@ class PapiConverter:
         if isinstance(variation, AccelerationSwissVariation):
             group_max_numbers: list[int] = variation.get_group_max_numbers(tournament)
             if len(group_max_numbers) > 0 and group_max_numbers[0] > 0:
-                rating_threshold_1 = max(
-                    tournament.tournament_players_by_pairing_number[
-                        group_max_numbers[0]
-                    ].rating,
-                    0,
-                )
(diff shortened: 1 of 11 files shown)


pascalaubry:
Closes #1604

src/plugins/hookspec.py
```diff
@@ -295,6 +295,12 @@ def get_tournament_tie_breaks_warning_message(
     ) -> str | None:
         """Warning message for the tie-breaks of a tournament."""
 
+    @hookspec(firstresult=True)
+    def get_tournament_pairing_settings_warning_message(
```
Molrn:
```suggestion
    def get_tournament_pairing_warning_message(
```
the warning is displayed on the pairing part of the card, it has nothing to do with the settings (to replace in other places too)

Molrn (dismissed):
Image  

There still is a difference (papi left, SC right) despite no player having the same rating: Christophe should have 2 points.

CheckMatt:
Why Christophe should have 2 points? Don't we try to have an A group with a even number of players?

By the way, I have seen in the French arbiter book that they suggest using Elo instead of pairings number (probably because they wrote just as Papi does)
image

Molrn:
> Why Christophe should have 2 points?

I just meant that there's a difference here between how it appears in Sharly Chess and in the exported Papi which should not be there, the exported Papi should give Christophe 2 points because SC give him 2 according to the input settings

CheckMatt:
Ok I understand, this is probably because Papi use `> Rating` whereas Pascal did as if it is `≥ Rating`

src/plugins/ffe/papi_converter.py
```diff
@@ -733,13 +745,38 @@ def write_papi_file(
...
+    ) -> tuple[int, int]:
+        rating_threshold_1: int = 0
+        rating_threshold_2: int = 0
+        variation: PairingVariation = tournament.pairing_variation
+        if isinstance(variation, AccelerationSwissVariation):
+            group_max_numbers: list[int] = variation.get_group_max_numbers(tournament)
+            if len(group_max_numbers) > 0 and group_max_numbers[0] > 0:
+                rating_threshold_1 = tournament.tournament_players_by_pairing_number[
+                    group_max_numbers[0] - 1
+                ].rating
```
CheckMatt, line 760: I think we can replace by `].rating - 1`

Molrn:
@CheckMatt But you are right there's a big problem with acceleration settings, and apparently it's been there for months... I'm looking into it, but:
- you should be able to configure group settings which you should be allowed to (see #1608 )
- Here group A is supposed to be 1-8, which would include Christophe GANDEMER's opponent, but it doesn't

CheckMatt:
> * Here group A is supposed to be 1-8, which would include Christophe GANDEMER's opponent, but it doesn't

I don't understand, you mean that the computation of SC should inclure GANDEMER's opponent? It's a bug that it doesn't?

Molrn:
nevermind, no problem with the ordering, here there just is an absent player on round 1 in group A. Sorry about that ^^'

Molrn (dismissed):

CheckMatt (approved):

CheckMatt:
@Molrn I don't understand why you remove le `-1` for the threshold ? :thinking:

Molrn:
we are live debugging it ^^ the problem was not a >= / > issue, it was that the wrong player was taken into account

Molrn (approved):
LGTM

PR #1607, opened 2025-12-21
Fixed 'remaining boards' label

pascalaubry:


Molrn (dismissed):
`0 parties en cours` still feels odd, I think replacing it by `(Ronde n°9, terminée)` would be more appropriate

Molrn (dismissed):
LGTM

PR #1613, opened 2025-12-23
Improve the display when no permission

pascalaubry:
There may be some rares cases where authenticated accounts can have more permissions than the anonymous account; in this case, display a clearer message.

### DIFF
--- a/src/web/templates/admin/event_layout.html
+++ b/src/web/templates/admin/event_layout.html
@@ -97,7 +97,7 @@
                     {{ _('You have no permissions') }}
                 </h5>
                 <p class="text-center">
-                    {{ _('You can ask the organizer for permission to access the event screens.') }}
+                    {{ _('You can ask the administrator for permission to access the event screens.') }}
                 </p>
             {% endif %}
         </div>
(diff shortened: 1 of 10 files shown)


src/web/templates/admin/event_layout.html
```diff
@@ -82,15 +82,24 @@ <h4 class="w-100 px-2 fs-5 mt-2 text-center">{{ admin_event.name }}</h4>
...
+                </p>
+                <p class="text-center">
+                    {{ _("If you have an account you may connect by clicking on the <i>Profile</i> button in the menu.") }}
+                </p>
+            {% else %}
+                <h5 class="my-5 text-center">
+                    {{ _('You have no permissions') }}
+                </h5>
+                <p class="text-center">
+                    {{ _('You can ask the organizer for permission to access the event screens.') }}
```
Molrn:
```suggestion
                    {{ _('You can ask the administrator for permission to access the event screens.') }}
```
Not necessarly the organizer, I think we should stick to this vocabulary here

Molrn (dismissed):

Molrn (approved):
LGTM

PR #1615, opened 2025-12-23
Improve account cards

pascalaubry:


Molrn (dismissed):
it is not clearer in the UI, and complexifies the code for no reason, I think we should not merge this one

pascalaubry:
> it is not clearer in the UI, and complexifies the code for no reason, I think we should not merge this one

There is one reason: when a permission is given for the single tournament of an event, it was written "for 1 tournament"; it has been replaced by all the tournaments, more readable.

Molrn:
I see, in that case I suggest hiding the tournament mentions when there only is one in the event (see #1620)  

image

pascalaubry:
Replaced by #1620

PR #1616, opened 2025-12-24
Add a warning tooltip on rankings table for no games played OTB

pascalaubry:
image


Amaras (dismissed):
Please manually compute the Bu. value before implementing a fix that isn't a fix.

This will introduce a desync with the FFE website, unless you also want to replace all Bu. and Tr. tie-breaks with manual ones.

Molrn (approved):
LGTM

PR #1620, opened 2025-12-24
Clarify the account cards with one tournament

Molrn:


pascalaubry (changes requested):
OK to replace #1615 by this one but actually there is still a confusion for users between roles and permissions (the goal of #1615 was to have roles and permissions at the same level to distinguish them).
Let's talk about it.

pascalaubry (approved):

PR #1626, opened 2025-12-28
Fixed the translation script

pascalaubry:


scripts/i18n/i18n_translate.py
```diff
@@ -216,13 +227,13 @@ def translate_string(
...
-                f'{percent}% The tokens have changed (the translation is deleted):'
+            print_interactive_warning(
+                f'{percent}% The tokens have changed (the translation should be manually updated):'
             )
-            print_interactive_error(
+            print_interactive_warning(
                 f'{percent}% {string} >>> {translated_string_with_tokens}'
             )
-            translated_string_with_tokens = ''
+            # translated_string_with_tokens = ''
```
Molrn:
```suggestion
```

scripts/i18n/i18n_translate.py
```diff
@@ -264,4 +275,4 @@ def translate_message(
 
 
 if __name__ == '__main__':
-    I18nTranslator('nl')
+    I18nTranslator('pt')
```
Molrn: should be a param of the script

Molrn (dismissed):

Molrn (approved):
Nothing to say on the script, works well. Comments on the doc, but still approving.

I think it would be clearer for i18n.md to only contain the locale statuses, and to refactor the rest of the document into the /contributing dir. 
- `Translators` would fit into `contributors.md`
- `Notes for the translators` deserves to be a standalone document in `contributing/translation.md` (content is IMO too important to be burried after all the statuses)
- `Notes for the devs` deserves too a standalone doc `contributing/development.md`, with in addition all the explanation to handle the pipeline:
  - installing the pre-commit hook
  - running i18n_update / the tests
  - maybe `migrate.py` which is rather important for devs too

pascalaubry:
I do agree with everything, merging this PR and starting a new one.

PR #1632, opened 2026-01-01
Fix session locale

pascalaubry:
Set the default locale if the session locale is wrong (happens when a translation is deleted)


Molrn (approved):
LGTM

PR #1633, opened 2026-01-01
Restrict reamining unauthorized tournament access

Molrn:
FFE / CR publish modal
CE import modal
Documents


pascalaubry (approved):
A bit confusing now not to see everything with a non administrator account but we will get used to it.
Really nice.

PR #1640, opened 2026-01-03
Batch update of timer hours dates

Molrn:


pascalaubry (dismissed):
The min hour should be `0` not `1` (did not check on previous releases).
image

pascalaubry (approved):

PR #1651, opened 2026-01-10
Archives deletion

Molrn:


src/web/controllers/admin/index_admin_controller.py
```diff
@@ -1081,6 +1081,28 @@ async def htmx_admin_restore_archive(
...
+    async def htmx_admin_delete_archive(
+        self,
+        request: HTMXRequest,
+        archive_name: str,
+    ) -> Template:
+        web_context = AdminWebContext(request, admin_tab='archives')
+        archive = ArchiveLoader.get_archive(archive_name)
+        if not archive:
+            raise NotFoundException(f'Unknown archive [{archive_name}]')
+        archive.file.unlink()
```
Amaras: This doesn't ensure that the file exists; it only ensures that there is a non-empty path.
Either add `missing_ok=True` to the `archive.file.unlink()` call, or ensure that the archive's file exists first.

Amaras (dismissed):
Okay, all good, except the slight possibility of a crash from a malicious user

pascalaubry (dismissed):
Good for me, just add a line to the release notes?

Amaras (approved):
LGTM, just do what Pascal said and we can merge

pascalaubry (approved):

PR #1653, opened 2026-01-11
Fix the text of player norm tooltips

pascalaubry:
image

image


src/web/templates/admin/print/norm_report.html
```diff
@@ -483,9 +483,9 @@ <h1>Certificate of Title Result (IT1)</h1>
                     </div>
                     <div class="item {{ 'error' if norm.score_too_low or norm.performance_too_low else '' }}">
                         <div class="error-info">
-                            {{ norm.score_too_low | safe }}
+                            {{ norm.score_too_low | safe | replace('%%', '%') }}
                             <br />
-                            {{ norm.performance_too_low | safe }}
+                            {{ norm.performance_too_low | safe  | replace('%%', '%') }}
```
Molrn: should be done in python where the i18n string is defined instead of being corrected in the template. also I think `preformance_too_low` was not impacted

Molrn (dismissed):
Considering it's a light fix, wouldn't it be more appropriate to put it to stable ?

src/web/templates/admin/print/norm_report.html
pascalaubry: Well, I was wondering if it should not be done in _() (this function is ours). Your opinion?

Molrn: fine by me, if you do so can you please also remove all the places we've put a .replace('%%', '%') ?

Molrn (approved):
LGTM

Molrn (approved):
LGTM

PR #1657, opened 2026-01-12
SCF plugin

timothyarmes:


src/plugins/scf/locale/en/LC_MESSAGES/scf.po
```diff
@@ -0,0 +1,37 @@
...
+msgstr ""
+
+msgid "Swiss Federation specific features"
+msgstr ""
+
+msgid "The player with SCF code [{scf_code}] already plays tournament [{tournament}]."
+msgstr ""
+
+msgid "Unrated *** SINGULAR"
+msgstr "msgid \"Unrated *** SINGULAR"
```
pascalaubry, line 36:
```suggestion
msgstr "Unrated"
```

src/plugins/scf/locale/fr/LC_MESSAGES/scf.po
```diff
@@ -0,0 +1,39 @@
...
+msgstr "Féderation Suisse des Échecs"
+
+msgid "Swiss Federation specific features"
+msgstr "Fonctionnalités spécifiques à la Féderation Suisse des Échecs"
+
+msgid "The player with SCF code [{scf_code}] already plays tournament [{tournament}]."
+msgstr "Le·la joueur·euse avec le code SCF [{scf_code}] joue déjà le tournoi [{tournament}]."
+
+msgid "Unrated *** SINGULAR"
+msgstr "Non-classé·e"
```
pascalaubry, line 38: Perfect with a NBSP.
```suggestion
msgstr "Non classé·e"
```

src/plugins/scf/scf_database.py
```diff
@@ -0,0 +1,484 @@
...
+        return _('SCF')
+
+    @staticmethod
+    def _dir() -> Path:
+        return scf.TMP_DIR
+
+    @property
+    def min_recovery_version(self) -> Version:
+        """The minimal app version for which the database can be recovered."""
+        return Version('3.4.0')
```
pascalaubry, line 113: Might also work with 3.4.0.
```suggestion
        return Version('3.5.0')
```

pascalaubry (changes requested):
Very nice, what a pleasure to see how our plugin infrastructure can easily host new plugins...

Two remarks:
1. I was surprised to see that the plugin has been activated by default when I started the app, is it normal?
2. It would be easier to distinguish federation plugins with a small indent between federations :
image
image

PR #1663, opened 2026-01-12
Sort federations by their 3-letter acronym

pascalaubry:
It seems to me more natural to sort the federation this way (acronyms are given by the federations themselved, not based on the EN translation).

Before:
image

After:
image


Molrn (dismissed):
Yep that's much nicer. In addition, wouldn't it be clearer if instead of the name being it english it'd be in the local language ? I feel like that's mostly the case when choosing a language

timothyarmes (dismissed):
I've just realised that these country names aren't translated!

The array should just contain the code, and the translation files should contain the translated country names.

pascalaubry:
I do agree: much better with localized strings.
But we only have 2 locales (3 soon), so the only way to do it IMO is manually translating the names of the countris in their main language and update `sharly_chess_config.py`.
In another PR?

src/common/sharly_chess_config.py
```diff
@@ -532,211 +532,211 @@ def local_url(self) -> str:
 
     # The federation names.
     federations: dict[str, str] = {
```
Molrn: as it's a constant, won't be translated back if you swap the app language, so it needs to be a property

Molrn (dismissed):

timothyarmes (approved):
FGTM

PR #1684, opened 2026-01-18
Reduce latency

Molrn:


pascalaubry (approved):
Works fine and I have no remark on the code, but @timothyarmes will be more relevant tha n me on this part.

PR #1691, opened 2026-01-19
Use the session results for player download

Molrn:


timothyarmes (approved):
Nice.  These change really go a long way to harmonising the entire code base.

PR #1694, opened 2026-01-20
Prevent refreshing not visible screens

Molrn:
screen now only refreshed when the tab is visible, + when it becomes visible to avoid consulting old data for 15s


pascalaubry:
I see how #1694 can (and obviously will) limit the number of requests but it also prevents us from running stress tests.
Shouldn't we refresh all the screens on TEST_ENV?

Molrn:
IG we won't be running stress test in TEST_ENV but more in DEV_ENV ? I'm ok with leaving it as such on dev env tho

pascalaubry (approved):

PR #1702, opened 2026-01-22
Save school on search and add other

Molrn:
Replaces #1698


pascalaubry (approved):
Nice, important to have it until we have bulk import.

PR #1705, opened 2026-01-22
Fix player replace form swap to create after search

Molrn:


pascalaubry (dismissed):
Not working here:
image
image
image

pascalaubry (dismissed):
Still one thing: when replacing a player, the FRA school should be preserved (more generally all the plugin data).

image

image

Molrn:
not all plugin data, just the new type of data I created "carry_over_data", with currently the tournament and the school. Plugins can add fields to this, for example the league, licence and licence number are overridden by the search. On it ;)

pascalaubry (approved):
OK for a second PR.

PR #1709, opened 2026-01-22
Fix the check-in filter of the Players tab

pascalaubry:
image


Molrn (changes requested):
that was on purpose, but we can remove it. IDK if all the other ones should be mandatory too, IMO we can just remove them entirely. "Absent next round" will not be there most cases and not having it does not bring anything. 

IMO it should either be none of them mandatory or just CHECKED_IN / NOT_CHECKED_IN

Molrn (approved):

PR #1715, opened 2026-01-24
Add options to Players Screens to display more players with minimal information

pascalaubry:
Minimal information is set by default by plugin `fra_schools`.

image

image

image

image

image


src/database/sqlite/event/migrations/m072_players_board_display.py
```diff
@@ -0,0 +1,47 @@
...
+            self.database.execute(
+                f'ALTER TABLE `{table}` ADD `players_board_format` INT'
+            )
+            self.database.execute(
+                f'ALTER TABLE `{table}` ADD `players_opponent_format` INT'
+            )
+            for params in {
+                (1, 1, 1, 0),
+                (4, 4, 5, 1),
+            }:
```
Molrn, line 29: Comment to add what this is

src/data/screen.py
```diff
@@ -512,17 +517,47 @@ def players_show_unpaired(self) -> bool:
...
+                if self.stored_screen:
+                    board_format = self.stored_screen.players_board_format
+                    assert board_format is not None
+                    return PlayersScreenBoardFormat(board_format)
+                else:
+                    if self.family is None:
+                        raise RuntimeError('Family reference unexpectedly None')
+                    return self.family.players_board_format
+            case _:
+                raise ValueError(f'type=[{self.type}]')
```
Molrn:
```suggestion
        if self.type != ScreenType.PLAYERS:
            raise ValueError(f'type=[{self.type}]')    
        if self.stored_screen:
            board_format = self.stored_screen.players_board_format
            assert board_format is not None
            return PlayersScreenBoardFormat(board_format)
        elif self.family is None:
            raise RuntimeError('Family reference unexpectedly None')
        return self.family.players_board_format
```
Those big useless match cases have to go, they add unnecessary complexity to the code

src/data/screen.py
```diff
@@ -512,17 +517,47 @@ def players_show_unpaired(self) -> bool:
...
+                if self.stored_screen:
+                    opponent_format = self.stored_screen.players_player_format
+                    assert opponent_format is not None
+                    return PlayersScreenOpponentFormat(opponent_format)
+                else:
+                    if self.family is None:
+                        raise RuntimeError('Family reference unexpectedly None')
+                    return self.family.players_opponent_format
             case _:
                 raise ValueError(f'type=[{self.type}]')
```
Molrn:
```suggestion
        if self.type != ScreenType.PLAYERS:
            raise ValueError(f'type=[{self.type}]')    
        if self.stored_screen:
            opponent_format = self.stored_screen.players_opponent_format
            assert opponent_format is not None
            return PlayersScreenOpponentFormat(opponent_format)
        elif self.family is None:
            raise RuntimeError('Family reference unexpectedly None')
        return self.family.players_opponent_format
```

src/data/screen.py
```diff
@@ -512,17 +517,47 @@ def players_show_unpaired(self) -> bool:
...
 
     @property
-    def players_show_opponent(self) -> bool:
+    def players_player_format(self) -> PlayersScreenPlayerFormat:
```
Molrn, line 518: same as the other screen properties

src/database/sqlite/event/event_database.py
```diff
@@ -1565,7 +1571,13 @@ def _write_stored_screen(
...
             if stored_screen.type == 'players'
             else None,
-            stored_screen.players_show_opponent
+            stored_screen.players_player_format
+            if stored_screen.type == 'players'
+            else None,
+            stored_screen.players_board_format
+            if stored_screen.type == 'players'
+            else None,
+            stored_screen.players_opponent_format
```
Molrn, line 1580: that kind of screen type logic has nothing to do in database.py

src/web/controllers/admin/base_event_admin_controller.py
```diff
@@ -369,3 +396,45 @@ def _render_modal(template_name: str, template_context: dict[str, Any]):
...
+            or PlayersScreenOpponentFormat.NAME_RATING_TYPE_POINTS
+        )
+
+    @staticmethod
+    def get_default_players_screen_columns(
+        event: Event,
+    ) -> int:
+        return (
+            plugin_manager.hook_for_event(event, 'get_default_players_screen_columns')()
+            or 1
```
Molrn: it was empty before, I think it should stay that way

Molrn (dismissed):

src/database/sqlite/event/event_database.py
pascalaubry: OK but what about all the similar tests for the other params?

Molrn: I thought the logic was duplicated, it appears it's just here, IG let's keep it here for now.

Molrn (approved):

PR #1720, opened 2026-01-25
Save non-search data when replacing a player

Molrn:


src/plugins/ffe/utils.py
```diff
@@ -327,6 +328,8 @@ def from_form_data(
         )
 
     def to_form_data(self, action: str | None = None) -> dict[str, str]:
+        if action == FormAction.REPLACE:
+            return {}
```
pascalaubry, line 332: FFE data is not set with these two lines:
```suggestion
```
image

image

image

pascalaubry (dismissed):

pascalaubry (approved):

PR #1725, opened 2026-01-26
Extend the search to clubs and schools

Molrn:


pascalaubry (approved):
Nice to search on the school full name, the city is taken into account.

image

PR #1736, opened 2026-01-28
Use string enums for Gender, Title and FFE Licence

Molrn:
This is a step towards player import, in order to have in our databases the same data we will ask to import (we don't wanna ask users to know our enum values, but instead clear understandable values).   
example with gender: 0, 1, 2 --> '', 'F', 'M'


pascalaubry (approved):
Nice refactor, and nice fix on m072.

PR #1745, opened 2026-02-02
Remove table header transparency

Molrn:


pascalaubry (approved):
Not sure it needs to be in the changelog but OK with it.

Issue #1738, opened 2026-01-29
Dynamic crosstable - Display player pairings sum up on hover


adrienave:
Currently, generated crosstable from _Documents_ tab has purely static content (probably because the focus was on the print side).

I think it would be nice if the crosstable could display dynamic content such as tournament logs of a player on hover (e.g. pairing for each round with associated result).

This way, it would be more consistent with the crosstable generated by PAPI (see screenshot below), and it would be nice for people (like me) who like to upload documents directly to their websites and let players interact with them.

<img width="902" height="423" alt="Image" src="https://github.com/user-attachments/assets/2561037e-a9a5-45cf-884f-4c7b89d01579" />

_In this screenshot, mouse pointer is hovering top player of the crosstable._

Molrn:
implemented in #1750

PR #1750, opened 2026-02-04
Show player history on crosstable player hover

adrienave:
Covers #1738.

As discussed on the Discord server, feature is only an option that is disabled by default.
The feature is not present on screens and should show up only for crosstable print document (with `player-history-popover` on).

Add new test type: `integration` type to test controllers.


conftest.py
```diff
@@ -217,3 +260,55 @@ def api_request_context(
...
+
+    jinja_template_engine = JinjaTemplateEngine(template_dirs)
+    jinja_template_engine.engine.add_extension('jinja2.ext.i18n')
+    jinja_template_engine.engine.install_gettext_callables(
+        gettext=gettext, ngettext=ngettext, newstyle=True
+    )
+    jinja_template_engine.engine.add_extension('jinja2.ext.do')
+    template_config: TemplateConfig = TemplateConfig(engine=jinja_template_engine)
+    app = Litestar(
+        route_handlers=[
```
adrienave: This is ugly and partially duplicated code from `settings.route_handlers`, but unfortunately directly importing `settings.route_handlers` fails since settings module does more things at top level like creating `SharlyChessEnvironment`, conflicting with test environment.

Fixing this would need deeper investigation / refactoring.

conftest.py
```diff
@@ -217,3 +260,55 @@ def api_request_context(
...
+
+
+# Integration tests fixtures
+
+
+@pytest.fixture(scope='session')
+def test_client() -> Iterator[TestClient[Litestar]]:
+    template_dirs: list[Path] = [
+        BASE_DIR / 'src/web/templates',
+        BASE_DIR / 'src/web/templates/admin/print',
```
adrienave: No idea why `admin/print` subpath has to be added, it should be able to find the templates recursively from `src/web/templates`, but tests aren't.

Maybe I'm missing an option on the template engine to loop over subfolders?

src/web/templates/admin/print/cells/name.html
```diff
@@ -0,0 +1,8 @@
+<div class="player-name">
+    {{ object.full_name }}
+    {% set admin_tournament = object.tournament %}
+    {% set player = object %}
+    <div class="player-history">
+        {% include '/admin/players/history_popover.html' %}
```
Molrn: Image  

few modifications required in this template:  

- bold font for headers, points and round    
- only iterate until the round for which the document is printed to avoid 'unpaired' in every row

src/web/templates/admin/event/print_options/player_history_popover.html
```diff
@@ -0,0 +1,8 @@
+<div class="col-12">
+    {{ admin_macros.switch_input(
+        name=option.id,
+        label='',
+        placeholder=_('Include player history popover'),
```
Molrn:
```suggestion
        placeholder=_('Display history on hover'),
        tooltip_text=_(
            'When hovering a player\'s name with the mouse, display '
            'its history. Warning: the document takes much longer to load.'
       )
        
```
"Popover" is a too technical word, + this is a tooltip and not a popover (see the difference [here](https://www.pierre-giraud.com/bootstrap-apprendre-cours/tooltip-popover/#:~:text=toooltip()%20avec%20jQuery.-,Les%20popovers,-Les%20popovers%20sont), popovers need to be clicked). 

Also on big tournaments this can take up to 30 seconds to render, so there needs to to be a warning about that.

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -2630,6 +2630,9 @@ msgstr "Inclure les joueur·euses"
 msgid "Include private player data (emails, phone numbers, etc.)"
 msgstr "Inclure les données privées des joueur·euses (mél, tél, etc.)"
 
+msgid "Include player history popover"
+msgstr "Inclure l'historique du joueur au survol"
```
Molrn: We use inclusive writing, so this'd be `du·de la joueur·euse` or `des joueur·euses`. As this is very verbous, I suggested bellow the following:
`Display history on hover`
`Afficher l'historique au survol`  
with a descriptive tooltip

src/data/columns/handlers.py
```diff
@@ -108,12 +108,16 @@ def get_player_crosstable_columns(
...
+        include_player_history_popover: bool = False,
     ) -> list[TournamentPlayerTableColumn]:
         return self.get_columns(
             [
                 pt.RankColumn,
                 pt.TitleColumn,
-                pt.NameColumn,
+                partial(
+                    pt.NameColumn,
+                    include_player_history_popover=include_player_history_popover,
```
Molrn: it feels wrong to propagate the parameter this deep. I suggest adding `include_player_history_popover` as a context variable using the `template_context` property of the document, and have the column just always being a template

conftest.py
Molrn, line 1: really not a fan of this new integration test, it seems high maintenance for not much. I think it'd be much preferable to do an e2e instead

tests/integration/test_event_documents_controller.py
```diff
@@ -0,0 +1,71 @@
...
+    yield event.tournaments_by_name[TOURNAMENT_NAME]
+
+
+@pytest.fixture(scope='function', autouse=True)
+def account(
+    event: Event, tournament: Tournament, test_client: TestClient
+) -> Iterator[Account]:
+    user_account = TestUtils.create_account(event.uniq_id, 'admin', overrides={'id': 3})
+    TestUtils.create_permission(
+        event.uniq_id, tournament.id, user_account.id, 'ADMINISTRATION'
```
Molrn: There is not admin account except the server user on http://127.0.0.1. Check all the e2e tests, they all are admins by default

tests/integration/test_event_documents_controller.py
Molrn, line 1: I suggest moving to `tests/e2e/test_documents.py`. The tests would be on the documents themselves, no reason to have it especifically on the controller

conftest.py
Molrn: Not mergeable as is, especially in a PR about something else. I suggest doing an e2e to test the doc

Molrn (changes requested):
Details on the feature, problems with the tests. 

Also you'd need the pipeline to pass (failed [here](https://github.com/Sharly-Chess/sharly-chess/actions/runs/21690718085/job/62549494556)), currently it fails because of the last ruff update (see #1751). until it's merged, you can manually install the latest ruff version:
```sh
pip install ruff~=0.15.0 
ruff format .  
```

src/web/templates/admin/print/cells/name.html
adrienave: Ok, makes perfect sense.
For points, I added the bold font on the shared template, since on the place where it's used originally, the player grid also have points in bold. Is it good?

For iterating until selected round, I completely agree with it as well, but since the logic is on the original template, what do you advise to avoid impacting the ranking tab? Would a `max_round` / `selected_round` variable overriding `admin_tournament.rounds + 1` when present be ok?

src/web/templates/admin/event/print_options/player_history_popover.html
adrienave: Interesting, I took popover for consistency with the name of the already existing template, which is also a tooltip in fact...

I'm going to refactor my references to it to `tooltip`.

src/web/templates/admin/print/cells/name.html
Molrn: yes a `max_round` context variable would be perfect ;) you can set it to `admin_tournament.rounds` in the template context of the method `htmx_admin_history_popover`

adrienave (commented):
I think I applied all your suggested changes, the PR is ready to be checked again.
Thanks for your hints, I learn a lot about the project.

I replaced integration tests by E2E tests ; it feels weird to me to test a feature only by very high level tests, ideally the project would benefit from more unit tests, and I thought integration tests could be a good intermediate.
But I understand the initial setup can feel like a burden.

tests/test_config.py
```diff
@@ -1,24 +1,24 @@
 """Test configuration and utilities."""
 
+import re
```
adrienave, line 3: Imports here and in ` conftest.py`  have been optimized by Pycharm, but I can revert these changes if you want since my PR is not about these files.

Molrn (approved):
Very clean PR and much appreciated feature, thank you very much ! LGTM :)

PR #1753, opened 2026-02-05
Distribute the players among the tournaments

pascalaubry:
image

image

image

### DIFF
--- a/src/data/columns/players_tab.py
+++ b/src/data/columns/players_tab.py
@@ -266,7 +266,7 @@ class RatingPlayersTabColumn(PlayersTabColumn):
         return True
 
     def _get_sort_key(self, player: Player) -> tuple:
-        return (player.single_tournament_player.rating,)
+        return (player.single_tournament_player.starting_rank_sort_key,)
 
     @property
     def cell_template(self) -> str | None:
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # _Sharly Chess_ release notes
 
+## Tournaments
+
+- Distribute the players among the tournaments (3.6.0)
+
 ## Pairings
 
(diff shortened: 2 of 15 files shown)


src/data/columns/players_tab.py
```diff
@@ -266,7 +266,7 @@ def is_tournament_column(self) -> bool:
         return True
 
     def _get_sort_key(self, player: Player) -> tuple:
-        return (player.single_tournament_player.rating,)
+        return (player.single_tournament_player.starting_rank_sort_key,)
```
Molrn, line 269:
```suggestion
        return (player.single_tournament_player.rating,)
```
rank sort key is the opposite order, so now the asc sort appears desc.

Molrn (changes requested):

src/data/columns/players_tab.py
pascalaubry: It is intentional: when you sort by rating, you are interested by the highest rated players (that was the case until a recent change on the Players tab).
Note: the goal of using `starting_rank_sort_key` instead of `rating` was not getting the correct order without having to click twice (the goal was to distribute the players the same way they are shown on the Players tab).

Molrn: image    

this icon means ascending, sorting by ascending sorts asc has to sort in a asc way. Consistency (not just amongst our columns, but also with the tabular app standards, clicking once for asc 2 for desc is standards) is more important than optimizing a usage that way. Clicking twice is not a hard enough price to pay to justify breaking standards IMO  

> the goal was to distribute the players the same way they are shown on the Players tab)

If so the distribution should adapt to the players tab, not the other way around. Distributing players by `key=(rating, last_name, first_name), reversed=True`  seems more than reasonable

Molrn:
replaced by #1757

Issue #871, opened 2025-07-30
Replace timestamps by date / datetime objects


Molrn:
The dates and datetimes being stored as timestamps in the DB should not condition their usage in the rest of the codebase. `date` / `datetime` objects are much cleaner to handle
It needs to be defined at Stored Object level, while keeping (for now) the timestamps in the DB

### Example
```py
class StoredBoard:
    # last_result_update: float | None
    last_result_update: datetime | None
```

### Tracking  
- [x] board.last_result_update (#1761)
- tournament (warning: SQL triggers)
  - [ ] last_update
  - [ ] last_player_update
  - [ ] last_pairing_update
- [ ] screen_set.last_update
- [ ] screen.last_update
- [ ] family.last_update
- To consider: timer.triggered_at

Amaras:
I'd recommend doing the same for database datetimes (see: https://sqlite.org/lang_datefunc.html)

Molrn:
@arun1587 let's decide the format here ;)

> I'd recommend doing the same for database datetimes  

@Amaras (not reacting 6 months too late ^^'): not sure we should change the format stored in the DB, considering we now have triggers depending on it, it's a bit of a maintenance issue. 
changing the database format means: 
- a sub-second precision is required (some of our fields need them)  
- recreating all the triggers
- for timers, ensure that the stored date has the same value as the one imput by the user and as the one converted in JS (AirDatepicker values can go crazy)

Keeping timestamps is IMO much safer, and is more a code-cleaning task to avoid handling timestamps manually than a risky type overall to change the types everywhere. Let me know how important you think it is, but bear in mind that the cost and potential maintenance issuea really are not the same

Amaras:
The first point is not problematic, since there is a `'subsecond'` argument to the datetime functions in SQLite.
For instance, here is what it looks like for a tournament I did in December:
```sql
SELECT board.id, board.white_player_id, board.black_player_id, board.`index`, datetime(board.last_result_update, 'unixepoch', 'subsecond') as last_updated
FROM board
LIMIT 5
OFFSET 20;
```
| id | white_player_id | black_player_id | index | last_updated |
|----|-----------------|-----------------|-------|--------------|
| 71 | 53 | 96 | 20 | 2025-12-06 14:41:45.985 |
| 72 | 57 | 83 | 21 | 2025-12-06 14:41:46.247 |
| 73 | 88 | 84 | 22 | 2025-12-06 14:41:46.505 |
[...]

PR #1761, opened 2026-02-08
chore: add datetime conversion helpers and convert Board.last_result_…

arun1587:
…update (#871)

Add  

-  timestamp_to_datetime, 
-  datetime_to_timestamp, 
-  timestamp_to_date, and 
-  date_to_timestamp
 helpers to utils/date_time.py. 

Convert `Board.last_result_update` from float to datetime, and update the results filtering in Screen to use datetime/timedelta comparisons.

This is the first step toward replacing all timestamp properties with datetime objects in the application layer (lets say PR #1 of #871, to keep the code churn at minimal).


tests/unit/test_date_time.py
Amaras, line 1: If you want to test obviously correct code, you might want to look into the [hypothesis](https://hypothesis.readthedocs.io/en/latest/) library.

I want to start thinking about integrating property-based testing anyway, so this might be the time to do it.

Amaras (commented):
Hi @arun1587, thank you for your interest in our project.

From a cursory look, I think your PR is missing something here, because *all* time values in the database are in the (floating-point) timestamp format.
I don't know if this is enough to cover all `Board` uses, especially without a database migration.

I haven't taken the time to study what the changes entail yet; one of us will report back when we've looked at it more closely

src/utils/date_time.py
```diff
@@ -60,3 +60,19 @@ def format_date_range(start_date: date, stop_date: date | None = None) -> str:
...
+def datetime_to_timestamp(dt: datetime | None) -> float | None:
+    return dt.timestamp() if dt is not None else None
+
+
+def timestamp_to_date(ts: float | None) -> date | None:
+    return datetime.fromtimestamp(ts).date() if ts is not None else None
+
+
+def date_to_timestamp(d: date | None) -> float | None:
+    return get_date_timestamp(d) if d is not None else None
```
Molrn:
```suggestion
```
We are trying to avoid dead code, so utils functions are only added when needed

tests/unit/test_date_time.py
Molrn, line 1: Tests verifying if / else conditions and outer libs are outside of the scope of tests we want to include to the project, please delete this file

Molrn (changes requested):

src/database/sqlite/event/event_database.py
```diff
@@ -1282,14 +1282,18 @@ def update_board_last_result_update(
...
             return None
         else:
-            date = time.time()
+            self.execute(
+                "UPDATE `board` SET `last_result_update` = datetime('now') WHERE `id` = ?",
+                (board_id,),
+            )
 
             self.execute(
-                'UPDATE `board` SET `last_result_update` = ? WHERE `id` = ?',
```
arun1587, line 1288: this can be simplified with returning clause, if we have min sqlite version > 3.35 iirc, enforcement in the project

arun1587:
@Amaras, you were right. i  think i missed your comment about the db level change, on the GH issue. i updated the PR with the db migration change (for board) as it removes thin utilities at the business logic layer and simpler. I can revisit the PR by end of the week if needed further changes.

src/database/sqlite/event/migrations/m074_convert_board_last_result_update_to_datetime.py
Molrn, line 1: We have triggers and automations based on datetimes being stored as floats in the DB. Having timestamps in the DB is not wrong in itself, it's more handling timestamps in our code that is. Considering the complexity and potential maintenance issues coming from this (not especially on this field but on all the other ones), I really think we should avoid changing the type of those DB values. Sorry I missed @Amaras's comment, would have reacted sooner if I understood that's what you were doing

Basically, timestamps in the DB are ok, but everywhere else it should be datetimes. It needs to be the same as for dates (you can take `tournament.start_date` for reference):  
- `datetime` (or `datetime | None`) type in the StoredClass (ex: StoredBoard)
- helper functions `{load_[optional_]timestamp_from_database_field() -> datetime [| None]`  and `dump_datetime_to_timestamp_field` similar to `load_date_from_database_field`
- no conversion in the wrapper class properties (ex: Board)

src/data/board.py
```diff
@@ -120,15 +121,15 @@ def result_str(self) -> str:
         return str(self.result)
 
     @property
-    def last_result_update(self) -> float | None:
-        return self.stored_board.last_result_update
+    def last_result_update(self) -> datetime | None:
+        if self.stored_board.last_result_update:
+            return datetime.fromisoformat(self.stored_board.last_result_update)
```
Molrn: stored_board.last_result_update should already have type datetime | None

Molrn (changes requested):

Molrn (approved):
LGTM

PR #1769, opened 2026-02-10
Add the rating range to the tournament cards

pascalaubry:
image

### DIFF
--- a/src/web/templates/admin/tournaments/tournament_card.html
+++ b/src/web/templates/admin/tournaments/tournament_card.html
@@ -66,7 +66,7 @@
             <div class="text-start">
                 {{
                     _(
-                        '%(rating)s (%(rating_type)s): %(min)d→%(max)d',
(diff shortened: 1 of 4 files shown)


src/web/templates/admin/tournaments/tournament_card.html
```diff
@@ -64,7 +64,21 @@
     <div class="card-block">
         <div class="card-block-content">
             <div class="text-start">
-                {{ tournament.rating }} ({{ tournament.player_rating_type.name }})
+                {{
+                    _(
+                        '%(rating)s (%(rating_type)s): %(min)d→%(max)d',
```
Molrn:
```suggestion
                        '%(rating)s (%(rating_type)s): %(min)d → %(max)d',
```
Feels a bit chunky as is, + does not line break

Molrn (dismissed):

Molrn (approved):
LGTM

PR #1770, opened 2026-02-10
Remove page breaks from prizes documents

pascalaubry:


Molrn (changes requested):
Image  

+ assignment doc not working (looking into it)

pascalaubry:
The only way I see to get rid of this empty page is to put title and subtitle from `base.html` into the first no-breakable item, but it implies all the documents.
Let's talk about it.

Molrn:
I don't get what the PR is trying to do. IMO only categories should have page breaks, I don't get why you'd put one on the tournament itself

pascalaubry:
> I don't get what the PR is trying to do. IMO only categories should have page breaks, I don't get why you'd put one on the tournament itself

The goal is not to have that:
image

Amaras (commented):
Why not simply add this CSS property?
```css
.prize-table {
    break-inside: avoid-page;
}
```

It would keep each group together and still allow for page breaks between groups.

Molrn:
image  
did the best I could, there still is this category issue remaining on the assignment doc, but with the grid layout I don't see how we could prevent that

pascalaubry:
> did the best I could, there still is this category issue remaining on the assignment doc, but with the grid layout I don't see how we could prevent that

Much better than before, thanks.

PR #1775, opened 2026-02-12
chore: Convert stored timestamp fields from float to datetime (#871)

arun1587:
The PR wraps up the code cleaning targets specified in https://github.com/Sharly-Chess/sharly-chess/issues/871

- Change StoredFamily, StoredScreen, StoredScreenSet, and StoredTournament timestamp fields (last_update, last_player_update, last_pairing_update) from float to datetime | None
- Convert float→datetime at the DB read boundary using load_optional_timestamp_from_database_field() in _row_to_stored_* methods; DB write path remains as float
- Update wrapper classes (Family, Screen, ScreenSet, Tournament) to return datetime | None directly and use format_datetime() for display
- Add .timestamp() bridges at comparison sites for uploaders and screen poller.

### DIFF
--- a/src/data/screen_set.py
+++ b/src/data/screen_set.py
@@ -463,7 +463,9 @@ class ScreenSet:
             return self.stored_screen_set.last_update
         else:
             assert self.family is not None
-            return self.family.last_update
+            return (
+                self.family.last_update.timestamp() if self.family.last_update else None
+            )
 
     @property
     def last_update_str(self) -> str:
--- a/src/data/screen_set.py
+++ b/src/data/screen_set.py
@@ -9,7 +9,7 @@ from common.i18n import _
 from data.board import Board
 from data.player import TournamentPlayer
 from datetime import datetime
-from utils.date_time import format_datetime
+
 from utils.enum import (
     ScreenType,
     PlayersScreenPlayerFormat,
@@ -465,10 +465,6 @@ class ScreenSet:
         else:
             assert self.family is not None
             return self.family.last_update
-
-    @property
-    def last_update_str(self) -> str | None:
-        return format_datetime(self.last_update) if self.last_update else None
 
     @property
     def numbers_str(self) -> str:
--- a/src/data/screen_set.py
+++ b/src/data/screen_set.py
@@ -8,7 +8,8 @@ from _weakref import ReferenceType
 from common.i18n import _
 from data.board import Board
 from data.player import TournamentPlayer
-from utils.date_time import format_timestamp_date_time
+from datetime import datetime
+from utils.date_time import format_datetime
 from utils.enum import (
     ScreenType,
     PlayersScreenPlayerFormat,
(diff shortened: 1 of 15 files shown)


src/data/screen_set.py
```diff
@@ -460,16 +461,16 @@ def last_tournament_player_by_rank(self) -> TournamentPlayer | None:
...
         if self.stored_screen_set:
             return self.stored_screen_set.last_update
         else:
             assert self.family is not None
             return self.family.last_update
 
     @property
-    def last_update_str(self) -> str:
-        return format_timestamp_date_time(self.last_update)
+    def last_update_str(self) -> str | None:
```
Molrn: all the last_update_str are unused, they can be removed

src/plugins/chess_results/chess_results_background_uploader.py
```diff
@@ -127,9 +127,13 @@ def chess_results_upload_needed(
         cls, tournament: Tournament | StoredTournament
     ) -> bool:
         return cls.chess_results_last_upload(tournament) < max(
-            tournament.last_update,
-            tournament.last_player_update,
-            tournament.last_pairing_update,
+            tournament.last_update.timestamp() if tournament.last_update else 0,
```
Molrn: the idea being to avoid using timestamps, converting to timesamps for this is not ideal. a `tournament.last_update or datetime.min` is preferable.   

Same for all the value comparisons of the PR. Basically it's not just the direct usages of these variables as timestamps which should adapted, but also all the indirect usages.

src/plugins/chess_results/chess_results_background_uploader.py
```diff
@@ -127,9 +127,13 @@ def chess_results_upload_needed(
         cls, tournament: Tournament | StoredTournament
     ) -> bool:
         return cls.chess_results_last_upload(tournament) < max(
```
Molrn: here `chess_results_last_upload` should return a datetime. This highlights a few variables variables missing  from the scope of #871 : the plugin data variables.  

You have:
- ChessEventTournamentPluginData.last_sync
- ChessResultsTournamentPluginData.last_upload
- FfeTournamentPluginData.last_rules_upload/last_upload

For each of these the values can be fetched using the database timestamp load / dump utils in the from_stored_value / to_stored_value methods

Molrn (changes requested):

src/web/controllers/user/screen_user_controller.py
```diff
@@ -104,10 +112,10 @@ def _user_screen_refresh_needed(
             assert web_context.family.event is not None
             if (
                 max(
-                    web_context.family.event.last_update,
-                    web_context.family.last_update or 0,
+                    datetime.fromtimestamp(web_context.family.event.last_update),
```
Molrn: event.last_update should have datetime type

src/plugins/chess_results/utils.py
```diff
@@ -294,7 +294,9 @@ def from_stored_value(cls, stored_value: dict[str, Any]) -> Self:
             auto_upload=stored_value.get('auto_upload', None),
             remark=stored_value.get('remark'),
             remark_default=stored_value.get('remark_default', True),
-            last_upload=stored_value.get('last_upload', 0.0),
+            last_upload=datetime.fromtimestamp(ts)
```
Molrn: please use the database timestamp dump and load functions to handle it the same way as the other fields

Molrn (changes requested):

Molrn (approved):

PR #1793, opened 2026-02-18
Add test and example events

pascalaubry:


examples/events/domloup-fide-37.db
Molrn, line 1: .db file, to remove. other .db too

Molrn (dismissed):
### Examples  

I think this much events created is actually less helpful when developping, and a bit overwhelming. It's nice to have a handful of selected events that represent the diversity of what the app can do, and in that our current selection is actually not so bad (mutli-tournaments with domloup, big with the France, scholar, handicap, complex prizes with the bretagne, etc...).  

If we want other events to be accessible easilly, I suggest creating a subfolder `other` and putting them here.  This way they can still easilly be copied, but aren't created as examples. 

### Testing   

I think it'd be nice to have a clear rule of one event per 3.x version, without forgetting to add a new one each time we release a new version

Molrn (approved):
LGTM

PR #1799, opened 2026-02-19
Fix various norm bugs

Amaras:
This PR aimed to fix various bugs in the player norm computation code.

1. The "NON" federation was not properly handled, since `str` and `Federation` do not compare properly.
2. There is a subtle bug in the count of players in 1.4.3.d and 1.5.6.a.

&nbsp;   1. The FIDE Handbooks count players in if they miss at most one round (excluding PAB/Rest games)

&nbsp;   2. We count them in if they don't miss any game (excluding PAB/Rest games)

Both are fixed in this PR.

NOTE: while writing this message, I got a phone call from Pierre Lapeyre about erroneous averages, so I will try to get the right values in this PR.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,6 +20,11 @@
 - Upgrade bbpPairings engine to v6.0.0 (3.6.0)
 - Display player history on crosstable document (3.6.0)
 
+## Norms
+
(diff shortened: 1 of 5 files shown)


timothyarmes (approved):
LGTM

PR #1800, opened 2026-02-19
fix: correct plugin data field names in chessevent and FFE templates …

arun1587:
…(#871)

the plugin data(ffe and chessevent) field usages in templates were updated with appropriate fields.


timothyarmes (approved):
LGTM

PR #1805, opened 2026-02-20
Remove documents options when not applicable to any tournaments

Molrn:
- Norm report only available if there are titled players
- Berger grid only if there is a RR tournament
- All tournament docs if there is a tournament


pascalaubry (dismissed):
Documents link disappeared:

Image

And also please have a look at the tests.

pascalaubry (dismissed):

Issue #1778, opened 2026-02-14
Cleanup: Convert DB timestamp columns from FLOAT to ISO 8601 TEXT (#871 Phase 2)


arun1587:
## The Focus

The event database stores all generated timestamps (last_update, last_player_update, last_pairing_update, last_result_update,..) as **FLOAT** values (Unix epoch). When inspecting the database for debugging or auditing, values like `1739512345.123456` are opaque and require manual conversion to understand.

Additionally, triggers manage timestamp updates for tournament change tracking. These triggers are invisible to Python, making the timestamp logic harder to test, debug, and maintain.

Phase 1 already converted all timestamp fields from `float` to `datetime` at the **Python level**. The DB still stores FLOAT and the Python layer converts on read/write, creating a mismatch between the application type (`datetime`) and the storage type (`FLOAT`).

### Convert the FLOAT timestamp columns to TEXT (ISO 8601 UTC)

Store timestamps as human-readable UTC strings in the database. Example: `2025-02-14 09:30:45.123456` instead of `1739512345.123456`.

**Columns in scope:**

| Table | Column | Sub-second precision |
|-------|--------|:--------------------:|
| `tournament` | `last_update` | yes |
| `tournament` | `last_player_update` | yes |
| `tournament` | `last_pairing_update` | yes |
| `board` | `last_result_update` | yes |
| `screen` | `last_update` | no |
| `screen_set` | `last_update` | no |
| `family` | `last_update` | no |
| `display_controller` | `last_update` | no |

@Molrn please correct me if i'm missing other targets or characteristics here.

Sub-second precision is preserved for the columns that currently use it (via `unixepoch('subsec')` in triggers) — these are timing-sensitive for screen refresh polling. The remaining columns are set on manual config saves where second-level precision is sufficient.

All values are stored in **UTC** (matching current behavior — Unix epoch is inherently UTC). 

### Replace timestamp triggers with Python helpers

Move timestamp update logic from SQLite triggers into explicit Python methods in `EventDatabase`, making it testable and visible in the application code.

**Triggers to remove:**

| Trigger | Current action |
|---------|---------------|
| `set_tournament_last_pairing_update_on_pairing_insert` | Sets `last_pairing_update` on pairing insert |
| `set_tournament_last_pairing_update_on_pairing_update` | Sets `last_pairing_update` on pairing update |
| `set_tournament_last_pairing_update_on_pairing_delete` | Sets `last_pairing_update` on pairing delete |
[...]

PR #1807, opened 2026-02-20
chore(#1778): Convert timestamp columns from FLOAT to UTC ISO TEXT

arun1587:
- added migration 074 to address the timestamp conversion in the triggers (plugin data as well addressed)
- ensure we use utc across the fields

### DIFF
--- a/src/database/sqlite/event/event_database.py
+++ b/src/database/sqlite/event/event_database.py
@@ -1,5 +1,4 @@
 import shutil
-import time
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


pascalaubry (changes requested):
Please rename your migration to m075 after #1814 is merged.
Too database-oriented for me, @Molrn will be much more relevant than me for this one.

Molrn (approved):
All is working fine, migration is clean ;) thanks a lot !

PR #1808, opened 2026-02-21
Fix slowness after document generation

Molrn:


pascalaubry (dismissed):
I approve because it works for me but I am not able to say this is the right fix.

PR #1809, opened 2026-02-21
Do not open new browser tabs when refreshing documents

pascalaubry:


pascalaubry:
Couldn't make it work properly (the form target attribute expects an iframe ID, opens in a new tab if not found but the tab is not rerfreshed when already opened).

PR #1827, opened 2026-02-28
Improve and fix session storage

Amaras:
Thanks to Gilles, I improved a few things:

* Fixed the wrong conditions and syntax in SQL queries
* Reduced the use of read-only session usage (do not use them on /view, /ws and favicon)
* Add explicit PEP-249 compliant transactions


src/web/settings.py
```diff
@@ -263,6 +263,9 @@ async def create_connection(path: os.PathLike[str]) -> aiosqlite.Connection:
         exclude=[
             r'^/static/*',
             r'.*\.(png|jpg|jpeg|gif|css|js|svg)$',
+            r'^/view/*',
+            r'^/ws$',
+            r'^/favicon.ico$',
```
Molrn:
```suggestion
            r'.*\.(png|jpg|jpeg|gif|css|js|svg|ico)$',
            r'^/ws$',
```
/ws is ok, however /view accesses session variables so it needs to go through the middleware (500 on view otherwise). Basically anything that renders a template should not be excluded

Molrn (changes requested):

src/web/sqlite_store.py
```diff
@@ -93,7 +108,7 @@ async def expires_in(self, key: str) -> int | None:
                 'SELECT expires_at FROM store where key = ?', parameters=(key,)
             ) as cursor:
                 row = await cursor.fetchone()
-            await db.commit()
+            await db.execute('COMMIT')
```
Molrn: This is a select query, why do you have to commit ?

Molrn:
replaced by #1831

PR #1828, opened 2026-03-01
parallelized processing in player query

lefebvrele:
Hello to the whole Sharly team, and thank you for your time and involvment in this project.

One of the issue I have with Sharly, is that it's pretty long to manually add new players to my tournaments.

So I'm submitting these changes that should accelerate a bit this process.

Basically, instead of synchronously waiting for each player to be fetched, I moved the logic to an asynchron function, which can be parallelized using asyncio.gather.
I tested these changes locally, and a query that would last ~4.5s before, now takes ~1.7s

### DIFF
--- a/src/web/controllers/admin/player_admin_controller.py
+++ b/src/web/controllers/admin/player_admin_controller.py
@@ -8,6 +8,7 @@ from logging import Logger
 import math
 from pathlib import Path
 from typing import Annotated, Any, Iterable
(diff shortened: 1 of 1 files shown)


Molrn:
All the post-processing is actually not necessary, it is a mistake introduced in the last version. I'm fixing that to remove it, thanks for bringing it to our attention

PR #1831, opened 2026-03-02
fix(session): resolve session management bugs causing HTTP 500 errors

GillesHorn:
## Problem

Multiple session management issues in the web interface cause HTTP 500 errors during normal use, especially under concurrent access. These issues stem from SQLite locking, redundant session writes, session data mutation, and missing database maintenance.

### 1. Retry with exponential backoff (sqlite_store.py)

All write operations (set, delete, delete_all, delete_expired) are wrapped in a retry loop with exponential backoff (50ms to 800ms, max 5 attempts). This handles transient `sqlite3.OperationalError: database is locked` errors that occur under concurrent access with SQLite.

### 2. Fix `renew_for` SQL query (sqlite_store.py)

The `SET` clause was incorrectly placed before `UPDATE`, producing invalid SQL. Corrected the statement to standard `UPDATE ... SET ... WHERE` syntax.

### 3. Fix `delete_expired` condition (sqlite_store.py)

Changed the comparison from `expires < :now` to `expires <= :now` and removed an unnecessary explicit `commit()` call (the `async with` context manager already commits).

### 4. Skip-unchanged session backend (session_backend.py)

New `SkipUnchangedSessionBackend` subclass of `ServerSideSessionBackend`. On load, it computes a SHA-256 hash of the session data. On store, it compares the current hash with the original: if unchanged, the database write is skipped entirely. This eliminates redundant writes on read-only requests (static assets, polling, etc.).

### 5. Fix session data mutation (session.py)

`SubKeySessionVariable.get()` was falling back to `self.default_value` (a mutable dict) and writing it into the session, causing unintended session mutations on every read. Now returns the default without modifying the session.

### 6. Configuration (settings.py)

- `PRAGMA busy_timeout=5000` on the session database connection
- Wired `SkipUnchangedSessionBackend` into `ServerSideSessionConfig`
- Added route exclusions for `/ws`, `.ico`, `.json`  endpoints that do not need sessions
- Set `auto_reload` based on `DEVEL_ENV` flag

## Testing

- Pre-commit hooks: all passed (trailing-whitespace, end-of-file, ruff, ruff-format)
- Unit tests: 86 passed, 4 failed (pre-existing failures confirmed identical on upstream/dev, unrelated to these changes)
- Stress tested with 320 concurrent requests: 0 errors, 0 retries needed, 0 tracebacks in logs

### DIFF
--- a/src/web/session_backend.py
+++ b/src/web/session_backend.py
@@ -13,7 +13,7 @@ read-only requests (screen polling, 304 responses, etc.).
 """
 
 from hashlib import sha256
-from typing import Any
+from typing import Any, cast
 
 from litestar.connection import ASGIConnection
 from litestar.datastructures import MutableScopeHeaders, Cookie
@@ -33,7 +33,9 @@ class SkipUnchangedSessionBackend(ServerSideSessionBackend):
         """Load session data and store a hash of the original state in the connection scope."""
         data = await super().load_from_connection(connection)
         # Store a fingerprint of the loaded data so we can detect changes later.
-        connection.scope[_SESSION_HASH_KEY] = _hash_session(data)
+        # Cast to dict because at runtime ASGI scopes are plain dicts, but
+        # Litestar types them as TypedDict which forbids dynamic keys.
+        cast(dict[str, Any], connection.scope)[_SESSION_HASH_KEY] = _hash_session(data)
         return data
 
     async def store_in_message(
@@ -69,7 +71,7 @@ class SkipUnchangedSessionBackend(ServerSideSessionBackend):
             )
         else:
             # Check whether the session actually changed.
-            original_hash = scope.get(_SESSION_HASH_KEY)
+            original_hash = cast(dict[str, Any], scope).get(_SESSION_HASH_KEY)
             current_hash = _hash_session(scope_session)
 
             if original_hash != current_hash:
@@ -90,6 +92,8 @@ def _hash_session(data: ScopeSession) -> str:
     """Compute a deterministic hash of session data for change detection."""
     if data is Empty or data is None:
         return ''
+    if not isinstance(data, dict):
+        return sha256(repr(data).encode()).hexdigest()
     # Use repr for a quick deterministic serialisation of a dict.
     # Sorting keys ensures deterministic ordering.
     return sha256(repr(sorted(data.items())).encode()).hexdigest()
--- a/src/web/session_backend.py
+++ b/src/web/session_backend.py
@@ -1,0 +1,95 @@
+"""Custom server-side session backend that skips redundant DB writes.
+
+Litestar's default ServerSideSessionBackend writes session data to the store
+on **every** HTTP response, even when the session dict was never modified.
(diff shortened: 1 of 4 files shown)


src/web/sqlite_store.py
```diff
@@ -39,7 +71,8 @@ async def get(
             if renew_for and storage_object.expires_at:
                 storage_object = StorageObject.new(storage_object.data, renew_for)
                 await db.execute(
-                    'UPDATE store(expires_at) WHERE key = ?', parameters=(key,)
+                    'UPDATE store SET expires_at = ? WHERE key = ?',
+                    parameters=(storage_object.expires_at, key),
                 )
                 await db.commit()
```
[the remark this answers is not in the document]
GillesHorn: DELETE (expired key cleanup) and UPDATE (renew_for) write paths in get() are now wrapped in _retry_on_locked() with their own connection scope, consistent with all other write operations. Fixed in the latest force-push.

src/web/settings.py
```diff
@@ -240,6 +244,7 @@ async def create_connection(path: os.PathLike[str]) -> aiosqlite.Connection:
     conn = await aiosqlite.connect(Path(path))
     # Apply high-performance pragmas
     await conn.execute('PRAGMA journal_mode = WAL')
+    await conn.execute('PRAGMA busy_timeout = 5000')
```
Molrn: seems to have no effect (timeout already handled by the connection pool)

src/web/settings.py
```diff
@@ -188,6 +191,7 @@ def __init__(
             loader=template_loader,
             autoescape=True,
             trim_blocks=True,
+            auto_reload=DEVEL_ENV,
```
Molrn: Unrelated, shouldn't be part of the PR. Also I see no reason why it should be different in dev and prod env, it brings more risks of us not identifying a problem

src/web/session_backend.py
Molrn, line 1: mypy errors to fix (see [this job](https://github.com/Sharly-Chess/sharly-chess/actions/runs/22569240253/job/65517376163))

Molrn (dismissed):
@Amaras has a similar draft PR, letting him the approving review as he might have valuable input. Also letting you 2 decide which PR to merge

src/web/settings.py
GillesHorn: Removed. The connection pool already handles acquisition timeout (acquisition_timeout=30), so PRAGMA busy_timeout at the connection level is redundant here.

src/web/settings.py
GillesHorn: Removed. This was an unrelated change and shouldn't be part of this PR. Keeping the same behavior in dev and prod is safer to avoid masking issues.

src/web/session_backend.py
GillesHorn: Fixed all 3 mypy errors:
mypy src now passes cleanly: "Success: no issues found in 343 source files".

Molrn (approved):

PR #1837, opened 2026-03-03
chore: register SQLite date/datetime adapters explicitly

timothyarmes:
Python 3.12 deprecated the default sqlite3 adapters/converters for date
and datetime; registering them explicitly future-proofs all platforms and
silences the DeprecationWarning.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/database/sqlite/adapters.py
+++ b/src/database/sqlite/adapters.py
@@ -1,0 +1,30 @@
+import sqlite3
+from datetime import date, datetime
+
+
+def adapt_date_iso(val: date) -> str:
+    """Adapt datetime.date to ISO 8601 date."""
+    return val.isoformat()
+
+
+def adapt_datetime_iso(val: datetime) -> str:
+    """Adapt datetime.datetime to timezone-naive ISO 8601 date."""
+    return val.isoformat(' ')
+
+
+def convert_date(val: bytes) -> date:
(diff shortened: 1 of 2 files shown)


Molrn (changes requested):
python changelog sources for this [here](https://docs.python.org/3/library/sqlite3.html#default-adapters-and-converters-deprecated) and [here](https://docs.python.org/3/library/sqlite3.html#adapter-and-converter-recipes). 

The way I understand it it does not affect us, as we are using SQLiteDatabase utils functions to pass strings to the execute params and not and not directly passing date / datetime objects to `SQLiteDatabase.execute("", params)`. But it however introduces a difference, because if we insert datetimes using `SQLiteDatabase.execute("INSERT datetime", (datetime.now()))` it won't be in the same format as `SQLiteDatabase.execute("INSERT datetime", (SQLiteDatabase.dump_optional_datetime_to_timestamp_field(datetime.now())))`.  


In general, I suggest we stay very cautious with AI suggested fixes unrelated to PRs which do not fix anything visible, and to only merge them if we actually constat that it fixes something. We'd have much less risk of introducing bugs. Therefore I suggest we close this PR

timothyarmes:
👍

PR #1854, opened 2026-03-07
Add switch on import modal hiding the duplicated rows

Molrn:


pascalaubry (changes requested):
I think that the switch Joueur·euses existant·es on the import modal should not be present when deleting existing players.

PR #1858, opened 2026-03-08
Fixed player check-in on already started tournaments

pascalaubry:
Fixed players with a bye displayed as not checked-in
Set all players not checked-in when closing check-in
Do not overwrite players' FPB and HPB when closing check-in

### DIFF
--- a/src/data/tournament.py
+++ b/src/data/tournament.py
@@ -731,14 +731,6 @@ class Tournament:
             rank_by_player_id[player.id] = previous_rank
         return rank_by_player_id
 
-    @property
-    def not_checked_in_players(self) -> list[TournamentPlayer]:
-        return [
-            tournament_player
-            for tournament_player in self.tournament_players
-            if not tournament_player.check_in
-        ]
-
     @cached_property
     def tournament_players_by_check_in_status(
         self,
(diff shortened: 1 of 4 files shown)


Amaras (dismissed):
Code looks good, pending testing by Pascal

timothyarmes:
I really feel that hit should be added to the tests along with anything related that isn't tested yet

src/data/tournament.py
```diff
@@ -1657,32 +1644,42 @@ def close_check_in(
...
-        if zpbs_last_rounds:
-            zpb_rounds = list(range(self.current_round + 1, self.rounds + 1))
-        elif zpbs_next_round:
-            zpb_rounds = [self.current_round + 1]
 
         with EventDatabase(self.event.uniq_id, write=True) as database:
-            database.set_tournament_check_in(self.id, False)
             if delete:
-                for tournament_player in self.not_checked_in_players:
+                for tournament_player in self.tournament_players_by_check_in_status[
```
Molrn: we now have a defined Enum `CheckInStatus`, using `False | True | None` values is legacy

src/data/tournament.py
```diff
@@ -731,14 +731,6 @@ def ex_aequo_rank_by_player_id(self) -> dict[int, int]:
             rank_by_player_id[player.id] = previous_rank
         return rank_by_player_id
 
-    @property
-    def not_checked_in_players(self) -> list[TournamentPlayer]:
-        return [
-            tournament_player
-            for tournament_player in self.tournament_players
-            if not tournament_player.check_in
-        ]
```
Molrn, line 740: why did you remove that property ? it had a clear usage, and serves the same purpose as `tournament_players_by_check_in_status[False]`, except it does not use legacy values. If you need the usage here to be updated, the update it here, but keep the property as `tournament_players_by_check_in_status` should'nt be used anymore

Molrn (dismissed):

Molrn (approved):

PR #1859, opened 2026-03-08
Fixed board numbering when changing ratings after round #4

pascalaubry:


Amaras (commented):
This is not correct according to the general rules for the handling of Swiss tournaments (FIDE Handbook C.04.2)
The key is score of the higher-ranked participant, then the sum of scores of the paired participants, then pairing number of the higher-ranked participant.

If you really want to do this, then it should be an FFE plugin feature, because it is an inaccuracy on their end

pascalaubry:
No interest in doing something to be compliant with the (wrong) way the FFE displays the boards, closing.

PR #1863, opened 2026-03-12
Add a reminder to update the players' ratings before pairings on round #1 and for tournaments lasting on multiple FIDE periods

pascalaubry:
On round #1:
image

On any round for tournaments lasting on multiple FIDE periods:
image


src/web/templates/admin/pairings/swiss_pairing_buttons.html
```diff
@@ -54,7 +54,17 @@
...
+                        hx-get="{{ url_for(
+                            'admin-pairings-ratings-warning-modal',
+                            event_uniq_id=admin_event.uniq_id,
+                            tournament_id=admin_tournament.id,
+                            round=admin_round
+                        ) }}"
+                        hx-trigger="click"
+                        data-bs-toggle="modal"
+                        data-bs-target="#modal-wrapper"
+                    {% elif admin_round != 1 %}
```
Molrn:
```suggestion
                    {% else %}
```
the else condition below should be removed, it is never accessed

src/web/templates/admin/pairings/ratings_warning_modal.html
Molrn, line 1: - `Yes, ...` with a X and button secondary + `No, ...` button primary feels counter intuitive
- redirection to the player tab is not clear after clicking a button labeled `Update the ratings`, it should redirect to the update modal directly
- text-danger is a bit much for a reminder seen each time, text-warning seems more appropriate
- for multi-period, label is not a question, `Yes / No` answers do not fit

### Suggestion
- modal body
  - (Text-warning): `Reminder: you might have to update the ratings before pairing the first round.`
  - `Update player ratings` dropdown button with the data sources, opening the diff modal (centered)
- buttons `Close` and `Pair`

src/data/tournament.py
```diff
@@ -168,6 +169,12 @@ def start_date_str(self) -> str:
     def stop_date_str(self) -> str:
         return format_date(self.stop_date)
 
+    @property
+    def multiple_fide_periods(self) -> bool:
+        """Returns True if the tournament lasts more than one month, False otherwise."""
+        r: relativedelta = relativedelta(self.stop_date, self.start_date)
+        return r.months + (12 * r.years) > 0
+
```
Molrn: feels overkill to import a new lib just to do so, I suggest simpler:
```suggestion
        return (self.stop_date - self.start_date).days > 30

```
Considering the usage the approximation seems acceptable

pyproject.toml
```diff
@@ -42,6 +42,7 @@ dependencies = [
   "psutil ~= 7.0.0",
   "pycountry ~= 24.6.1",
   "pyexcel-ods3 ~= 0.6.1",
+  "python-dateutil ~= 2.9.0",
```
Molrn:
```suggestion
```
too low of a usage for a new lib IMO

Molrn (dismissed):
Not a hotfix, IMO should go to dev

src/web/templates/admin/pairings/ratings_warning_modal.html
```diff
@@ -0,0 +1,74 @@
...
+                <i class="bi-shuffle pe-1"></i>
+                {{ _('Pair') }}
+            </button>
+        </div>
+    </div>
+</div>
+
+<script>
+    requestRefresh();
+</script>
```
Molrn:
```suggestion
```
RequestRefresh is required to indicate that while in a modal, data might have been modified, and that closing it would display outdated data. It is not the case here. 
The request refresh I was talking about was on the player update endpoint. With your solution, updating players from the pairing tab leads to the players tab, which feels wrong for navigation. Instead, a modification has to be made to the players update endpoint, to request a refresh from here, by returning the following template: 
```
        return HTMXTemplate(
            template_name='common/empty.html',
            re_swap='none',
            trigger_event='request_refresh',
            after='receive',
        )
```

Molrn (dismissed):

Molrn (approved):
LGTM, small issue with the messages not being displayed which I fixed with a redirect

PR #1864, opened 2026-03-13
Remove None type on last_update fields

Molrn:
`last_update` fields can't ever be None, This typing makes it much heavier to use for no reason. We can remove the usage of `{datetime | None} or datetime.min` which is rather heavy

### DIFF
--- a/src/data/family.py
+++ b/src/data/family.py
@@ -250,7 +250,7 @@ class Family:
         )
 
     @property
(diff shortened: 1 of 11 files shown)


pascalaubry (approved):
I just wonder if `src/plugins/utils.py` is the right place for `tournament_results_modified_since()`.

PR #1865, opened 2026-03-14
Sharly-Chess.com Integration

Molrn:
image  

UI used for the Sharly-Chess.com Sync modal. Features are mapped to endpoints with `# TODO`, such that we can work on it at the same time without stepping on each other's toes.


Molrn:
replaced by #1883

PR #1868, opened 2026-03-15
Fixed error message when pairing is impossible

pascalaubry:
Before:
image

After:
image


Molrn (dismissed):
I don't think that's the correct approach. If we can detect when pairing is not possible, the `Pair` button should instead be disabled with a tooltip explaining why

Molrn (approved):
LGTM

PR #1874, opened 2026-03-20
Clarify gender naming

Molrn:
Use Woman / Man + W / M (genders) instead of the current Female / Male + F / M (sexes). It's not the case for every sports, but chess categories are gender-based and not sex-based (ex: [Transgender registration regulations](https://handbook.fide.com/chapter/TransgenderRegulations) in the FIDE handbook), we should accurately mirror that.  
- enum values kept in the DB for compatibility reasons
- `W` and `F` accepted in CSV import


pascalaubry (dismissed):
A few things missed, please see #1875

pascalaubry (approved):

PR #1880, opened 2026-03-21
Fixed players list document when players have no club

pascalaubry:


Molrn (dismissed):
No need to hotfix this IMO, it was the intended behavior so not a fix. Also it should be done the same on all splitters, currently league and school are empty, and category is '-' when it should probably be something more explicit

Molrn (approved):

PR #1882, opened 2026-03-22
Use UTF-8 encoding for CSV exports

Molrn:


pascalaubry (approved):
LGTM (not tested with Vietnamese characters).

PR #1885, opened 2026-03-23
Fix the translation of plural strings

pascalaubry:


Molrn (approved):
LGTM, noticed that a few weeks ago forgot to report it ^^'

PR #1888, opened 2026-03-24
Fix DE tie-break calculation of double round-robin tournaments

### LINKED ISSUE #1656: direct encounter not working correctly with double round robin
## Bug description
direct encounter not working correctly with double round robin.

## How to reproduce
Steps to reproduce the bug:
1. In a double round robin tournament, have the first tiebreak as direct encounter. 
2. Make sure 2 people end the tournament with the same amount of points, and they played 1-1 against each other
3. See the error. Sometimes the player that shouldn't win on the second tiebreak still wins, and it says he wins by direct encounter.

## Behaviour expected
check all the games of the players, not just the first one

## Screenshots

Image

Image

Image

## Environment
- OS: Windows 11
- Sharly Chess version: 3.4.5

pascalaubry:
Closes #1656

### DIFF
--- a/src/data/tie_breaks/tie_breaks.py
+++ b/src/data/tie_breaks/tie_breaks.py
@@ -12,6 +12,7 @@ from common.i18n import _, ngettext
 from data.pairing import Pairing
 from data.pairings import PairingSystem
 from data.pairings.systems import RoundRobinPairingSystem, SwissPairingSystem
+from data.pairings.variations import DoubleBergerRoundRobinVariation
 from data.player import TournamentPlayer
 from data.tie_breaks.categories import (
     TieBreakCategory,
@@ -109,7 +110,6 @@ class TieBreak(OptionHandler[TieBreakOption], ABC):
         tie_break_index: int,
         *,
         after_round: int,
-        double_round_robin: bool,
     ) -> dict[int, int]:
         """Computes the values of all the players in a dict[player_id, value] format."""
         raise NotImplementedError(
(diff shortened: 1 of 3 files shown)


src/data/tie_breaks/tie_breaks.py
```diff
@@ -1720,7 +1737,10 @@ def _compute_player_min_max_points(
...
             if round_ <= after_round and pairing.opponent_id in group_player_ids
         ]
-        not_played = len(player_group) - len(group_pairings) - 1
+        max_played = len(player_group) - 1
+        if double_round_robin:
+            max_played *= 2
+        not_played = max_played - len(group_pairings)
         group_points = sum(
             pairing.result.points(point_values) for pairing in group_pairings
         )
```
Amaras: For multiple encounters, DE uses the average of the scores, not the sum.

> 6.1     If some or all the tied participants have met each other, the sum of the scores from these encounters is used to produce separate standings (for use in 6.2 and 6.3), with the following caveats: (...)
> 6.1.2      contrary to the provisions of Article 4.4, if two participants have met more than once, the addend to be used by them in the aforementioned sum is the average score of these games.

pascalaubry: The result was the same and the calculation less expensive at the very end of a double RR tournament, but not exact before last round.
Fixed now, takes into account all the cases were some players played more than once (even on Swiss tournaments).

Molrn (approved):
fits the handbook's rule, LGTM

PR #1890, opened 2026-03-26
Fix flashes on refresh on some browsers

### LINKED ISSUE #1836: Closing modal flashes white on firefox
On modal which have a `requestRefresh()` in Firefox 148.0, closing the modal triggers a disturbing white flash when reloading

Molrn:
closes #1836. The issue was reproduced on Firefox and Zen. With the fix, I haven't noticed a difference on any other browser. @timothyarmes considering you're the one that added that property would you mind checking it does not break anything on safari pls ?


Molrn:
replaced by #1891

PR #1901, opened 2026-03-31
Fix Windows / Mac Toga icon

Molrn:
Before (export only):
image

After (tested on an export):  
image


timothyarmes (approved):
Works on Mac

PR #1906, opened 2026-04-01
Update keyboard shortcuts tooltip

pascalaubry:
image

### DIFF
--- a/docs/contributing/i18n.md
+++ b/docs/contributing/i18n.md
@@ -12,33 +12,35 @@ Flags:
 
 <!-- DO NOT EDIT! (START BY_LOCALE) -->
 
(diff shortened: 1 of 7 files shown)


Molrn:
page up and page down are not clear, and the menu is now too long.  
Ctrl + arrow is occupied on some browsers.  
I really wish we went with what we first agreed on:  no nav for first / last round, and shift up / down for tournaments. Makes sense to have the nav on one modifier (`shift`) + the key it matches

Molrn (approved):
image

Also made it so the first board is focused when pressing a up / down key with no board selected

timothyarmes (approved):
Nice

PR #1900, opened 2026-03-30
Move `Upload rules` and `Make visible` FFE actions to the upload modal

Molrn:
closes #1293
closes #1294

PR #1908, opened 2026-04-01
Move Upload rules and Make visible FFE actions to the upload modal

Molrn:
replaces #1900


pascalaubry (approved):
LGTM (already tested on #1900)

PR #1909, opened 2026-04-01
Apply white flash fix only to Firefox

Molrn:


pascalaubry (approved):
Not perfect as you said on Discord but much better anyway.

PR #1912, opened 2026-04-02
fix/player-search-Issue

lefebvrele:
I noticed an issue on my code, this is the fix for it

It turns out that event.stopPropagation() doesn't behave the way I thought

It's not a huge bug but it may be annoying:
If the arbiter uses the search player feature and the form on the modal is already filled, pressing "enter" would submit, instead of selecting another player.


pascalaubry:
Replaced by #1914 (hotfix).

PR #1913, opened 2026-04-03
Fix player update

pascalaubry:


Molrn (approved):
Pushed a fix I think is more appropriate (unicity still checked on update and replace but not on the updated / replaced player).

pascalaubry (commented):
Better indeed.

PR #1928, opened 2026-04-05
Clean display of pairing exceptions

pascalaubry:


src/data/pairings/engines.py
```diff
@@ -60,18 +60,23 @@ def generate_pairings(
...
         round_: int,
         partial_pairings: bool = False,
-    ) -> bool:
-        """Generate the pairings of the round *round_* for tournament *tournament*."""
+    ) -> str:
+        """Generate the pairings of the round *round_* for tournament *tournament*, returns an error message or an empty string on success."""
         if self.pairings_generation_disabled_message(tournament, round_):
-            raise ValueError(
-                f'Pairings generation not allowed for round {round_} '
-                f'of tournament [{tournament.name}].'
```
Molrn, line 68: This should not happen, having it raised is important

src/data/pairings/engines.py
```diff
@@ -60,18 +60,23 @@ def generate_pairings(
...
             )
-        stored_boards = self._generate_stored_boards(
-            tournament, round_, partial_pairings
-        )
+        try:
+            stored_boards = self._generate_stored_boards(
+                tournament, round_, partial_pairings
+            )
+        except Exception as e:
+            return str(e)
```
Molrn:
```suggestion
        except Exception as e:
            logger.exception(e)
            return _('An unexpected error occurred, consult the logs for more details.')
```
Kinda opposed to having untranslated error messages for users, especially because they have an access to logs in which we can be more verbous. instead, what's done everywhere else is logging the error then sending a message telling them to consult the logs. 

Also note the usage of logger.exception instead of logger.error, here's the difference:  
```py
        try:
            raise OSError('test')
        except OSError as e:
            logger.error(e)
            logger.exception(e)

## Logs
2026-04-06 05:27:04 ERROR     test  ## logger.error   
2026-04-06 05:27:04 ERROR     test  ## logger.exception
Traceback (most recent call last):
  File "src\data\pairings\engines.py", line 73, in generate_pairings
    raise OSError('test')
OSError: test
```

Molrn (changes requested):

Molrn (approved):

PR #1935, opened 2026-04-08
Fix koya tie break 

Molrn:


RELEASE_NOTES.md
```diff
@@ -27,6 +27,7 @@
 - Display absent players on top of unpaired players on the Pairings tab (4.0.0)
 - Added a reminder to update the players' ratings before pairings on round #1 and for tournaments lasting on multiple FIDE periods (4.0.0)
 - Clean display of pairing exceptions (4.0.2)
+- Fixed koya tie-break (4.0.3)
```
pascalaubry:
```suggestion
- Fixed Koya tie-break calculation (4.0.3)
```

pascalaubry (dismissed):
Fine, just a detail and shouldn't `test_koya` be under `RoundRobinTieBreakTestCase`?

Molrn:
> Fine, just a detail and shouldn't test_koya be under RoundRobinTieBreakTestCase? 

not the way it's built, one class is applied to one file only

PR #1938, opened 2026-04-08
Prevent prize criteria duplication

Molrn:


pascalaubry (changes requested):
Can you also fix the order of the criteria on the cards (any order accepted, but always the same)?

Image

Molrn:
> Can you also fix the order of the criteria on the cards (any order accepted, but always the same)?

not the same PR, it's a very different task

pascalaubry (approved):

PR #1977, opened 2026-04-29
[hotfix] prevent navigation shortcut conflicts

lefebvrele:
huge oversight from my part:
some common browser features (ctrl a, ctrl c, ctrl t...)would also trigger our navigation shortcut

PR #1978, opened 2026-04-30
[hotfix] prevent navigation shortcut conflicts (new PR for #1977)

lefebvrele:


pascalaubry (approved):
Code works fine, found the following bug (I will fix it in another PR).

Hitting '+' in this situation fire the Create button (even when the button is disabled).

Image

PR #1980, opened 2026-04-30
Fix 'bye' usage in French

pascalaubry:


Molrn (commented):
The usage of `Bye` instead of `point joker` was intentional, I semt to understand that `Bye` could be used in french to refer to HPB or FPB (examples [here ](https://fr.wikipedia.org/wiki/Bye)and [here](https://www.echecs.asso.fr/Actu.aspx?Ref=11938)). It also breaks a lot of layouts to go from `Bye` in english to `Forfait ou point joker` in french. Dans le contexte de l'application, j'ai l'impression que cet usage reste clair, notamment puisqu'on a comme traduction de HPB `Demi-Point Joker (bye)`.

I really think we shouldn't merge those and contribute to democratise the usage of Bye in french Chess, and avoid putting `Point joker` everywhere as it seems to be a papi-only usage (I can't find other traces of it online, and the [Livre de l'arbitre](https://dna.ffechecs.fr/wp-content/uploads/sites/2/2025/10/LA-octobre2025.pdf) mentions `Bye` and never `Point joker` (see page 133)).

Molrn:
the usage here seems to match what's been defined in the handbook where Byes are opposed to Forfait and exempt and match FPB / HPB
image

PR #1983, opened 2026-04-30
Fix keyboard shortcuts triggering disabled buttons (see #1978)

pascalaubry:


Molrn (dismissed):
Added 2 other cases, LGTM otherwise ;)

PR #1986, opened 2026-05-01
Also revert the result when permuting the colors of a board

pascalaubry:


src/data/board.py
```diff
@@ -146,12 +146,6 @@ def permute_colors(self):
...
         self.replace_player(white_player, 'black')
-        if self.result != Result.NO_RESULT:
-            white_result = self.result
-            self.black_pairing.stored_pairing.result = white_result.value
-            self.white_pairing.stored_pairing.result = (
-                white_result.opposite_result.value
-            )
         with EventDatabase(self.tournament.event.uniq_id, True) as database:
             database.update_stored_board(self.stored_board)
             if self.result != Result.NO_RESULT:
```
Molrn: this if condition also is to remove

Molrn (dismissed):

PR #1987, opened 2026-05-02
Shorten dates of the current year

Molrn:
image
Shorten all the dates that are of the current year to only use month and day


Molrn:
Ambiguous / not clear

Issue #1976, opened 2026-04-29
Make prizes more comptability-friendly


Molrn:
## Context  
It is currently hard to have a clear idea of how much monetary and non-monetary prizes are distributed for an event, totals only ever show up at category levels, and hybrid prizes can't have a value.   

## Suggestion  
- Add a new field `complementary_value` 
- Split all totals into 2 total values: monetary and non-monetary
- Add totals at 
  - group level (prize tab)
  - Tournament level (groups modal, all prize docs if > 1 categories)
  - Event level (all prize docs if > 1 tournaments)

PR #1989, opened 2026-05-02
feat-1976: Add monetary/non-monetary prize totals and complementary_v…

arun1587:
…alue for hybrid prizes

addresses: https://github.com/Sharly-Chess/sharly-chess/issues/1976


timothyarmes:
replaced with #2198

PR #1994, opened 2026-05-05
Fix zero k-factor

timothyarmes:


Molrn (approved):
Image  

LGTM, surprised at how many k=0 there are in the FIDE DB ^^'

PR #2000, opened 2026-05-06
Add quick creation for the simple prize criteria

Molrn:
image


timothyarmes (dismissed):
I like it, but if you're creating a general category then your criteria are lost without explanation.  I'd suggest Hiding them if general is selected.

PR #2009, opened 2026-05-11
Use category range for age prize criterion

Molrn:
I realized from .sce sent to us that people mostly did multi-select for categories instead of using the lower and greater switches, leading to situations where the criteria is displayed as `Age (U8, U10, U12, U14, U16, U18, U20)` instead of `Age <= U20`.  I replaced it by a category range to match both the tournament criteria and the quick creation form:  

image


pascalaubry (dismissed):
Works fine, please just add a release note before merging.

PR #2015, opened 2026-05-16
Improve internet detection - don't break when behind a captive portal

timothyarmes:


pascalaubry (dismissed):
Works fine, please just add a release note before merging.

PR #2018, opened 2026-05-17
Prevent moving player from SC.com synced to not synced tournament

Molrn:


src/plugins/sce/locale/en/LC_MESSAGES/sce.po
```diff
@@ -184,6 +187,9 @@ msgstr ""
 msgid "Modified"
 msgstr ""
 
+msgid "Moving a player from a tournament synchronised with Sharly-Chess.com to not synchronized tournament is not allowed."
```
pascalaubry:
```suggestion
msgid "Moving a player from a tournament synchronized with Sharly-Chess.com to not synchronized tournament is not allowed."
```

pascalaubry (dismissed):
Maybe limited by the impossibility of distrbuting the players if one of the tournaments is not synced but worls fine as-is.
Just a suggestion on a string.

PR #2019, opened 2026-05-17
Clarify FFE rules upload

Molrn:


pascalaubry (dismissed):
Nice, please just add a release note.

PR #2022, opened 2026-05-20
Remove min games options from norm docs

timothyarmes:


Amaras (approved):
LGTM, as long as the tests pass

PR #2025, opened 2026-05-22
Decorator to reduce unnecessary loading

timothyarmes:


src/data/event_load_spec.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Top-level event tables.
+    load_players: bool = True
+    load_tournaments: bool = True
+    load_timers: bool = True
+    load_screens: bool = True
+    load_families: bool = True
+    load_rotators: bool = True
+    load_display_controllers: bool = True
+    load_accounts: bool = True
```
Molrn, line 33: Accounts always need to be loaded to allow non-admin users access
```suggestion
```

src/data/event_load_spec.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # When set, only this tournament is loaded (the others are skipped
+    # entirely). Useful for endpoints that target one tournament.
+    selected_tournament_id: int | None = None
+
```
Molrn, line 38:
```suggestion
    selected_tournament_id: int | None = None
    selected_screen_id: int | None = None
    selected_family_id: int | None = None
    selected_display_controller_id: int | None = None
    selected_rotator_id: int | None = None
    selected_rotator_screen_index: int | None = None
```
Might be the most important ones as user-screens are the most used requests. If selected, it'd load the event data required for the screen. 
Ex: display controller showing a rotator set on a boards screen  
- load all the screens (required for screen menu)
- load the display controller from its id
- load the rotator associated to it 
- get the screen matching the rotating screen index
- load its screen sets and timers
- load each screen set's tournament
- load each tournament's players and boards

src/data/event_load_spec.py
```diff
@@ -0,0 +1,124 @@
...
+    # Top-level event tables.
+    load_players: bool = True
+    load_tournaments: bool = True
+    load_timers: bool = True
+    load_screens: bool = True
+    load_families: bool = True
+    load_rotators: bool = True
+    load_display_controllers: bool = True
+    load_accounts: bool = True
+
```
Molrn, line 34:
```suggestion
    load_tournament_options: bool = False

```
would allow to only load one tournament on pairings / prize tabs, while preserving the tournament select options on top of the tab

Molrn (changes requested):
Putting on this PR all the comments and ideas I've had in order to make this work. This behavior currently poses one big issue: menu counts.   

Image  

As long as we've got this, every main page needs to have everything loaded, or a count like it's done for metadata on the events page. In addition to this complexity, the player and tournament count depend on the connected person's account, making it much harder to apprehend.  
Not sure how important these counts are, but they're definitely a big setback for this issue.  

It might be safer (and more effective) to start with only adding precise load specs to the user endpoints (screens, rotators, families, display controller, see suggestion below), and keeping the admin as a separate issue for now, as it seems to me it would require a much bigger / risky overall

PR #2026, opened 2026-05-22
Improve SC.com duplicates management

Molrn:
- Prevent duplicates creation on SC.com registration re-creation (Pascal's issue)
- Match identical duplicates automatically
- In the duplicates modal, when deleting one duplicate, mark the other one as not duplicate (see below: deleting one or the other remove the other one from the modal)
image


Molrn:
replaced by #2029

PR #2028, opened 2026-05-23
Allow new FFE licence format AB1234

Molrn:


src/plugins/ffe/ffe_entity.py
```diff
@@ -489,7 +488,7 @@ def get_cell_content(self, player: Player) -> Any:
     def _augment_stored_player(self, stored_player: StoredPlayer, value: str):
         if not value:
             return
-        if not re.match(r'^[A-Z]\d{5}', value):
+        if not PlayerFFELicence.validate(value):
             raise SharlyChessException(
                 _('Invalid format (expected: {format}).').format(format='A12345')
             )
```
pascalaubry, line 496:
```suggestion
raise SharlyChessException(
_('Invalid format (expected: {format1} or {format2}).').format(format1='A12345', format2='AB1234')
)
```

pascalaubry (changes requested):
IMO There are other places where the format is used (`FFESqlServer.string_matches_ffe_licence_number` for sure, maybe the same for the local data source).

PR #2036, opened 2026-05-27
Fix player withdrawn behavior

Molrn:
currently the withdrawn status only applies from the current round, which makes it appear wrong on previous rounds in the pairings tab when the player was not paired. Also prevent returning / removing a player on the last round as it does nothing


pascalaubry (approved):
Perfect.

PR #2046, opened 2026-06-07
Prevent logging of all files when looking for releases

Molrn:
Also fixes a problem with operations on Path with names >= 256 chars on windows


pascalaubry (approved):
Can you merge to `hotfix/4.2.1` please?

PR #2054, opened 2026-06-12
Reorganize Toga application

Molrn:
image

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,43 +2,4 @@
 
 ## General
 
-- Improved detection of internet connectivity (4.2.0)
-- Prevent reloading templates (4.2.1)
-
-## Tournaments
-
-- TRF26 import / export (4.2.0)
-
-## Players
-
-- Clarify actions on absent players if over half absent players (4.2.1)
-
-## Documents
-
-- Added a new document to get a summary of norms achieved, or a forecast of norms that can be achieved for the final round (4.2.0)
-- Added options to the norm document for the number of rounds and the event type (4.2.0)
-- Added detailed calculations for the norm report to allow for easy validation (4.2.0)
-- Added automatic removal of specific played games in order to try to fulfil the norm requirements (4.2.0)
-- Fixed a norm calculation bug that incorrectly considered CM and WCM as title-holders (4.2.0)
-- Added support for norm exemption 1.4.2c (last round forfeit as loss) (4.2.0)
-- Added custom team ranking documents (4.2.0)
-- For norm calculations, 1.4.3 exemption now also waive 1.4.4 as confirmed to us by the FIDE (4.2.1)
-- For norm calculations, players under the FID federation are now disregarded when considering foreign federation counts (4.2.1)
-
-## Screens
-
-- Restored check-in on input screens (4.2.0)
-- Fixed default global background color (4.2.1)
-- Fixed display controller redirection on results or check-in input (4.2.1)
-
-## _Sharly-Chess.com_
-
-- Added missing timeouts (4.2.0)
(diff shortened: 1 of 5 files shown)


pascalaubry (commented):
Code is fine for me but we need to discuss the "Support us" button.

pascalaubry:
I think that we could simplify:
- the Welcome panel, by replacing `Page d'accueil <url_button>` by a simple button with text `Ouvrir le navigateur`.
 - the Logs panel, by having on a single line: Niveau de log : <level_select> Affichage : ☐ Couleurs ☐ Niveau ☐ Date et heure <clear_logs_button>

Molrn:
> Browser button   

Currently we have "Ouvrir l'interface d'administration" which means nothing to the user. The change to "Home page: {link}" intends to have something explicit: there is a link (users know it will open a browser) and a label expliciting what that link means. "Ouvrir le navigateur" is IMO not explicit enough, like any link opens the browser, it's important to indicate what it opens. I don't feels this would simplify it, it would in opposite obscure it. The home page link does not strike me as something too techy which would bring problems to users, trying to hide it makes it feel n$more complex than it is

> Logs panel   

Params were intentionally hidden behind a button, it overloads the interface for no reason and it's not explicit that it is used as settings for the logs. 
If you're not satisfied with this implementation, I had another in mind, where the "Log settings" button would open a window with all the parameters. Do you prefer me to do it this way ?

> we need to discuss the "Support us" button.   

Missed your comment, definitely, available to discuss it anytime

pascalaubry:
> > Browser button
> 
> Currently we have "Ouvrir l'interface d'administration" which means nothing to the user. The change to "Home page: {link}" intends to have something explicit: there is a link (users know it will open a browser) and a label expliciting what that link means. "Ouvrir le navigateur" is IMO not explicit enough, like any link opens the browser, it's important to indicate what it opens. I don't feels this would simplify it, it would in opposite obscure it. The home page link does not strike me as something too techy which would bring problems to users, trying to hide it makes it feel n$more complex than it is

Maybe anothe text, but URLs are too techy (and users can already find all the URLs on the Network panel).

> > Logs panel
> 
> Params were intentionally hidden behind a button, it overloads the interface for no reason and it's not explicit that it is used as settings for the logs. If you're not satisfied with this implementation, I had another in mind, where the "Log settings" button would open a window with all the parameters. Do you prefer me to do it this way ?

No need to add another window, just remove the unnecessary `Log parameters` button and keep everything on a single line.

Molrn:
"just" do the thing I explicitely said I thought was a bad idea and that I modified from the previous behavior ? kinda needs to be argued. It's normal to have a button for the settings, having them always displayed is weird and overloading when you first open the page, makes the page seem much more complex than it really is. Please reconsider or at least argue your position. 

> Maybe anothe text, but URLs are too techy (and users can already find all the URLs on the Network panel).  

They can't find the URL of the Home page on the networks panel. It's the only URL standard users (i.e. not using screens / networks) might need, and they'll be exposed to it considering it's a browser app. Hiding it behind something it is not might make people unsure of how they could access it again if they close their browser tab. Imma keep thinking of labels, if I find something clearer I'll put it here

Molrn:
@pascalaubry It gave me the idea to add a button "Access log files", it will definitely come in handy when asking for people's logs. Gives another reason to split params and buttons  
image

Molrn:
@pascalaubry removed the support button and changed the browser button label. For the logs it's really important to have the settings separated I think we should keep it, or if you are really disturbed by them this way there is always the solution I suggested of putting them in a dialog

pascalaubry (approved):
Please consider #2081 and add a release note before merging.

PR #2055, opened 2026-06-13
Fix norms 1.4.3b.  Don't apply 1.4.4 in this case

timothyarmes:


Amaras:
Related but not directly.
There are a few quirks coming from how the searcher is structured: it is possible in 11+ rounds tournaments to have it become confused and try to remove round 10/11, even if it was a loss, since there is no concept of dropping the whole tail.

Below is the false positive it gave me during the Top 16, where it dropped the loss at round 10 using 1.4.1e.
[faux_positif_141e.pdf](https://github.com/user-attachments/files/28925233/faux_positif_141e.pdf)

Proposed fix: separate the 1.4.1e and 1.4.1f subsets searcher and changing the candidates generation, making sure that if you drop round 10, you also drop rounds 11 and beyond.

src/data/norms/forecaster.py
```diff
@@ -76,6 +76,30 @@ def can_forecast_round(self, round_: int) -> bool:
             return False
         return pairing.result == Result.NO_RESULT
 
+    def round_result_decided(self, round_: int) -> bool:
+        """True iff the player has a real game in `round_` whose result is
+        already entered — the forecast for her is settled even though the
+        round may still be open for other players."""
```
pascalaubry, line 82:
```suggestion
        """True if the player has a real game in `round_` whose result is
        already entered — the forecast for her is settled even though the
        round may still be open for other players."""
```

pascalaubry (commented):
Needs @Amaras's approval IMO.

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -7002,8 +7008,8 @@ msgstr "Une semaine"
...
+msgid "What the player needs in round {round} to achieve each norm. \"any OTB result\" means a played win, draw or loss all work — but the round must still be played (a no-show becomes a forfeit-loss and disqualifies the norm). \"norm already achieved\" means it is locked in regardless of round {round} — either the result is already in, or it holds without this round (1.4.1e)."
+msgstr "Ce dont le joueur·euse a besoin à la ronde {round} pour réaliser chaque norme. « any OTB result » signifie qu'une victoire, une nulle ou une défaite jouée conviennent toutes — mais la ronde doit être jouée (un forfait devient une défaite par forfait et disqualifie la norme). « norm already achieved » signifie qu'elle est acquise quel que soit le résultat de la ronde {round} — soit le résultat est déjà saisi, soit elle tient sans cette ronde (1.4.1e)."
```
Amaras: The strings "any OTB result" and "norm already achieved" are translated elsewhere in the document, so they should be translated here as well

src/common/profiling.py
Amaras, line 1: It shouldn't be in a PR dedicated to something else

src/data/event.py
```diff
@@ -502,12 +502,42 @@ def get_player_duplicate(
...
+            for ref_player in self.players:
+                for player in self.players:
+                    if player.id == ref_player.id:
+                        continue
+                    if any(
+                        duplicate_hook(
+                            stored_player=ref_player.stored_player, player=player
+                        )
+                    ):
+                        return True
```
Amaras: That looks like something that is out of scope of this PR

src/data/loader.py
Amaras, line 1: probably should be pushed outside this PR

src/database/sqlite/sqlite_database.py
Amaras, line 1: outside the scope of this PR

src/web/controllers/admin/index_admin_controller.py
Amaras, line 1: outside the scope of this PR

src/web/controllers/admin/pairings_admin_controller.py
Amaras, line 1: outside the scope of the norms PR

src/web/profiling_middleware.py
Amaras, line 1: way outside the scope of the norms PR

src/web/settings.py
Amaras, line 1: outside the scope of the norms PR

Amaras (changes requested):
One translation nitpick, and the rest are not about norms.
Please make a separate PR to get them into dev, I don't want *yet another* PR piggybacking on an already complex one.

src/data/norms/forecaster.py
timothyarmes: iff is correct.

Amaras (commented):
Alright for me, need to solve the mypy issues before I can merge

PR #2058, opened 2026-06-15
Fix display controllers redirected to screens on user input

Molrn:
To reproduce: 
- Create screen A (input) and screen B (players)
- create a display controller DC
- associate A to DC
- Open DC in tab T --> Displays A
-  input a result from DC
- Associate B to DC

--> T displays A instead of B


pascalaubry (approved):
Vicious bug, well done.

PR #2074, opened 2026-06-22
Translate team tie-breaks' acronyms

pascalaubry:


Molrn (commented):
All the tie-break acronyms are untranslated. Is there a basis for those translations, like they exist outside of the SC context ? I really think we should stick to FIDE-formatted acronyms, it's odd to have a mix of full caps + no caps. 

In addition, if the acronyms are modified it's important to keep an understandable naming in the TRF file (ex: papi Buchholz --> OTHER_PAPI_BUCHHOLZ_STANDARD) and ensure the TRF can be exported and imported without losing a tie-break

timothyarmes:
Will do this differently.

PR #2082, opened 2026-06-26
Split data and program files

Molrn:
To test the data change: run the change in a first terminal, then restart the app in a second terminal to apply the change. It relies on env variables, so staying in the same terminal they do not get refreshed. If running in a IDE (ex: Pycharm), you need to restart the IDE

### DIFF
--- a/src/common/__init__.py
+++ b/src/common/__init__.py
@@ -5,11 +5,19 @@ import shutil
 import sys
 from collections import namedtuple
 from pathlib import Path
+from typing import Any
 from urllib.parse import urlparse
 from packaging.version import Version
 
 from common.exception import SharlyChessException
 from utils.program_variables import MACOS_SUPPORT_DIR, ProgramVar
+
+
+if sys.platform == 'win32':
+    import winreg
+else:
+    # Avoid winreg mypy errors when not running on windows
+    winreg: Any = {}
 
 APP_NAME: str = 'sharly-chess'
 SHARLY_CHESS_VERSION: Version = Version(importlib.metadata.version(APP_NAME))
@@ -79,8 +87,6 @@ def _app_base_dir() -> Path:
 def _default_prod_data_dir() -> Path:
     match sys.platform:
         case 'win32':
-            import winreg
-
             doc_reg_key = (
                 r'Software\Microsoft\Windows\CurrentVersion\Explorer\User Shell Folders'
             )
--- a/src/common/__init__.py
+++ b/src/common/__init__.py
@@ -1,6 +1,7 @@
 import importlib.metadata
 import os
 import re
+import shutil
 import sys
 from collections import namedtuple
 from pathlib import Path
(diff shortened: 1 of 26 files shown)


src/utils/program_variables.py
```diff
@@ -0,0 +1,108 @@
...
+else:
+    # Avoid winreg mypy errors when not running on windows
+    winreg: Any = {}
+
+# OS-specific Paths
+MACOS_SUPPORT_DIR = (
+    Path.home() / 'Library' / 'Application Support' / 'com.sharlychess.thp'
+)
+MACOS_DATA_PLIST = MACOS_SUPPORT_DIR.parent / 'com.sharlychess.plist'
+MACOS_DATA_PLIST.parent.mkdir(parents=True, exist_ok=True)
```
timothyarmes: This isn't conditional, it'll create the folder on all platforms...

src/common/__init__.py
```diff
@@ -82,38 +84,111 @@ def app_base_dir() -> Path:
...
+if previous_dir_val:
+    # The data dir changed: move the previous content over if the new dir is empty.
+    previous_dir = Path(previous_dir_val)
+    if previous_dir.exists() and not any(DATA_DIR.iterdir()):
+        is_child = previous_dir in DATA_DIR.parents
+        for elem_path in previous_dir.glob('*'):
+            if is_child and (elem_path == DATA_DIR or elem_path in DATA_DIR.parents):
+                continue
+            shutil.move(elem_path, DATA_DIR)
+        if not is_child:
```
timothyarmes: glob('*') skips hidden files, so a macOS .DS_Store survives the move, and rmdir() raises OSError: Directory not empty. Hidden files also never migrated. Fix: iterate incl. hidden or shutil.rmtree...

src/gui/server_gui_toga.py
```diff
@@ -1032,17 +1131,35 @@ async def _message_on_ui():
         except Exception:
             return False
 
-    def quit_app(self) -> None:
+    def quit_app(self, restart: bool = False) -> None:
```
timothyarmes: Can't test - where is this called with restart = True?

src/utils/enum.py
```diff
@@ -12,6 +12,15 @@
     from data.tournament import Tournament
 
 
+class Extension(StrEnum):
+    EVENT_DB = 'sce'
+    LEGACY_EVENT_DB = 'db'
```
timothyarmes, line 17: Why both SOURCE_DB and LEGACY_EVENT_DB?

src/plugins/ffe/papi_converter.py
```diff
@@ -223,7 +223,7 @@ def convert_player_database(self, source_file: Path, target_file: Path) -> bool:
                 raise SharlyChessException(
                     'Player database conversion error: SQLite database was not created.'
                 )
-            return True
+        return True
```
timothyarmes: Oops!

timothyarmes (changes requested):

src/gui/server_gui_toga.py
Molrn: It was an idea I tried that I forgot to remove the code of, with the idea of instant restarting instead of waiting for the user to do so. Problem with it was that restarting this way changed the port (port don't instant free themselves when the process using them is stopped) so I abandoned the idea. Removing the restart code

src/utils/enum.py
Molrn: we might wanna keep change the source DB while the legacy must stay as .db. all the extensions have their usage name and not directly their ext, I think the usage is cleaner this way. Also Enums can have multiple times the same value without any problem, it's pretty common for these kinds of enums that in usage are never called with their constructor like Extension('db')

PR #2090, opened 2026-06-29
Handle updates and version recovery from Toga

Molrn:
Missing:
- take linux into account
- installer exe
- adapt export


Molrn:
replaced by  #2104

PR #2104, opened 2026-07-06
Improve Setup / Update pipeline

Molrn:
### Missing  
- Windows signing via InnoSetup  

### Testing (Windows)
- Install [Inno Setup 6](https://jrsoftware.org/isdl.php/Inno-Setup-Downloads)
- Run `python scripts/export/build.py`  
- run `python scripts/export/windows/generate_setup_files.py`
- Add to the .env:
```
SHARLY_CHESS_FAKE_LATEST_VERSION=5.0.0
DEV_WINDOWS_UPDATER_PATH="export/Sharly Chess Updater 5.0.0.dev1.exe"
```

You can the installer by executing `export/Sharly Chess Setup 5.0.0.dev1.exe`, or run the update from the Toga app


scripts/export/windows/generate_setup_files.py
```diff
@@ -0,0 +1,89 @@
...
+from pathlib import Path
+
+from common import SHARLY_CHESS_VERSION, BASE_DIR
+from common.logger import get_logger
+
+logger = get_logger()
+
+IS_VERSION = 6
+ISCC_EXE = Path(rf'C:\Program Files (x86)\Inno Setup {IS_VERSION}\ISCC.exe')
+ISS_SCRIPT_FILE = BASE_DIR / 'windows-setup.iss'
```
pascalaubry, line 16: Suggestion: move this file to limit the number of files at the root of the project (is it possible?).
```suggestion
EXPORT_DIR = BASE_DIR / 'export'
ISS_SCRIPT_FILE = EXPORT_DIR / 'windows-setup.iss'
```

scripts/export/windows/generate_setup_files.py
```diff
@@ -0,0 +1,89 @@
...
+UPDATER_EXE = BASE_DIR / 'export' / f'Sharly Chess Updater {SHARLY_CHESS_VERSION}.exe'
+SETUP_EXE = BASE_DIR / 'export' / f'Sharly Chess Setup {SHARLY_CHESS_VERSION}.exe'
```
pascalaubry: Suggestion: limit the usage of space in filenames.
```suggestion
UPDATER_EXE = EXPORT_DIR / f'sharly-chess-updater-{SHARLY_CHESS_VERSION}.exe'
SETUP_EXE = EXPORT_DIR / f'sharly-chess-setup-{SHARLY_CHESS_VERSION}.exe'
```

src/common/data_recovery.py
```diff
@@ -0,0 +1,290 @@
...
+                dst = DATA_DIR / file.relative_to(DEFAULT_DATA_DIR)
+                dst.parent.mkdir(exist_ok=True)
+                shutil.copy(file, dst)
+
+        if stored_version != SHARLY_CHESS_VERSION:
+            ProgramVar.VERSION.write_value(str(SHARLY_CHESS_VERSION))
+
+        if DEVEL_ENV and IS_NEW_INSTALL and not recovered:
+            if (Path() / 'events' / '.scc').exists():
+                cls._recover_legacy_version(Version('4dev1'), Path())
```
pascalaubry: Why `4dev1`?

pascalaubry (dismissed):
Nice, a few suggestions inline and some others on branch `feature/version-update-strings`.

Error on `generate_setup_files`:
```
Error on line 64 in C:\OneDrive\echecs\sharly-chess\dev\windows-setup.iss: No files found matching "C:\OneDrive\echecs\sharly-chess\dev\dist\sharly-chess-5.0.0.dev1\_internal\tools\*.exe"
Compile aborted.
Inno Setup Compiler failed.
```

src/common/data_recovery.py
Molrn: it's version 4, it's in dev, we have no way to know which version it is and it doesn't have to be accurate as is it only used for looging, only the path is important

scripts/export/windows/generate_setup_files.py
Molrn: Unfortunately it isn't possible, the syntax does not allow path backtracking (ex: `.\..\..\`) so to access different dirs at the root it has to be at the root

scripts/export/windows/generate_setup_files.py
Molrn: A bit on the fence on this. when looking at the usages the ones with spaces in them seem to be the ones who are trying to be appealing to none tech users (ex: thunderbird, pdf editor), so renaming that way is a bit in order to appear less techy. I get why spaces can be annoying in general, but I think at least snake case would be nice. I suggest `SharlyChess-{Setup|Updater}-{version}.exe`, you ok with that ?

Usages: 
image

pascalaubry: OK let's keep the spaces.

pascalaubry (approved):

PR #1600, opened 2025-12-20
Detect antivirus software

pascalaubry:
PR #1599 must be revied before this one.
Note: the tests pass locally.

PR #2108, opened 2026-07-09
Add program and data dir to Microsoft Defender exceptions list

Molrn:


windows-setup.iss
```diff
@@ -180,3 +245,12 @@ fr.UpdateSuccessfull=Sharly Chess a été mis à jour avec succès. L'applicatio
...
 fr.ShouldDeleteData=Voulez-vous également supprimer les données de l'application ?
+
+en.DefenderGroup=Microsoft Defender exception list:
+fr.DefenderGroup=Liste d'exclusions Microsoft Defender :
+
+en.DefenderProgram=Add the program folder
+fr.DefenderProgram=Ajouter le dossier du programme
+
+en.DefenderData=Add the data folder
+fr.DefenderData=Ajouter le dossier des données
```
pascalaubry, line 259:
```suggestion
en.DefenderGroup=Add the following folders to Microsoft Defender exceptions:
fr.DefenderGroup=Ajouter les dossiers suivants aux exclusions Microsoft Defender :

en.DefenderProgram=The program folder
fr.DefenderProgram=Le dossier du programme

en.DefenderData=The data folder
fr.DefenderData=Le dossier des données
```

pascalaubry (changes requested):
Suggestion on strings and a remark: we should test if Defender is running before proposing to add exceptions (see `src/antivirus/programs/windows/windows_defender.py` on #1600).

Molrn:
> we should test if Defender is running  

There is a check with IsDefenderActive. However checking if the folder has already been added takes too long, and it does not bring any value as adding it a second time is not impactful. I'd rather not add that test and stay with a simple "if active then offer the option"

windows-setup.iss
Molrn: seems a bit weird in the usage, all the tasks have a action verb before them and not into their group names... I think we should keep the current version

pascalaubry (approved):
OK with everything.

PR #2109, opened 2026-07-09
Remove the version from Toga

pascalaubry:
image

image


Molrn:
@pascalaubry I think the version should be removed from the home page entirely, I left it because this would be the spot of a support button and the window felt way too small and a bit off without a second line on the bottom section. But in general, the main window's title is the app's name, and I think we established before that we didn't want the version in the app name (cf installer feedback), as this goes towards making the app's versions less distinct

PR #2115, opened 2026-07-13
Fix the data path input on Mac

timothyarmes:


Molrn (approved):
displays well on windows and linux without a shorten path, LGTM

PR #2118, opened 2026-07-17
Fix backward migrations

Molrn:


Amaras (approved):
That's what I did manually, so approving

PR #2122, opened 2026-07-19
Speed up event loading & result entry

timothyarmes:
## Summary

  Large events were painfully slow — ~1 s per result entry on a 581-player
  event. The root cause was a missing index, compounded by an N+1 load and
  repeated O(N²) board-number computation. This PR fixes the hot paths and
  adds structural guards so the board layout's derived state can't silently
  go stale.

  Measured on `france-rapide-2023` (581 players, round 9):

  | path | before | after |
  |------|-------:|------:|
  | Event load from DB | 773 ms | 27 ms |
  | ` └ boards-by-round query` | 745 ms | 7 ms |
  | `set_for_round` | 114 ms | 20 ms |
  | **Result entry (PUT)** | **~1000 ms** | **~45 ms** |

  ## Performance

  - **Index the pairing/board hot paths** (migration `m088`): the
    `board`/`pairing` join in `load_tournament_stored_boards_by_round` ran
    against an unindexed `pairing` table — O(boards × pairings). Adds indexes
    on `pairing(board_id)`, `pairing(tournament_id, player_id, round)`,
    `board(team_board_id)`, `team_board(tournament_id)`. The composite index
    also speeds the result-entry `UPDATE`. This is the dominant win.
  - **Kill the pairing N+1**: tournament players loaded their pairings one
    query per player (581 queries); now one batched query grouped by player.
  - **Compute board numbers once per round** instead of rebuilding the whole
    map (+ a plugin hook) on every `board.number` access — the O(N²) in
    `set_for_round` and full-page renders. Memoised per `(round, layout
    version)`.
  - **Skip a wasted compute on result entry**: the hotkey handler built a
    full pairings context (a whole `set_for_round`) only to read a
    team-only flag; individual events now skip it.
  - **Per-pass memoisation** of opponent-score scans (Buchholz & other
    tie-breaks) during ranking, gated so no caller outside a compute pass can
    read a stale value.

  ## Robustness (can't-go-stale-by-accident)

  - `boards_by_id` is now a `BoardsById` dict that bumps a `version` on every
    mutation; the board-number cache keys off it, so add/remove invalidates
    automatically — no bump to remember at call sites.
  - `Board` gains `index` / `white_player_id` / `black_player_id` setters that
    keep the raw field, the cached weakref, and the number cache in sync.
  - `StoredBoard` is now `frozen=True`: accidental raw writes are a **compile
    error** (even via an alias). The handful of legitimate mutations go through
    one explicit `set_stored_fields(obj, **fields)` helper (in-place, so records
    shared between the wrapper graph and the stored lists stay consistent).


src/database/sqlite/event/event_database.py
```diff
@@ -907,44 +907,66 @@ def delete_all_tournament_stored_tie_breaks(self, tournament_id: int):
...
+            'SELECT `id`, `last_name`, `ratings`, `first_name`, '
+            '`date_of_birth`, `year_of_birth`, `gender`, `mail`, `phone`, '
+            '`comment`, `owed`, `paid`, `title`, `fide_id`, `federation`, '
+            '`club`, `fixed`, `check_in`, `team_id`, `team_index`, '
+            '`plugin_data` FROM `player`'
+        )
+        assert self.cursor is not None
+        return [
+            StoredPlayer(
+                player_id,
```
Molrn: It's hard to maintain and very error prone to rely this much on class order. I get the fix, but at least the order should be contained in this function, and named arguments used for the stored object like `StoredPlayer(id=player_id)`

Molrn (changes requested):

src/database/sqlite/event/event_store.py
```diff
@@ -96,7 +109,7 @@ class StoredPairing:
     effective_points: float | None = None
 
 
-@dataclass
+@dataclass(frozen=True)
```
Molrn, line 105: When looking it up it looks like frozen dataclasses are slower to instantiate. Why is that here ?

src/database/sqlite/event/event_store.py
```diff
@@ -11,6 +11,19 @@
 from utils.enum import EventType
 
 
+def set_stored_fields(obj: Any, **fields: Any) -> None:
```
Molrn, line 14: If this is for bypassing the frozen state, why have it frozen in the first place ? this is slower than `stored_object.attribute = value`

src/data/event.py
```diff
@@ -541,14 +543,72 @@ def get_player_duplicate(
...
+            for index, ref_player in enumerate(players):
+                for player in players[index + 1 :]:
+                    if any(
+                        fallback_hook(
+                            stored_player=ref_player.stored_player, player=player
+                        )
+                    ) or any(
+                        fallback_hook(
+                            stored_player=player.stored_player, player=ref_player
+                        )
```
Molrn: - what is a pair-based plugin ?
- there currently only is one plugin implementing this, FFE. How is it possible this requires double the code, duplication for types of plugin hooks I do not see anywhere in the doc, and is now atrocious to read ?

src/data/tournament.py
```diff
@@ -314,15 +358,23 @@ def last_pairing_update(self) -> datetime:
...
+    @cached_property
     def pairing_variation(self) -> 'PairingVariation':
+        """The configured variation, resolved once for this event snapshot.
+
+        A :class:`Tournament` belongs to a single request's freshly loaded event.
+        Resolving the variation walks the event-aware plugin manager, and hot paths
+        such as point computation and pairing-row rendering consult it once per
+        player.  The stored variation cannot change during the lifetime of this
+        snapshot, so repeating that lookup only adds work.
+        """
```
Molrn: I'm not saying much in general but these kinds of AI comments are pissing me off, we don't need 10 lines about why this is a cached property. It only makes sense at review time, after that it remains in the code base and make it harder to navigate. Please tune them down in your claude settings if possible and only keep them if relevant

Molrn (changes requested):

src/database/sqlite/event/event_store.py
timothyarmes: It's specifically there since the optimisation requires that we MUST use the getter/setters function.  It would be easy later on to forget that, so by freezing it we can't accidentally write to the raw data class like we so in so many other places in the code.  It's defensive, any speed difference here is even measurable.

src/database/sqlite/event/event_store.py
timothyarmes: See above.  There are places that we nevertheless need to update the raw data.  This makes it explicit and forces recognition that we know what we're doing.

src/data/tournament.py
timothyarmes: I added that because I didn't think it was obvious.

PR #2132, opened 2026-07-25
More useful norms details before tournament end

timothyarmes:


Amaras (commented):
Code looks correct, and easier than what my naive implementation would have been

Amaras (approved):
And the approval I forgot

PR #2136, opened 2026-07-27
Add more and better Molter recipes

timothyarmes:


pascalaubry (commented):
I must say I have to trust you on the Molter part.

Just wondering if `/docs/technical-appendices/molter/build*.py` should not be moved to `/scripts/molter/`.

PR #2138, opened 2026-07-28
Update Windows dev updater path

pascalaubry:


timothyarmes (commented):
This change is already on the release fix branch

PR #2145, opened 2026-07-30
Display the appropriate metric only for each prize category

timothyarmes:


pascalaubry (approved):
I will fix the FromQuery/FromPath issue in another PR.

PR #2157, opened 2026-08-09
Prevent from adding and importing players with invalid ratings

pascalaubry:


timothyarmes:
I think there's a merge issu here.  74 changed files

timothyarmes (changes requested):
I think the min value should be 0

pascalaubry:
> I think the min value should be 0

As discussed, the field can be left empty.

PR #2160, opened 2026-08-09
Improve team table numbering

pascalaubry:
Sort the team pairs by highest primary score, sum of the PSs, highest secondary score, sum of the SSs, TPN.

### DIFF
--- a/src/data/pairings/engines.py
+++ b/src/data/pairings/engines.py
@@ -736,7 +736,7 @@ class _TeamPairingBase(PairingEngine, ABC):
                 - 0. 0 for PAB, 1 otherwise
                 - 1. the higher primary score
                 - 2. the lower primary score
-                - 3. the highest pairing number
+                - 3. the lowest pairing number of the high ranked team
                 """
                 a_id, b_id = pair
                 if b_id is None:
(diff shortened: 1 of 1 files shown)


src/data/pairings/engines.py
```diff
@@ -719,26 +719,30 @@ def _next_free_index() -> int:
...
                 pair: tuple[int, int | None],
-            ) -> tuple[int, tuple[float, float], tuple[float, float], float, float]:
+            ) -> tuple[int, float, float, float, float, float]:
+                """Returns the sort key to use to sort the pair for table numbering, as a tuple.
+                - 0. 0 for PAB, 1 otherwise
+                - 1. the higher primary score
+                - 2. the lower primary score
+                - 3. the higher secondary score, if defined
+                - 4. the lower secondary score
+                - 5. the highest pairing number
```
timothyarmes:
```suggestion
                - 5. the lowest pairing number
```

timothyarmes (dismissed):

src/data/pairings/engines.py
pascalaubry: I really meant "lowest" because we use `-tpn` to get the lowest TPN first.

PR #2167, opened 2026-08-10
TRF fixes and improvements

timothyarmes:


src/web/controllers/admin/pairings_admin_controller.py
```diff
@@ -62,6 +62,14 @@
 
 logger = get_logger()
 
+# Bonus / penalty points are exported in the TRF26 299 record, whose
+# point fields are ``[-]11.5``: four columns holding either a sign or a
+# second integer digit, never both. Rounded inwards to a multiple of the
+# form's 0.5 step, which is anchored on the minimum — an unaligned bound
+# would make ordinary values like -1 unenterable.
+MIN_POINT_ADJUSTMENT = -9.5
+MAX_POINT_ADJUSTMENT = 99.5
```
pascalaubry, line 71: Can we cap with the same positive value and still be TRF-compliant (I understand that we must fit to the 4-char limit but it is strang to allow 10 bonus and not 10 penalty)?

pascalaubry (approved):

PR #2175, opened 2026-08-11
N2F tie break + pairing sheets for Parité and N2F

timothyarmes:


pascalaubry (changes requested):
Everything works fine but I do not get the N2F logo (on the pairing sheet).

pascalaubry (approved):

PR #2181, opened 2026-08-12
Treat PTS as a tiebreak

timothyarmes:


pascalaubry (commented):
Now PTS is not necessarily the first tie-break, I wonder if we should not warn that using some tie-breaks as the first tie-break makes no sense.

pascalaubry (approved):

PR #2187, opened 2026-08-14
Fix the team ranking document

timothyarmes:


pascalaubry (approved):
Works fine, thanks for theBerger grid I forgot.

PR #2189, opened 2026-08-16
Fix the tie-breaks modal

pascalaubry:
Make the "Quick-apply" section always visible and move it up before the custom sets.

image


timothyarmes (approved):
Nicer

PR #2195, opened 2026-08-17
Update docs + tests

timothyarmes:


pascalaubry (approved):
On my laptop, tests passed in 893 seconds, now they do in 403.

PR #2198, opened 2026-08-19
Complementary value for Hybrid prizes

timothyarmes:


locale/en/LC_MESSAGES/messages.po
```diff
@@ -9977,3 +9979,6 @@ msgstr ""
 msgid "← Back to IT1 certificate"
 msgstr ""
 
+#~ msgid "value: {value}"
+#~ msgstr ""
+
```
```suggestion
```

locale/fr/LC_MESSAGES/messages.po
```diff
@@ -9979,3 +9981,6 @@ msgstr "—"
 msgid "← Back to IT1 certificate"
 msgstr "← Retour au certificat IT1"
 
+#~ msgid "value: {value}"
+#~ msgstr "valeur : {value}"
+
```
```suggestion
```

pascalaubry (changes requested):
Just two remarks on these prizes:

Image

Here I would display the non-monetary 50€.

Image

And here I would remove the parenthesis.

Image

PR #2202, opened 2026-08-21
Add Docker support for dev and production use

GuyGuy-59:
Adds a Dockerfile/docker-compose setup to run Sharly Chess headless in a container, plus the fixes needed to make that a good experience rather than a workaround: the app already falls back to console mode without GTK, but running it in a container surfaced a few real bugs along the way, fixed here too since they'd otherwise undermine the whole point of offering a smooth Docker path.

- Dockerfile: python:3.13-slim base, installs the system libraries needed to build toga-gtk's own dependencies (pycairo, PyGObject) even though the container never uses the GUI, then pre-fetches the front-end libraries, pairing tools and compiled translations at build time so a fresh container needs no network access to start.
- docker-entrypoint.sh answers the one-time "install example events?" interactive prompt so the container doesn't crash on EOF with no stdin attached.
- docker-compose.yml (production) and docker-compose.dev.yml (bind-mounts src/default-files/locale for live editing without rebuilding).
- Client._get_account() (src/data/access_levels/client.py) only ever granted administrator rights to requests from 127.0.0.1: under Docker's bridge networking the host's own requests are NAT'd to the container's default gateway instead, so nobody could ever pass that check. Now also trusts that gateway address, but only when SHARLY_CHESS_DOCKER=1 (set in the image) - native installs are unaffected.
- IndexAdminController.htmx_admin_tab (GET /{admin_tab:str}) matched any path segment and unconditionally touched the session to build a web context, which crashed on paths session middleware excludes (e.g. browser requests for /sw.js, /workbox-*.js) instead of a plain 404. Now validated against the actual set of tab keys first.
- LocalSourceDatabase._update ran its source-file download outside any try/except, so a GitHubLocalSourceDatabase (FIDE/FFE/FRA Schools) downloaded without its decryption password configured crashed its background thread instead of failing the update gracefully.
- Added a small UI (in the existing "Data sources" modal) to set the FIDE/FFE/FRA Schools local-database password from the app itself, instead of only via its generate_*_credentials.py script - handy everywhere, and pretty much required in Docker where there's no interactive shell open on the source tree by default.
- .gitattributes: force LF for *.sh, so docker-entrypoint.sh still builds correctly for contributors on Windows (core.autocrlf=true).


docs/technical-appendices/dev-setup.md
```diff
@@ -20,6 +20,32 @@ python --version
...
+
+```
+docker compose -f docker-compose.yml -f docker-compose.dev.yml up --build
+docker compose -f docker-compose.yml -f docker-compose.dev.yml restart
+```
+
+### Notes
+
+- Federation-specific plugins (FFE, FIDE, Chess-Results) each need their own credentials file to work; without them they log a warning at startup instead of failing. The FFE, FIDE and FRA Schools local-database passwords can be set from the running app itself, in the "Data sources" modal of the settings page.
+- The container is not meant to be exposed beyond the host it runs on: whichever client the host's Docker bridge network presents requests as (see `common.network.docker_gateway_ip`) is trusted as the local/administrator machine, the same way `127.0.0.1` is trusted outside of Docker.
```
timothyarmes, line 47: This needs thinking about.  It implies that in order to use the app we need to give out the passwords, which doesn't seem like the best approach

src/web/controllers/admin/index_admin_controller.py
```diff
@@ -340,6 +352,8 @@ async def htmx_admin_tab(
         show_details: FromQuery[bool | None] = None,
         filter_tags: FromQuery[str | None] = None,
     ) -> Template:
+        if admin_tab not in self.ADMIN_TAB_KEYS:
+            raise NotFoundException(f'Unknown admin tab [{admin_tab}].')
```
timothyarmes, line 356: Why is this needed?  There are a bunch more in the new championships branch, having them hard coded here means more maintenance and it's easy to miss.

src/web/controllers/admin/index_admin_controller.py
```diff
@@ -52,6 +52,7 @@
 from database.sqlite.event.event_database import EventDatabase
 from database.sqlite.event.event_store import StoredEvent
 from database.sqlite.local_source_database import (
+    GitHubLocalSourceDatabase,
```
timothyarmes, line 55: What's this used for?

docker-entrypoint.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+# On a brand-new data volume, Sharly Chess asks interactively whether to
+# install example event databases (dev-mode only, see
+# src/common/data_recovery.py). There is no non-interactive flag for it, so
+# answer "n" via stdin; `exec` keeps this process as PID 1 so signals
+# (docker stop) still reach it directly for a graceful shutdown.
+exec python src/sharly_chess.py --path /data "$@" <<'EOF'
```
timothyarmes, line 7: This feels a bit hacky - maybe best not to ask in that environment?

timothyarmes (changes requested):

PR #2220, opened 2026-08-25
Feature/player search filters

lefebvrele:
Draft pull request for player search filters


src/web/templates/admin/common/js/searchFilter.js
```diff
@@ -0,0 +1,74 @@
+var filtersInitialized = false;
+var filters = {};
+var filterRefreshTimer;
+var ENABLED_FILTERS_BY_DATASOURCE = {
```
timothyarmes: Here a comment file know about a plugin datasource.  It can't work like that.

timothyarmes (dismissed):

src/web/templates/admin/common/js/searchFilter.js
lefebvrele: fixed in my latest commit, the mapping has been moved to the backend, and I use a hook for both ffe datasources

timothyarmes (approved):

PR #2222, opened 2026-08-26
Add Keizer pairing system for individual tournaments

timothyarmes:


pascalaubry (approved):
Quite disturbing to always leave the pairing settings editable but let's go.

PR #2232, opened 2026-08-30
Display a clearer error message when pairings fail

timothyarmes:


pascalaubry (dismissed):
Did not test on the right branches, saw bug that you fixed with this PR.

pascalaubry (approved):

PR #2233, opened 2026-08-30
Clear SCE auth error when re-authenticating

timothyarmes:


pascalaubry (dismissed):
Also had this problem tonight.
Works now, thanks.

PR #2252, opened 2026-09-02
Compact, single line, non overflowing timer

timothyarmes:


pascalaubry (dismissed):
CSS is OK but the text is broken:

Image

PR #2254, opened 2026-09-04
Allow access to screens by clicking in the card/row when the only action it to access them

timothyarmes:


RELEASE_NOTES.md
```diff
@@ -92,6 +92,7 @@ In order to maintain compatibility with the display of table numbers on the _FFE
 - Fixed access to rotators and display controllers from the network (5.0.2)
 - Base clients on the server date to display the timers (5.0.3)
 - Improved timer display (5.0.3)
+- Network spectators can now open a screen by clicking anywhere on its card or list row when viewing it is the only available action (5.0.3)
```
pascalaubry: Since we just restored the 4.2 behaviour concerning the click-to-open, I would replace by something like:
```suggestion
- On public screens clients, improved list items and cards of screens, rotators and display controllers (5.0.3)
```

pascalaubry (approved):

